PR #6, opened 2018-03-02
basic sge implementation

### LINKED ISSUE #3: Add Grid Engine
It might be valuable to extend this repository with a solution for Grid Engine.  My hope is that this is relatively easy for someone with modest grid engine experience.  Looking at the current solutions for [PBS](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/pbs.py) or [SLURM](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/slurm.py) might be helpful (they're about 100 lines, mostly docstrings)

jhamman:
closes #3 

I don't think this is ready to merge. Turns out the SGE cluster I used to have access to is no more. Perhaps we should get a basic CI implementation setup before pushing more content into the repo?

### DIFF
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -120,7 +120,5 @@ class SGECluster(JobQueueCluster):
         self.jobs = dict()
         self.n = 0
         self._adaptive = None
-        self._submitcmd = 'qsub'
-        self._cancelcmd = 'qdel'
 
         logger.debug("Job script: \n %s" % self.job_script())
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -1,0 +1,126 @@
+import logging
+import os
+import socket
+import sys
+
+from distributed import LocalCluster
+from distributed.utils import get_ip_interface
+
+from .core import JobQueueCluster
+
+logger = logging.getLogger(__name__)
+
+dirname = os.path.dirname(sys.executable)
+
+
+class SGECluster(JobQueueCluster):
+    """ Launch Dask on a SGE cluster
+
+    Parameters
+    ----------
+    name : str
+        Name of worker jobs. Passed to `$SGE -N` option.
+    queue : str
+        Destination queue for each worker job. Passed to `#$ -q` option.
+    project : str
+        Accounting string associated with each worker job. Passed to
+        `#$ -A` option.
+    threads_per_worker : int
+        Number of threads per process.
+    processes : int
+        Number of processes per node.
+    memory : str
+        Bytes of memory that the worker can use. This should be a string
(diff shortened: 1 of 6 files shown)


dask_jobqueue/sge.py
```diff
@@ -0,0 +1,126 @@
...
+                       'resource_spec': resource_spec,
+                       'base_path': dirname,
+                       'memory': memory,
+                       'death_timeout': death_timeout,
+                       'extra': extra}
+        self.jobs = dict()
+        self.n = 0
+        self._adaptive = None
+        self._submitcmd = 'qsub'
+        self._cancelcmd = 'qdel'
```
mrocklin: These should maybe be class variables?

dask_jobqueue/sge.py
```diff
@@ -0,0 +1,126 @@
...
+        if interface:
+            host = get_ip_interface(interface)
+            extra += ' --interface  %s ' % interface
+        else:
+            host = socket.gethostname()
+
+        project = project or os.environ.get('SGE_ACCOUNT')
+        if not project:
+            raise ValueError("Must specify a project like `project='UCLB1234' "
+                             "or set SGE_ACCOUNT environment variable")
```
mrocklin: Thoughts on #7 ?

dask_jobqueue/tests/test_sge.py
```diff
@@ -0,0 +1,67 @@
...
+import pytest
+
+from dask.distributed import Client
+from distributed.utils_test import loop  # noqa: F401
+from pangeo import SGECluster
+
+
+def test_basic(loop):
+    with SGECluster(walltime='00:02:00', threads_per_worker=2, memory='7GB',
+                    interface='ib0', loop=loop) as cluster:
```
mrocklin: The interface keyword option here will likely fail on a CI system.

jakirkham:
> Perhaps we should get a basic CI implementation setup before pushing more content into the repo?

So Travis CI is enabled, but does not have a configuration file yet. Would encourage just stealing stuff from [`dask-drmaa`]( https://github.com/dask/dask-drmaa ) to get this working. There's a few files that you would want to grab outside of `.travis.yml`. They would be the `docker-compose.yml`, `Dockerfile`s, various assorted shell scripts, and some text files use for configuration.

dask_jobqueue/sge.py
guillaumeeb: Yep, agree for submit and cancel commands.

jhamman:
@lesteve - afaik, you're the only current dask-jobqueue user/developer with access to an sge cluster. You may be interested in this initial implementation of the SGECluster. It will require some reworking following #10 but may provide a starting point.

lesteve:
Thanks I'll try to take a look.

Issue #7, opened 2018-03-02
Not setting `project` should not raise 


bocklund:
One suggestion I have is not to raise an exception unless there is exceptional behavior. In the case of not having a project set, not all systems require users to set a project, so we should not enforce that by raising IMO. This would also apply to #6

mrocklin:
If people are ok with this idea would you be interested in submitting a PR
to resolve this?

bocklund:
I'd be happy to help. distributed Clients running on MPI is a core use case for us over at [ESPEI](https://github.com/phasesresearchlab/espei)

jhamman:
I think we could take some time to rethink the specification of cluster specific configuration options (project included). I've not worked on two computers that use identical syntax for the batch script headers so this is something we're going to run into repeatedly.

jhamman:
for the sake of argument, what if we changed the current cluster signature from:

```Python
PBSCluster(name='dask',
           queue='regular',
           project=None,
           resource_spec='select=1:ncpus=36:mem=109GB',
           threads_per_worker=4,
           processes=9,
           memory='7GB',
           walltime='00:30:00',
           interface=None,
           death_timeout=60,
           extra='',
           **kwargs)
```

to something more flexible like:

```Python
cluster_kws = {}
job_args =  ['-l select=1:ncpus=36:mem=109GB', '-l walltime=00:30:00',
                 '-q regular', 'A PROJECT']
worker_kws = dict(threads_per_worker=4, processes=9, memory='7GB',
                  interface=None, death_timeout=60, extra='',)

PBSCluster(name='dask', cluster_args=cluster_args, job_args=job_args,
           worker_kws=worker_kws)
```

where the job script is created programmatically:

```Python
header = ['#!/bin/bash', '#PBS -N %(name)s' % name]
header.extend(['#PBS %(arg)s' for arg in cluster_args])
header = '\n'.join(header)
```

mrocklin:
I agree that this might provide a valuable escape hatch for unique cluster situations.  I am slightly concerned that people are no longer prompted with options like `wall_time=`, `threads_per_worker=` etc, that they might use if presented with the option.  I wonder if we can do a mix of explicit keywords alongside catch-all dictionary keywords.

guillaumeeb:
From my point of view, it would be better to keep explicit keyword at least for basic PBS or Dask options.
And as @mrocklin said, mixing it with dictionary keywords or lists as proposed by @jhamman  seems really good.
[...]

Issue #2, opened 2018-03-01
CI for Schedulers


jakirkham:
Linking to @jhamman's [comment]( https://github.com/pangeo-data/pangeo/issues/85#issuecomment-360842913 ), one of the challenges here is setting up CI for Dask. There has been some already towards using SGE (particularly at dask-drmaa) for testing. Configuring other schedulers (e.g. SLURM, PBS, LSF, etc.) is possible, but can be a bit tricky. Docker helps with this, but it requires a fair bit of configuration to get right.

One resource that might be useful are [these Docker images]( https://bitbucket.org/deardooley/agave-docker/src/dc81db2504cb6f3f7f52778493040bd4005433f4/test-containers/schedulers/?at=master ), which include SLURM and TORQUE (similar to PBS). That said, I'm not sure what license applies to these. Also they may not work out of the box (as was my experience with using their Grid Engine some time back). That said, they should at least provide useful insights on how to get working Docker images for these different cases. For simplicity, would propose building and hosting images for respective schedulers so that CI can just download them.

luizirber:
This repo has an example using SLURM: https://github.com/Snakemake-Profiles/slurm

luizirber:
I can take a look for SLURM (can you assign this issue to me?)

mrocklin:
To be honest I rarely use github assignment.  It looks like this is difficult to do without having you as part of the organization (which is currently somewhat challenging).  We might have to go with informal assignment for the time being.

Regardless, thank you for stepping up!

jakirkham:
Grid Engine should basically be just copying over what dask-drmaa already has. Maybe that is the best place to start and then we can extend from there. Is Travis CI enabled here?

mrocklin:
Nope.  Nothing yet.

jakirkham:
Would you be willing to do that? I could copy over the testing code. Just don't think I have permissions to add CI is all.

mrocklin:
Do you know how I can give you that permission?

jakirkham:
Think I need to have admin on the repo as it requires the ability to add webhooks. Not sure if there is anything more granular. Honestly don't know if I'd need admin for anything else. Leave it up to you whether to grant it or not.

mrocklin:
OK, you have admin rights.  I guess the difference here is the owner/member of the org distinction?  Further reason to push on https://github.com/dask/dask/issues/3223

mrocklin:
[...]

PR #10, opened 2018-03-11
[WIP] Issue 7 : pbs cluster init, class inheritance refactoring

guillaumeeb:
So looking for comment here as this is my first pull request ever. 

I've made some significant changes from the master branch:
* As issue #7 first mentioned, I've remove the check on project being defined or not,
* I've made the PBS header construction, but also the dask-worker command construction more flexible, as suggested by @jhamman,
* I've moved some logic from PBS/SLURM classes to the JobQueueCluster parent class. It follows an abstract class pattern, but I've decided not to use ABC Metaclass for the sake of simplicity and python 2.7 easier compatibility as discussed with @mrocklin. So I've juste defined some abstract method by raising NotImplementedError in the JobQueueCluster class.

A question too on SLURMCluster. I've modified its implementation, but I've left the fixed header template logic here, as I am not a SLURM user. Anyway I still have some questioning about the CPU and memory resources in this class : we use only the processes (so nprocs) of dask here, but I would expect to see something like processes * nthreads; moreover, I don't see any memory limit in the template.

Finaly, I've not yet tested this in real condition, only some basic play with job script construction into a python terminal (and by the way I believe we should add some basic unit tests like that, don't you think?).

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -1,11 +1,14 @@
 from contextlib import contextmanager
 import logging
 import subprocess
-import toolz
 import socket
+import os
+import sys
 
 from distributed.utils import tmpfile, ignoring, get_ip_interface
 from distributed import LocalCluster
+
+dirname = os.path.dirname(sys.executable)
 
 logger = logging.getLogger(__name__)
 
@@ -16,43 +19,119 @@ class JobQueueCluster(object):
     This class should not be used directly, use inherited class appropriate
     for your queueing system (e.g. PBScluster or SLURMCluster)
 
+    Abstract init parameters
+    ------------------------
+    name : str
+        Name of Dask workers.
+    threads : int
+        Number of threads per process.
+    processes : int
+        Number of processes per node.
+    memory : str
+        Bytes of memory that the worker can use. This should be a string
+        like "7GB" that can be interpretted both by PBS and Dask.
+    interface : str
+        Network interface like 'eth0' or 'ib0'.
+    death_timeout : float
+        Seconds to wait for a scheduler before closing workers
+    local_directory : str
+        Dask worker local directory for file spilling.
+    extra : str
+        Additional arguments to pass to `dask-worker`
+    kwargs : dict
+        Additional keyword arguments to pass to `LocalCluster`
+
     See Also
     --------
     PBSCluster
     SLURMCluster
     """
-    def __new__(cls, *args, **kwargs):
-        #Prevent class instantiation
-        print(str(cls))
-        raise NotImplemented
+
+    _script_template ="""
+#!/bin/bash
+
+%(job_header)s
+
+%(worker_command)s 
+""".lstrip()
 
     def __init__(self,
-                 threads_per_worker=4,
+                 name='dask-worker',
+                 threads=4,
                  processes=6,
-                 memory='20GB',
+                 memory='16GB',
                  interface=None,
                  death_timeout=60,
-                 worker_extra='',
+                 local_directory=None,
+                 extra='',
                  **kwargs
                  ):
+        """
+        This initializer should be considered as Abstract, and never used directly.
+        :param name:
+        :param threads:
+        :param processes:
+        :param memory:
+        :param interface:
+        :param death_timeout:
+        :param local_directory:
+        :param extra:
+        :param kwargs:
+        """
 
 
         if interface:
             host = get_ip_interface(interface)
-            worker_extra += ' --interface  %s ' % interface
+            extra += ' --interface  %s ' % interface
         else:
             host = socket.gethostname()
 
         self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
-        self.memory = memory.replace(' ', '')
 
         self.jobs = dict()
         self.n = 0
         self._adaptive = None
 
+        self._command_template = "%s/dask-worker %s" % (dirname, self.scheduler.address)
+        if threads != None: self._command_template += " --nthreads %d" % threads
+        if processes != None: self._command_template += " --nprocs %d" % processes
+        if memory != None: self._command_template += " --memory-limit %s" % memory
+        if name != None:
+            self._command_template += " --name %s" % name
+            self._command_template += "-%(n)d" #Keep %(n) to be replaced later.
+        if death_timeout != None: self._command_template += " --death-timeout %s" % death_timeout
+        if local_directory != None: self._command_template += " --local-directory %s" % local_directory
+        if extra!= None: self._command_template += extra
+
     def job_script(self):
         self.n += 1
-        return self._template % toolz.merge(self.config, {'n': self.n})
+        return self._script_template % {'job_header': self.job_header,
+                                        'worker_command': self._command_template % {'n': self.n}
+                                        }
+
+    @property
+    def job_header(self):
+        """
+        Abstract attribute for the Job scheduler script header part.
+        :return: A string representation of the Job script header part.
+        """
+        raise NotImplementedError
+
+    @property
+    def submitcmd(self):
+        """
+        Abstract attribute for job scheduler submission command
+        :return:
+        """
+        raise NotImplementedError
+
+    @property
+    def cancelcmd(self):
+        """
+        Abstract attribute for job scheduler cancel command
+        :return:
+        """
+        raise NotImplementedError
 
     @contextmanager
     def job_file(self):
@@ -67,7 +146,7 @@ class JobQueueCluster(object):
         workers = []
         for _ in range(n):
             with self.job_file() as fn:
-                out = self._call([self._submitcmd, fn])
+                out = self._call([self.submitcmd, fn])
                 job = out.decode().split('.')[0]
                 self.jobs[self.n] = job
                 workers.append(self.n)
@@ -127,7 +206,7 @@ class JobQueueCluster(object):
             return
         workers = list(map(int, workers))
         jobs = [self.jobs[w] for w in workers]
-        self._call([self._cancelcmd] + list(jobs))
+        self._call([self.cancelcmd] + list(jobs))
         for w in workers:
             with ignoring(KeyError):
                 del self.jobs[w]
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -19,8 +19,8 @@ class JobQueueCluster(object):
     This class should not be used directly, use inherited class appropriate
     for your queueing system (e.g. PBScluster or SLURMCluster)
 
-    Abstract init parameters
-    ------------------------
+    Parameters
+    ----------
     name : str
         Name of Dask workers.
     threads : int
@@ -41,19 +41,30 @@ class JobQueueCluster(object):
     kwargs : dict
         Additional keyword arguments to pass to `LocalCluster`
 
+    Attributes
+    ----------
+    submit_command: str
+        Abstract attribute for job scheduler submit command, should be overriden
+    cancel_command: str
+        Abstract attribute for job scheduler cancel command, should be overriden
+
     See Also
     --------
     PBSCluster
     SLURMCluster
     """
 
-    _script_template ="""
+    _script_template = """
 #!/bin/bash
 
 %(job_header)s
 
-%(worker_command)s 
+%(worker_command)s
 """.lstrip()
+
+    #Following class attributes should be overriden by extending classes.
+    submit_command = None
+    cancel_command = None
 
     def __init__(self,
                  name='dask-worker',
@@ -68,17 +79,9 @@ class JobQueueCluster(object):
                  ):
         """
         This initializer should be considered as Abstract, and never used directly.
-        :param name:
-        :param threads:
-        :param processes:
-        :param memory:
-        :param interface:
-        :param death_timeout:
-        :param local_directory:
-        :param extra:
-        :param kwargs:
         """
-
+        if not self.cancel_command or not self.submit_command:
+            raise NotImplementedError('JobQueueCluster is an abstract class that should not be instanciated.')
 
         if interface:
             host = get_ip_interface(interface)
@@ -92,16 +95,23 @@ class JobQueueCluster(object):
         self.n = 0
         self._adaptive = None
 
+        #dask-worker command line build
         self._command_template = "%s/dask-worker %s" % (dirname, self.scheduler.address)
-        if threads != None: self._command_template += " --nthreads %d" % threads
-        if processes != None: self._command_template += " --nprocs %d" % processes
-        if memory != None: self._command_template += " --memory-limit %s" % memory
-        if name != None:
+        if threads is not None:
+            self._command_template += " --nthreads %d" % threads
+        if processes is not None:
+            self._command_template += " --nprocs %d" % processes
+        if memory is not None:
+            self._command_template += " --memory-limit %s" % memory
+        if name is not None:
             self._command_template += " --name %s" % name
             self._command_template += "-%(n)d" #Keep %(n) to be replaced later.
-        if death_timeout != None: self._command_template += " --death-timeout %s" % death_timeout
-        if local_directory != None: self._command_template += " --local-directory %s" % local_directory
-        if extra!= None: self._command_template += extra
+        if death_timeout is not None:
+            self._command_template += " --death-timeout %s" % death_timeout
+        if local_directory is not None:
+            self._command_template += " --local-directory %s" % local_directory
+        if extra is not None:
+            self._command_template += extra
 
     def job_script(self):
         self.n += 1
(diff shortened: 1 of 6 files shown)


dask_jobqueue/pbs.py
```diff
@@ -60,66 +62,44 @@ class PBSCluster(JobQueueCluster):
 
     >>> cluster.adapt()
     """
+
+    _submitcmd = 'qsub'
+    _cancelcmd = 'qdel'
```
mrocklin: It might be nice to rename these to the more explicit (if perhaps also longer) `submit_command` and `cancel_command` and remove the properties that point to them.

dask_jobqueue/pbs.py
```diff
@@ -60,66 +62,44 @@ class PBSCluster(JobQueueCluster):
...
-        self.n = 0
-        self._adaptive = None
-        self._submitcmd = 'qsub'
-        self._cancelcmd = 'qdel'
+
+        header_lines = ['#PBS -N %s' % name]
+        if queue != None: header_lines.append('#PBS -q %s' % queue)
+        if project != None: header_lines.append('#PBS -A %s' % project)
+        if resource_spec != None: header_lines.append('#PBS -l %s' % resource_spec)
+        if walltime != None: header_lines.append('#PBS -l walltime=%s' % walltime)
```
mrocklin: Most other dask related projects use the `flake8` style checker, which would probably raise some warnings here for combining the if block into a single line.  You might want to run the following from the root of this directory:

    flake8 dask_jobqueue

But we should probably add this to CI so that we can automate this.

dask_jobqueue/core.py
```diff
@@ -14,17 +19,119 @@ class JobQueueCluster(object):
...
+        This initializer should be considered as Abstract, and never used directly.
+        :param name:
+        :param threads:
+        :param processes:
+        :param memory:
+        :param interface:
+        :param death_timeout:
+        :param local_directory:
+        :param extra:
+        :param kwargs:
```
mrocklin: So far we've been using numpydoc style docstrings: http://dask.pydata.org/en/latest/develop.html#docstrings

dask_jobqueue/core.py
```diff
@@ -14,17 +19,119 @@ class JobQueueCluster(object):
...
+        """
+        raise NotImplementedError
+
+    @property
+    def cancelcmd(self):
+        """
+        Abstract attribute for job scheduler cancel command
+        :return:
+        """
+        raise NotImplementedError
```
mrocklin: I'm in favor of removing these properties and just going with class attributes if possible

dask_jobqueue/pbs.py
```diff
@@ -1,42 +1,44 @@
...
-    threads_per_worker : int
+    resource_spec : str
+        Request resources and specify job placement. Passed to `#PBS -l`
+        option.
+    walltime : str
+        Walltime for each worker job.
+    job_extra : list
+        List of other PBS options, for example -j oe. Passed with '#PBS ' prefix
+    local_directory : str
+        Dask worker local directory for file spilling.
```
mrocklin: Why remove threads_per_worker above?  Also did you want to logically separate the dask options from the PBS options?

mrocklin: Ah, scrolling down I see that you've done just that.  If so then local_directory should probably be an inherited parameter.  I *think* it should be generic.

dask_jobqueue/pbs.py
guillaumeeb: Okay, I'll try this, comming from Java background, it was more natural to me to override kind of methods instead of attributes :).

dask_jobqueue/pbs.py
guillaumeeb: That's fine with me. Perhaps we could enable Travis CI to do some basic tests and style checking, without having all the docker testing env yet? Should we raise an issue separated from #2 ?

dask_jobqueue/pbs.py
guillaumeeb: Yup I separated options, and for more consistancy with other keywords I used `threads` instead of `thread_per_worker`. 
`local_directory` is generic, but I wanted to override it in PBSCluster, because standard PBS installation always provide a $TMPDIR variable which resolve to something like /tmp/{JOB_ID}/. I wanted to make this as the default local_directory for PBSCluster workers.

dask_jobqueue/pbs.py
jhamman: I've done this in #13. You can also run the same checker locally by running: `flake8 -j auto dask_jobqueue`

guillaumeeb:
So I've pushed another commit with @mrocklin comments taken care of. Tell me how you feel about it.

guillaumeeb:
Oh, And I've tested on our cluster, and it worked like a charm (even if I did not push the really far).

dask_jobqueue/core.py
```diff
@@ -14,17 +19,116 @@ class JobQueueCluster(object):
...
+            host = socket.gethostname()
+
+        self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+
+        self.jobs = dict()
+        self.n = 0
+        self._adaptive = None
+
+        #dask-worker command line build
+        self._command_template = "%s/dask-worker %s" % (dirname, self.scheduler.address)
```
jhamman: slightly safer to write this as:

```Python
self._command_template = os.path.join(dirname, 'dask-worker' % self.scheduler.address)
```

dask_jobqueue/core.py
```diff
@@ -14,17 +19,116 @@ class JobQueueCluster(object):
...
+    @property
+    def job_header(self):
+        """
+        Abstract attribute for the Job scheduler script header part.
+
+        Returns
+        -------
+        A string containing multiple lines to be used as header of job file to be sumitted.
+        """
+        raise NotImplementedError
```
jhamman: Let's just make this a class attribute that defaults to an empty string.

dask_jobqueue/pbs.py
```diff
@@ -49,7 +51,7 @@ class PBSCluster(JobQueueCluster):
     Examples
     --------
     >>> from dask_jobqueue import PBSCluster
-    >>> cluster = PBSCluster(project='...')
+    >>> cluster = PBSCluster(queue= 'regular', project='DaskOnPBS')
```
jhamman: pep8 (remove space after `=`)

dask_jobqueue/pbs.py
```diff
@@ -60,66 +62,43 @@ class PBSCluster(JobQueueCluster):
...
-                 threads_per_worker=4,
-                 processes=9,
-                 memory='7GB',
+                 resource_spec='select=1:ncpus=24:mem=100GB',
                  walltime='00:30:00',
-                 interface=None,
-                 death_timeout=60,
-                 extra='',
+                 job_extra=[],
+                 local_directory='$TMPDIR',
```
jhamman: I have a slight preference to set this to None by default. Either that or `os.getenv('TMPDIR', None)`.

dask_jobqueue/pbs.py
```diff
@@ -60,66 +62,43 @@ class PBSCluster(JobQueueCluster):
...
-                 name='dask',
-                 queue='regular',
+                 name='dask-worker',
+                 queue=None,
                  project=None,
-                 resource_spec='select=1:ncpus=36:mem=109GB',
-                 threads_per_worker=4,
-                 processes=9,
-                 memory='7GB',
+                 resource_spec='select=1:ncpus=24:mem=100GB',
```
jhamman: I'm not totally sure what to do here. One option is to leave this as `None` and force (most) users to set this. Keeping it as is may be okay too but everyone will have to change it (except for you :wink:). I have a slight preference for setting it to None.

dask_jobqueue/pbs.py
```diff
@@ -1,42 +1,44 @@
...
         Accounting string associated with each worker job. Passed to
         `#PBS -A` option.
-    threads_per_worker : int
+    resource_spec : str
+        Request resources and specify job placement. Passed to `#PBS -l`
+        option.
+    walltime : str
+        Walltime for each worker job.
+    job_extra : list
+        List of other PBS options, for example -j oe. Passed with '#PBS ' prefix
```
jhamman: maybe say: Each option will be prepended with the `#PBS ` prefix.

dask_jobqueue/slurm.py
```diff
@@ -54,27 +51,24 @@ def __init__(self,
...
-            Additional keyword arguments to pass to `LocalCluster`
+            Additional keyword arguments to pass to `JobQueueCluster` and `LocalCluster`
         """
-        self._template = """
-#!/bin/bash
 
+        super(SLURMCluster, self).__init__(name=name, processes=processes, **kwargs)
+
+        #TODO has this been tested? This seems weird to use only processes, and not processes * threads
+        # There are no memory limit given to Slurm either?
```
jhamman: I think the initial implementation left those out but we should provide both. @jedwards4b - do you have a comment here?

dask_jobqueue/slurm.py
```diff
@@ -86,43 +80,22 @@ def __init__(self,
...
-                       'death_timeout': death_timeout,
-                       'extra': extra}
-        self.jobs = dict()
-        self.n = 0
-        self._adaptive = None
-        self._submitcmd = 'sbatch'
-        self._cancelcmd = 'scancel'
+                       # Not used
+                       'memory': memory
+                       }
```
jhamman: see comment above. I think we'll want to include `memory`.

dask_jobqueue/core.py
guillaumeeb: Okay, I tried to keep one of my "contract" of Abstract Class, but I guess it is still too Java oriented :).

dask_jobqueue/pbs.py
guillaumeeb: I believed that $TMPDIR was a PBS default, but apparently it should be PBS_TMPDIR. Can you check on the cluster you are using?
Maybe it is more generic to just remove this arg by default.

dask_jobqueue/pbs.py
guillaumeeb: Yep I have the same concern. I found your defaults too high, so I set mines, but it is probably wrong too.

On our cluster if we don't configure it, the default will be `select=1:ncpus=1:mem=5GB`. So worker jobs will probably be killed promptly if dask worker is set with more than 1 thread or process. And I believe it would be really bad for PBS to launch one job for each 1 process dask-worker.

My preference would be more close to what is mentioned in #12, if nothing is provided, resource_spec is computed using dask-worker options. Still not enthusiast about that? It's a good compromise, keep resource_spec keyword, but build it from dask-worker options if not provided. The only concern if for memory limit, there may be a need to extrapolate from String to integer...

So maybe it is simpler to set it to None :)...

mrocklin: We might inspect available memory and cpu cores using tools like psutil and multiprocessing

```python
In [1]: import psutil
In [2]: psutil.virtual_memory().total
Out[2]: 16680607744
In [3]: psutil.cpu_count()
Out[3]: 4
```

dask_jobqueue/pbs.py
jhamman: On our system (NCAR's cheyenne), `TMPDIR` is set to `/var/tmp/{JOB_ID}`.  So given the divergence, I suggest just removing this as a default.

dask_jobqueue/pbs.py
jhamman: That may be okay. The challenge is that you'll routinely have the situation where the machine that scheduler is running on is spec'd differently than the machines where the workers will be running on.  I think this may be worth doing as a first guess.

guillaumeeb: I'm not inclined to base the resource_spec on the submitting node resource...
I would much prefer to guess it from the given threads and processes. Else I would default to None, maybe with a warning?

dask_jobqueue/pbs.py
guillaumeeb: Then our systems are configured with the same variable (if not set to the same directory), it is difficult to generalise from two examples, but TMPDIR seems to be used for a scratch dir for you and me. Stilled removed it maybe I will add a comment for this.

dask_jobqueue/core.py
```diff
@@ -14,17 +19,108 @@ class JobQueueCluster(object):
...
+                 name='dask-worker',
+                 threads=4,
+                 processes=6,
+                 memory='16GB',
+                 interface=None,
+                 death_timeout=60,
+                 local_directory=None,
+                 extra='',
+                 **kwargs
+                 ):
```
jhamman: Can we add these options to the individual cluster doc strings?

guillaumeeb: Not sure I get your point, where do you want to put them? They are already in the JobQueueCluster doc strings. You wan to put it in every implementation?

dask_jobqueue/pbs.py
jedwards4b: These will tend to be different on every machine - what about incorporating a user config file 
that would allow local customization?

guillaumeeb: @jedwards4b Hmm, could you elaborate?
From my understanding of your suggestion, I would prefer to only have to change a keyword value than having a user config file, but it is just my feeling.

guillaumeeb: As for the comment from @mrocklin above, for me this could be put into the JobQueueCluster class, for choosing default processes and/or threads and memory. Maybe we ought to set those to None by default?
Commit incoming for demonstrating the automatic resource_spec assignment.

guillaumeeb: If we cannot reach a common view about this, I propose to continue discussing it in another issue, maybe #12 or a new one. Then for this pull request, I can put resource_spec to None by default, remove the solution I proposed with my last commit if you don't like it, and eventually merge all this?

jedwards4b: These values are different from one machine to another, but typically don't need to change from one run to another on the same machine, sure a cli should be provided but a config file would allow the user not to have to remember and add the same cli again and again.

dask_jobqueue/core.py
jhamman: My point was just that the JobQueueCluster doc string doesn't show up on the PBSCluster doc string so it requires some unnecessary searching to find all the possible arguments to the constructor.

dask_jobqueue/slurm.py
```diff
@@ -54,27 +51,24 @@ def __init__(self,
...
         """
-        self._template = """
-#!/bin/bash
 
+        super(SLURMCluster, self).__init__(name=name, processes=processes, **kwargs)
+
+        #TODO has this been tested? This seems weird to use only processes, and not processes * threads
+        # There are no memory limit given to Slurm either?
+
+        #Keeping template for now has I don't know much about slurm.
```
mrocklin: Comments like these can probably be removed

dask_jobqueue/tests/test_pbs.py
```diff
@@ -5,11 +5,11 @@
 
 from dask.distributed import Client
 from distributed.utils_test import loop  # noqa: F401
-from pangeo import PBSCluster
+from dask_jobqueue import PBSCluster
 
 
 def test_basic(loop):
-    with PBSCluster(walltime='00:02:00', threads_per_worker=2, memory='7GB',
+    with PBSCluster(walltime='00:02:00', threads=2, memory='7GB',
```
mrocklin: I'm curious if this name is perhaps a little ambiguous now.  It might be worth it for the shorter name, I'm not sure.

mrocklin:
In general it's fine with me.  I'm also happy to defer to others' judgment.

dask_jobqueue/slurm.py
guillaumeeb: Removed.

dask_jobqueue/core.py
guillaumeeb: There was already some duplicated doc, I updated it.
Does not feel really good about that though, don't like to duplicate docstrings... Is there a better way?

dask_jobqueue/slurm.py
guillaumeeb: And created a new issue #22

dask_jobqueue/tests/test_pbs.py
guillaumeeb: If it would be me, I would name every corresponding attribute with suffix liker "per_worker" or prefix like "worker_". Like mentioned in https://github.com/dask/dask-jobqueue/issues/7#issuecomment-370579557.
As it stands, I don't see why threads should be treated differently than memory. Or maybe we should change memory to memory_per_worker, as for processes it actually translates in "number of workers"?

dask_jobqueue/core.py
leej3: You could directly take the class docs and add to them. i.e. in each child class definition you do something like:
```
        doc = JobQueueCluster.__doc__
        self.__doc__ +=  "specific docs for Slurm cluster"
```
Perhaps there's a less hacky way though.

jhamman: @rabernat - do I recall you introducing the xarray devs to a decorator approach to doc string templating? I can't seem to dig it up but perhaps someone knows what I'm talking about.

rabernat:
http://docrep.readthedocs.io/en/latest/

guillaumeeb:
@jhamman I've included docrep for docstrings depency as suggested by @rabernat.
Anything else?

jhamman:
Thanks @guillaumeeb - I hope to work on this a bit tomorrow so I'm merging now.

lesteve:
@rabernat I did not know about docrep, nice! Just curious have you used docrep in a project already? If that's the case I would be interested in hearing the kind of experience you had with it!

rabernat:
Docrep is used in this module: https://github.com/xgcm/xgcm/blob/master/xgcm/grid.py

It's useful when many different docstrings have to share very similar components...avoids repetition and thus potential errors.

lesteve:
Nice, thanks for the link!

PR #19, opened 2018-03-21
No queue arguments needed on SLURM for most clusters.

### LINKED ISSUE #15: From what environment should I call SLURMCluster? Parsing Issue in sbatch?
I am running on a university [cluster](https://www.rc.ufl.edu/services/hipergator/) with a SLURM submission scheduler. I installed conda and sourced the pangeo environment without error. 

I git cloned this repo, opened python and tried

```
(pangeo) [b.weinstein@gator4 dask-jobqueue]$ python
Python 3.6.4 | packaged by conda-forge | (default, Dec 23 2017, 16:31:06) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dask_jobqueue import SLURMCluster
>>> 
>>> cluster = SLURMCluster(project="ewhite")
>>> cluster.start_workers(10)
sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

sbatch: error: Invalid argument: ewhite

[2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
```
Where ewhite is the group tag (see below).

In previous submission scripts, I would typically start with something like

```
#!/bin/bash
#SBATCH --job-name=NEON_LIDAR   # Job name
#SBATCH --mail-type=END               
#SBATCH --mail-user=#######@gmail.com
#SBATCH --account=ewhite
#SBATCH --qos=ewhite-b

#SBATCH --ntasks=1                 # Number of MPI ranks
#SBATCH --cpus-per-task=1            # Number of cores per MPI rank
#SBATCH --mem=4000
#SBATCH --time=00:59:00       #Time limit hrs:min:sec
#SBATCH --output=NEON_LIDAR.out   # Standard output and error log
#SBATCH --error=NEON_LIDAR.err
```
which works without a problem.

## Attempts

I changed 
```
#SBATCH -A %(project)s
```
to 
```
#SBATCH --account %(project)s
```
to match previous submission scripts. No change.

So I tried removing the line, which yielded

```
>>> cluster.start_workers(10)
sbatch: error: Invalid argument: 00:30:00

sbatch: error: Invalid argument: 00:30:00

sbatch: error: Invalid argument: 00:30:00

...
```
This is an error in the walltime argument. This makes me believe that it doesn't have to do with the argument value, but how all the arguments are parsed and passed to bash. The most likely explanation is that i'm calling this script from an incorrect location, and the variable expansion or quotes are not being parsed in the way I anticipate.

bw4sz:
Removed queue from default arguments. The default was empty, which caused the following arguments to be poorly parsed by SLURM. The university clusters I have worked with do not use the -p flag. Of course, it may vary by cluster. This reverts #18, which probably wasn't needed. 

Closes #15

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -4,9 +4,7 @@ import subprocess
 import toolz
 
 from distributed.utils import tmpfile, ignoring
-
 logger = logging.getLogger(__name__)
-
 
 class JobQueueCluster(object):
     """ Base class to launch Dask Clusters for Job queues
@@ -75,6 +73,7 @@ class JobQueueCluster(object):
         Also logs any stderr information
         """
         logger.debug("Submitting the following calls to command line")
+        
         for cmd in cmds:
             logger.debug(' '.join(cmd))
         procs = [subprocess.Popen(cmd,
--- a/dask_jobqueue/slurm.py
+++ b/dask_jobqueue/slurm.py
@@ -48,12 +48,9 @@ class SLURMCluster(JobQueueCluster):
         ----------
         name : str
             Name of worker jobs. Passed to `#SBATCH -J` option.
-        queue : str
-            Destination queue for each worker job.
-            Passed to `#SBATCH -p` option.
         project : str
             Accounting string associated with each worker job. Passed to
-            `#SBATCH --account` option.
+            `#SBATCH -A` option.
         threads_per_worker : int
             Number of threads per process.
         processes : int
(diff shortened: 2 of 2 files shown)


jhamman:
Let's put this on hold until after #10 gets merged. I think there will need to be some minor modifications to the SLURMCluster but they should come after #10.

guillaumeeb:
Yup, for consistency, we will have to refactor SLURMCluster as mentionned in #22. As for this pull request, I'm not entirely sure removing queue keyword is a good thing. Don't know Slurm, but I believe being able to specify a queue is a good thing.

After #10 is merged, we may put the `queue` keyword as `None` by default, and by building the job_script dynamically, we could avoid add the `-p` option being positioned in the script.

PR #21, opened 2018-03-23
fix slurm config when queue not specified

leej3:
I ran into a problem when the queue is not specified. The sbatch file does not have the appropriate -p argument and the sbatch submission fails. This fixes it by getting the default slurm queue/partition.

Otherwise this tool worked really nicely. Thanks

### DIFF
--- a/dask_jobqueue/slurm.py
+++ b/dask_jobqueue/slurm.py
@@ -2,6 +2,8 @@ import logging
 import os
 import socket
 import sys
+import subprocess
+import re
 
 from distributed import LocalCluster
 from distributed.utils import get_ip_interface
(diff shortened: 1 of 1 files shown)


bocklund:
This should close the original issue of #7 

If we want to implement some of those ideas, maybe we should open another issue with concrete actionable items?

guillaumeeb:
As mentioned in #19, we should wait on #10 merge. By the ways #19 and #21 kind of disagree. 
As I said in https://github.com/dask/dask-jobqueue/pull/19#issuecomment-375795861, maybe it is better to set queue=None as a default, and if so, not add the -p option in the header of the job_script?
Will Slurm use the default queue (as PBS does) if none is specified?

leej3:
it does yes so removing it from the job script would also work

Issue #26, opened 2018-03-29
failure to autoscale unless workers are already present


jhamman:
I am testing the PBSCluster along with autoscaling. It seems that I am unable to get the cluster to launch any workers without explicitly starting at least one worker. I would expect that this configuration would scale from 0 to 10 (180 processes) without further interaction/configuration.

```Python
    cluster = PBSCluster(queue='default',
                         walltime='01:00:00',
                         project='MyAccount',
                         resource_spec='1:ncpus=36:mpiprocs=36:mem=109GB',
                         interface='ib0',
                         threads=4,
                         processes=18)
    client = Client(cluster)
    cluster.adapt(minimum=0, maximum=10)
```

@mrocklin - this may actually be a problem with the dask adaptive cluster but I wanted to discuss here to see if I am missing something obvious specific to PBS.

jhamman:
Here's what I'm seeing in some more detail:

```
In [1]: from dask_jobqueue import PBSCluster
   ...:

In [2]: cluster = PBSCluster(project='P48500028', processes=9, queue='premium', walltime='00:05:00',
   ...:                      resource_spec='select=1:ncpus=36', interface='ib0')
   ...:
In [3]: cluster.adapt(minimum=1, maximum=10)
Out[3]: <distributed.deploy.adaptive.Adaptive at 0x2aaac87b5c18>

In [4]: cluster._adaptive.log
Out[4]:
deque([(1522436688.2028198, 'up', {'n': 1}),
       (1522436689.096044, 'up', {'n': 1}),
       (1522436690.096046, 'up', {'n': 1}),
       (1522436691.096431, 'up', {'n': 1}),
       (1522436692.0965822, 'up', {'n': 1}),
       (1522436693.0963206, 'up', {'n': 1}),
       (1522436694.0966222, 'up', {'n': 1}),
       (1522436695.0969365, 'up', {'n': 1}),
       (1522436696.0968359, 'up', {'n': 1}),
       (1522436697.0959227, 'up', {'n': 1}),
       (1522436698.0963962, 'up', {'n': 1}),
       (1522436699.0967467, 'up', {'n': 1}),
       (1522436700.0968142, 'up', {'n': 1}),
       (1522436701.0960631, 'up', {'n': 1}),
       (1522436702.096041, 'up', {'n': 1}),
       (1522436703.096142, 'up', {'n': 1}),
       (1522436704.0963974, 'up', {'n': 1}),
[...]

PR #27, opened 2018-04-04
SGE implementation

### LINKED ISSUE #3: Add Grid Engine

lesteve:
I revived #6 and added simple tests. I was able to get `test_basic` to pass but not `test_adaptive` yet. 

I'll wait that #25 is merged before working on this further.

Fix #3. Closes #6.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -5,6 +5,7 @@ import socket
 import os
 import sys
 import docrep
+import shlex
 
 from distributed.utils import tmpfile, ignoring, get_ip_interface, parse_bytes
 from distributed import LocalCluster
@@ -153,8 +154,8 @@ class JobQueueCluster(Cluster):
         workers = []
         for _ in range(n):
             with self.job_file() as fn:
-                out = self._call([self.submit_command, fn])
-                job = out.decode().split('.')[0]
+                out = self._call(shlex.split(self.submit_command) + [fn])
+                job = out.decode().split('.')[0].strip()
                 self.jobs[self.n] = job
                 workers.append(self.n)
         return workers
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -4,8 +4,9 @@ import subprocess
 import socket
 import os
 import sys
-import docrep
 import shlex
+
+import docrep
 
 from distributed.utils import tmpfile, ignoring, get_ip_interface, parse_bytes
 from distributed import LocalCluster
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -71,14 +71,14 @@ class SGECluster(JobQueueCluster):
         if walltime is not None:
             header_lines.append('#$ -l h_rt=%(walltime)s')
         header_lines.extend(['#$ -cwd', '#$ -j y'])
-        self._header_template = '\n'.join(header_lines)
+        header_template = '\n'.join(header_lines)
 
-        self.config = {'name': self.name,
-                       'queue': queue,
-                       'project': project,
-                       'processes': self.worker_processes,
-                       'walltime': walltime,
-                       'resource_spec': resource_spec,}
-        self.job_header = self._header_template % self.config
+        config = {'name': self.name,
+                  'queue': queue,
+                  'project': project,
+                  'processes': self.worker_processes,
+                  'walltime': walltime,
+                  'resource_spec': resource_spec,}
+        self.job_header = header_template % config
 
         logger.debug("Job script: \n %s" % self.job_script())
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -2,13 +2,14 @@ import logging
 import os
 import sys
 
-from .core import JobQueueCluster
+from .core import JobQueueCluster, docstrings
 
 logger = logging.getLogger(__name__)
 
 dirname = os.path.dirname(sys.executable)
 
 
+@docstrings.with_indent(4)
 class SGECluster(JobQueueCluster):
     """ Launch Dask on a SGE cluster
 
@@ -26,8 +27,6 @@ class SGECluster(JobQueueCluster):
         Walltime for each worker job.
     interface : str
         Network interface like 'eth0' or 'ib0'.
-    kwargs : dict
-        Additional keyword arguments to pass to `LocalCluster`
     %(JobQueueCluster.parameters)s
 
     Examples
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -86,7 +86,7 @@ class SGECluster(JobQueueCluster):
         header_lines.extend(['#$ -cwd', '#$ -j y'])
         self._header_template = '\n'.join(header_lines)
 
-        self.config = {'name': getattr(self, 'name', 'default-name'),
+        self.config = {'name': self.name,
                        'queue': queue,
                        'project': project,
                        'processes': self.worker_processes,
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -25,8 +25,6 @@ class SGECluster(JobQueueCluster):
         option.
     walltime : str
         Walltime for each worker job.
-    interface : str
-        Network interface like 'eth0' or 'ib0'.
     %(JobQueueCluster.parameters)s
 
     Examples
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -1,12 +1,8 @@
 import logging
-import os
-import sys
 
 from .core import JobQueueCluster, docstrings
 
 logger = logging.getLogger(__name__)
-
-dirname = os.path.dirname(sys.executable)
 
 
 @docstrings.with_indent(4)
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -1,0 +1,139 @@
+import logging
+import os
+import socket
+import sys
+
+from distributed import LocalCluster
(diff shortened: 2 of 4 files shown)


dask_jobqueue/tests/test_sge.py
```diff
@@ -1,9 +1,70 @@
...
+                assert time() < start + 10
+
+            start = time()
+            while cluster.jobs:
+                sleep(0.100)
+                assert time() < start + 10
+
+
+@pytest.mark.skipif('SGE_ACCOUNT' in os.environ, reason='SGE_ACCOUNT defined')  # noqa: F811
+def test_errors(loop):
```
jhamman: I don't think this test is relevant anymore.

dask_jobqueue/tests/test_sge.py
```diff
@@ -1,9 +1,70 @@
...
+
+            start = time()
+            while len(client.scheduler_info()['workers']) > 0:
+                sleep(0.100)
+                assert time() < start + 10
+
+            start = time()
+            while cluster.jobs:
+                sleep(0.100)
+                assert time() < start + 10
```
jhamman: What is this testing exactly?  @mrocklin - given the recent development of `Adaptive`, can you help us flesh out what kind of tests we should be looking at here?

dask_jobqueue/tests/test_sge.py
lesteve: Yeah I was wondering about that and I felt the same way.

mrocklin: I think that this test came from a cluster where `project=` was necessary to run, and was often set implicitly with an environment variable (perhaps `PBS_ACCOUNT` on the cluster on which I was working at the time).  If this is not generally true for SGE clusters then I agree that this test should be removed.

dask_jobqueue/tests/test_sge.py
mrocklin: I've simplified the test for now.

mrocklin: (I hope you don't mind my pushing to your fork @lesteve )

lesteve: > I've simplified the test for now.

Thanks!

> (I hope you don't mind my pushing to your fork @lesteve )

Not at all, pushing into a PR's fork is a feature I use this feature quite often myself, it is very convenient and saves a lot of back and forth for minor things!

lesteve:
OK I am removing the WIP. A few comments:
* I could not find an equivalent of PBS_ACCOUNT in SGE so I removed the associated code and `test_errors` test
* AFAICT adaptive is currently broken, so I am leaving `test_adaptive` for later.

IMO this should be merged like this since this is already a clear improvement. It is a first implementation of SGE (probably by no means perfect) and allows to test some of the common code in `dask_jobqueue/core.py` across schedulers.

More details about the adaptive problems I saw (maybe related to https://github.com/dask/dask-jobqueue/issues/26 not sure):
According to https://github.com/dask/dask-jobqueue/blob/bdb2c2630d440aad8f7e9a47aeb7cfc2b6776129/dask_jobqueue/core.py#L220-L224, `scale_down` only does something `if isinstance(workers, dict)`. The problem is that workers is a list of `tcp://...` worker addresses. So `scale_down` does nothing i.e. `self.jobs` stays the same so that `scale_up` doesn't see the need to schedule more jobs:
https://github.com/dask/dask-jobqueue/blob/bdb2c2630d440aad8f7e9a47aeb7cfc2b6776129/dask_jobqueue/core.py#L217-L218
In other words your `*Cluster` Python object think he still have enough workers but your scheduling system has no jobs in running state that can process the work.

I guess this is all fixable by looking at the [doc](http://distributed.readthedocs.io/en/latest/adaptive.html#adaptive-class-interface), the `distributed.LocalCluster` implementation and possibly dask_drmaa as well. I may not find time to look at this for a few days so if anyone wants to pick this up, please be my guest!

dask_jobqueue/sge.py
```diff
@@ -0,0 +1,97 @@
...
+    #Override class variables
+    submit_command = 'qsub -terse'
+    cancel_command = 'qdel'
+
+    def __init__(self,
+                 queue=None,
+                 project=None,
+                 resource_spec=None,
+                 walltime='0:30:00',
+                 interface=None,
```
guillaumeeb: I believe interface=None should not be here. It is a keyword from the parent class, and you don't seem to use it.

dask_jobqueue/sge.py
```diff
@@ -0,0 +1,97 @@
...
+    Parameters
+    ----------
+    name : str
+        Name of worker jobs. Passed to `$SGE -N` option.
+    queue : str
+        Destination queue for each worker job. Passed to `#$ -q` option.
+    project : str
+        Accounting string associated with each worker job. Passed to
+        `#$ -A` option.
+    threads_per_worker : int
```
guillaumeeb: For all the parameters inherited from JobQueueCluster, you should use the same mechanism as in PBSCluster or SLURMCluster, with the docstrings module: `%(JobQueueCluster.parameters)s`

dask_jobqueue/core.py
```diff
@@ -153,8 +154,8 @@ def start_workers(self, n=1):
         workers = []
         for _ in range(n):
             with self.job_file() as fn:
-                out = self._call([self.submit_command, fn])
-                job = out.decode().split('.')[0]
+                out = self._call(shlex.split(self.submit_command) + [fn])
```
guillaumeeb: I don't know shlex, is this mandatory? It serves what purpose?

dask_jobqueue/sge.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.config = {'name': self.name,
+                       'queue': queue,
+                       'project': project,
+                       'processes': self.worker_processes,
+                       'walltime': walltime,
+                       'resource_spec': resource_spec,}
+        self.job_header = self._header_template % self.config
+
+        logger.debug("Job script: \n %s" % self.job_script())
```
guillaumeeb: We should probably debug log only the header here, but this is also true for PBS or SLURM, and it is really a minor detail...

guillaumeeb:
And about your other remarks:
* I totally agree to remove the PBS_ACCOUNT part and according test,
* For adaptive, you're completely right about the scale_down method, don't know though if it also breaks the scale_up. If you started with lets say only 1 worker, the scale_up should probably be called at some point, and not the scale_down, so something should happen. But yeah, we should look elsewhere!

dask_jobqueue/core.py
lesteve: In SGE, by default qsub returns quite a verbose output, e.g. something like `Your job 56 ("test.sh") has been submitted`. In order for just the job id to be returned, you need to use `qsub -terse`. This is why `submit_cmd = 'qsub -terse'`.

That means that you need to split `submit_cmd`. I think `shlex.split` is the way to do it for sh commands. We could just do `str.split but it may break e.g. if one of the arguments is quoted with a space inside.

dask_jobqueue/sge.py
lesteve: Good catch, done.

dask_jobqueue/sge.py
lesteve: Done.

lesteve:
@guillaumeeb thanks for the review! I tackled your comments, let me know if you have some further comments!

dask_jobqueue/sge.py
guillaumeeb: I believe the sge.py file still misses some import and annotation, e.g. 
https://github.com/dask/dask-jobqueue/blob/2e406ae462b5c82fa679075e45bb2bfa59615610/dask_jobqueue/pbs.py#L5
and
https://github.com/dask/dask-jobqueue/blob/2e406ae462b5c82fa679075e45bb2bfa59615610/dask_jobqueue/pbs.py#L10

You also still have the description of interface kw in the docstring!

lesteve: Ah right I fixed that, I am not so familiar with the docrep magic!

dask_jobqueue/sge.py
```diff
@@ -0,0 +1,83 @@
...
+        Destination queue for each worker job. Passed to `#$ -q` option.
+    project : str
+        Accounting string associated with each worker job. Passed to
+        `#$ -A` option.
+    resource_spec : str
+        Request resources and specify job placement. Passed to `#$ -l`
+        option.
+    walltime : str
+        Walltime for each worker job.
+    interface : str
```
guillaumeeb: You still have the interface keyword in this docstring, once remove I think we're done :)

lesteve: Arrggh good point, I just pushed the fix.

dask_jobqueue/core.py
```diff
@@ -153,8 +154,8 @@ def start_workers(self, n=1):
...
+                job = out.decode().split('.')[0].strip()
```
lesteve: And by the way the `.strip()` here is necessary because `qsub -terse` output finishes with a `\n`.

jhamman:
Thanks @lesteve!

mrocklin:
Thanks @lesteve!

lesteve:
Nice to see this merged! I'll try to find some time to fix the adaptive problem.

PR #29, opened 2018-04-05
Basic unit tests

guillaumeeb:
I've added some simple unit tests (testing that the worker job scripts was well formed) in order to have basic regression tests.
Those tests are also run into a Docker container, but no workers are launched, only constructor are tested, but basic methods could also be tested.

I was wondering if we still needed the conda install part in the .travis.yml:
```
  # Install miniconda
  - ./ci/conda_setup.sh
  - export PATH="$HOME/miniconda/bin:$PATH"
- conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION dask distributed flake8
```
If everything is done another time in Docker images, is there a need to do this in the base Travis envrionment?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ matrix:
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
+    - python: "3.6"
+      env:
+        - OS=ubuntu-14.04
+        - JOBQUEUE=basic
 env:
   global:
     - DOCKER_COMPOSE_VERSION=1.6.0
--- a/dask_jobqueue/tests/test_pbs.py
+++ b/dask_jobqueue/tests/test_pbs.py
@@ -7,6 +7,7 @@ from dask.distributed import Client
 from distributed.utils_test import loop  # noqa: F401
 from dask_jobqueue import PBSCluster
 
+#Run test only if py.test -E pbs is called
 pytestmark = pytest.mark.env("pbs")
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ log
 *.swp
 *.swo
 .cache/
+.pytest_cache
--- a/ci/basic.sh
+++ b/ci/basic.sh
@@ -7,8 +7,8 @@ function jobqueue_before_install {
     docker-compose version
 
     # just build and start a container with python dependencies
-    docker build -t dask-jobqueue:latest basic/
-    docker run -d --name test_server -v `pwd`../..:/dask-jobqueue dask-jobqueue:latest sleep 600
+    docker build -t dask-jobqueue:latest ci/basic/
+    docker run -d --name test_server -v `pwd`:/dask-jobqueue dask-jobqueue:latest sleep 600
 
     docker ps -a
     docker images
--- a/ci/basic.sh
+++ b/ci/basic.sh
@@ -23,5 +23,5 @@ function jobqueue_script {
 }
 
 function jobqueue_after_success {
-
+    echo "Hurrah"
 }
--- a/ci/basic.sh
+++ b/ci/basic.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -x
+
+function jobqueue_before_install {
+    docker version
+    docker-compose version
+
+    # start a container
+    docker build -t dask-jobqueue:latest basic/
+    docker run -d --name test_server -v `pwd`../..:/dask-jobqueue dask-jobqueue:latest sleep 600
+
+    docker ps -a
+    docker images
+}
+
+function jobqueue_install {
+    docker exec -it test_server /bin/bash -c "cd /dask-jobqueue; python setup.py install"
+}
+
+function jobqueue_script {
+    docker exec -it test_server /bin/bash -c "cd /dask-jobqueue; py.test dask_jobqueue --verbose"
+}
+
+function jobqueue_after_success {
+
+}
--- a/ci/basic.sh
+++ b/ci/basic.sh
@@ -6,7 +6,7 @@ function jobqueue_before_install {
     docker version
     docker-compose version
 
-    # start a container
+    # just build and start a container with python dependencies
     docker build -t dask-jobqueue:latest basic/
     docker run -d --name test_server -v `pwd`../..:/dask-jobqueue dask-jobqueue:latest sleep 600
 
--- a/ci/basic/Dockerfile
+++ b/ci/basic/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu:14.04
+
+ENV LANG C.UTF-8
+
+RUN apt-get update && apt-get install curl bzip2 git gcc -y --fix-missing
+
+RUN curl -o miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
+    bash miniconda.sh -f -b -p /opt/anaconda && \
+    /opt/anaconda/bin/conda clean -tipy && \
+    rm -f miniconda.sh
+ENV PATH /opt/anaconda/bin:$PATH
+RUN conda install -n root conda=4.4.11 && conda clean -tipy
+RUN conda install -c conda-forge dask distributed blas pytest mock ipython pip psutil && conda clean -tipy
+
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -1,4 +1,15 @@
+import pytest
+
+from dask_jobqueue import JobQueueCluster
+
 
 def test_jq_core_placeholder():
     # to test that CI is working
     pass
+
+
+def test_errors():
+    with pytest.raises(NotImplementedError) as info:
+        JobQueueCluster()
+
+    assert 'abstract class' in str(info.value)
--- a/dask_jobqueue/tests/test_pbs_basic.py
+++ b/dask_jobqueue/tests/test_pbs_basic.py
@@ -1,0 +1,71 @@
+from dask_jobqueue import PBSCluster
+
+#No mark in order to run these simple unit tests all the time
+
+
+def test_header():
+    with PBSCluster(walltime='00:02:00', processes=4, threads=2, memory='7GB') as cluster:
+
+        assert '#PBS' in cluster.job_header
+        assert '#PBS -N dask-worker' in cluster.job_header
+        assert '#PBS -l select=1:ncpus=8:mem=27GB' in cluster.job_header
+        assert '#PBS -l walltime=00:02:00' in cluster.job_header
+        assert '#PBS -q' not in cluster.job_header
+        assert '#PBS -A' not in cluster.job_header
+
+    with PBSCluster(queue='regular', project='DaskOnPBS', processes=4, threads=2, memory='7GB',
+                    resource_spec='select=1:ncpus=24:mem=100GB') as cluster:
+
+        assert '#PBS -q regular' in cluster.job_header
+        assert '#PBS -N dask-worker' in cluster.job_header
+        assert '#PBS -l select=1:ncpus=24:mem=100GB' in cluster.job_header
+        assert '#PBS -l select=1:ncpus=8:mem=27GB' not in cluster.job_header
+        assert '#PBS -l walltime=' in cluster.job_header
+        assert '#PBS -A DaskOnPBS' in cluster.job_header
+
+    with PBSCluster() as cluster:
+
+        assert '#PBS -j oe' not in cluster.job_header
+        assert '#PBS -N' in cluster.job_header
+        assert '#PBS -l select=1:ncpus=' in cluster.job_header
+        assert '#PBS -l walltime=' in cluster.job_header
+        assert '#PBS -A' not in cluster.job_header
+        assert '#PBS -q' not in cluster.job_header
+
+    with PBSCluster(job_extra=['-j oe']) as cluster:
+
+        assert '#PBS -j oe' in cluster.job_header
+        assert '#PBS -N' in cluster.job_header
+        assert '#PBS -l select=1:ncpus=' in cluster.job_header
+        assert '#PBS -l walltime=' in cluster.job_header
+        assert '#PBS -A' not in cluster.job_header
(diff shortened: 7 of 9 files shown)


guillaumeeb:
ping @jhamman for review and comment.

lesteve:
This looks like a very welcome addition. I would say that using docker is overkill for the basic tests and I would stick to conda. I would only keep docker for the case where we need to create a toy cluster.

Also I would be in favour of having a single file test_pbs.py, I haven't tested it but I am fairly confident you can decorate tests that need a toy cluster.

guillaumeeb:
I agree that docker is overkill, that's what I felt implementing it, I can remove that and stick with conda only, which justifies its deployment in Travis install.
I will look at using the pytest.mark as a decorator for tests that need an actual cluster only!

Thanks for the feedback.

mrocklin:
You might also consider using a fixture

```python
@pytest.fixture
def cluster():
    if cluster_not_available:
        pytest.skip("Cluster not available")

def test_my_test(cluster):
    ...
```

guillaumeeb:
I've taken into account remarks. This looks fine to me, my only concern is that we are installing conda in most cases just to do a flake8 check, except when launching the basic tests I wrote (JOBQUEUE=none). We could do all that directly into the "none" test environment, but I'm not sure it is better, in this case, the travis.yml file will look cryptic.

Also @lesteve, I've got a weird error on my last commit for sge in test_basic. Looks like a kind of timeout? Could you check that?

lesteve:
> Also @lesteve, I've got a weird error on my last commit for sge in test_basic. Looks like a kind of timeout? Could you check that?

Not sure about this one, it looks like the workers are not closed quickly enough. I managed to reproduce it locally once in ~50 test runs (`test_basic` generally finishes in 4-5 seconds on my machine). For now I think we should keep the test as it was (with a timeout of 10 seconds) and we should look into this in more details in case it keeps happening.

.travis.yml
```diff
@@ -17,18 +17,21 @@ matrix:
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
+    - python: "3.6"
```
lesteve: I don't think you need to add a build entry for now since the tests without an environment marker will be run in all the builds as you can see from this [log](https://travis-ci.org/dask/dask-jobqueue/jobs/367037400#L2974) (some functions in test_pbs.py are run even if `JOBQUEUE=sge`).

dask_jobqueue/tests/test_pbs.py
```diff
@@ -7,10 +6,80 @@
...
+        assert '#PBS -l select=1:ncpus=8:mem=27GB' not in job_script
+        assert '#PBS -l walltime=' in job_script
+        assert '#PBS -A DaskOnPBS' in job_script
+
+        assert '/dask-worker tcp://' in job_script
+        assert '--nthreads 2 --nprocs 4 --memory-limit 7GB' in job_script
+
+
+#Run test only if py.test -E pbs is called, ie. when a docker compose PBS cluster is launched
+@pytest.mark.env("pbs")  # noqa: F811
```
lesteve: I think you can remove all the `noqa: F811` from all the files. They were for unused `loop` variables and I believe now `loop` is used everywhere.

By the way @mrocklin is adding `loop` to all the test functions a good general guideline (which what was done before)? I have to admit I don't understand this kind of subtleties yet.

mrocklin: We do this in normal dask.distributed tests that use the normal/synchronous/blocking API.  The client is given this IOLoop explicitly and we clean it up after the test.  I honestly don't know if we're hooking everything together so that the IOLoops started by `*Cluster` are the same as the IOLoop that we hand in with the `loop` context manager.

We probably don't need to worry about this for now.  If at some point we run into odd conflicts or run out of file descriptors then this is something to come back to.

More information on common patterns when writing dask.distributed tests here: http://distributed.readthedocs.io/en/latest/develop.html#writing-tests

.travis.yml
guillaumeeb: You are right, I was honestly not aware of this. However, I like the idea to have a simple test which does not depend on docker or on starting toy clusters. This could help to identify basic problems, don't you think so?

dask_jobqueue/tests/test_pbs.py
guillaumeeb: From my tests and my understanding, the #noqa comment is still mandatory as `loop` is an import used as an arg name, which flake8 does not like much.

I did not used the loop parameters inside the basic test_header tests as they are not using a Tornado Loop. I understand from @lesteve comment and @mrocklin answer that it is the good choice.

lesteve: > From my tests and my understanding, the #noqa comment is still mandatory as loop is an import used as an arg name, which flake8 does not like much.

My bad I thought this was weird that the `noqa` was one the line on the `pytest.mark.env` and not on the line of the `def test_whatever(loop)` but it turns out this is how to make flake8 happy indeed.

.travis.yml
lesteve: OK fine with me.

guillaumeeb:
So with this I believe that #35 (and #37) are fixed.

guillaumeeb:
@lesteve  timeout problem again for sge test ending.

lesteve:
@guillaumeeb I am afraid I could not avoid to put my maintainer cap for a bit so I opened #39 based on your changes about only the Python 2 related fixes. Don't hesitate to comment on it!

In general I think keeping PRs focussed on solving a single well-defined issue is the best way to get them merged quickly. Maybe this is because I am used to more mature (aka conservative) projects like scikit-learn, so I am more than open to criticism about it.

lesteve:
@guillaumeeb you will need t fix conflicts since #39 was merged, sorry about that!

guillaumeeb:
@lesteve ok, no problem, but probably not until this week-end.

guillaumeeb:
So @lesteve, I see you did not like my comments above the docrep workaround :slightly_smiling_face: ? 

Just asking for future changes, isn't it a good practice to comment why we made some changes and linked that to a given issue? I like to comment a lot of things, but it is maybe a Java habit, or industrial project one...

lesteve:
> So @lesteve, I see you did not like my comments above the docrep workaround slightly_smiling_face ?

I think what is probably happening is that my threshold to deem that a comment is useful is probably higher than yours ... I would try to improve the code when possible rather than add a comment (e.g. general code organization, naming of function/variable etc ...). I tend to use comments to explain why we do it a certain way that may not be completely obvious at first. This can happen for some optimized code for example: when only glancing at the code it is very tempting to simplify the code, all the tests will very likely pass but you will "silently" lose some performance.

In my experience a lot of comments start being semi-useful (maybe as a kind of note to self in case I am ever confused by this again) and end up very quickly being completely useless and even misleading in some cases because it becomes out of sync with the code.

I'll comment inline to give a few examples in this PR where I personally would not have used any comments. This is only my personal opinion others may disagree.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -7,10 +6,80 @@
 
 from dask_jobqueue import PBSCluster
 
-pytestmark = pytest.mark.env("pbs")
 
+#No default mark in order to run these simple unit tests all the time
```
lesteve: About comments: I would remove this kind of comments too. Maybe this behaviour of pytest can be surprising at first but for someone who has heard of pytest mark I think it makes sense. Also would you expect anyone that adds a test function to also put this comment before the new function?

Note: I would remove similar comments in test_slurm.py too.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -7,10 +6,80 @@
...
+        assert '#PBS -l select=1:ncpus=24:mem=100GB' in job_script
+        assert '#PBS -l select=1:ncpus=8:mem=27GB' not in job_script
+        assert '#PBS -l walltime=' in job_script
+        assert '#PBS -A DaskOnPBS' in job_script
+
+        assert '/dask-worker tcp://' in job_script
+        assert '--nthreads 2 --nprocs 4 --memory-limit 7GB' in job_script
+
+
+#Run test only if py.test -E pbs is called, ie. when a docker compose PBS cluster is launched
```
lesteve: About comments: I would remove this one for the same reasons.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -34,7 +103,9 @@ def test_basic(loop):  # noqa: F811
             assert not cluster.jobs
 
 
-def test_adaptive(loop):  # noqa: F811
+#Run test only if py.test -E pbs is called, ie. when a docker compose PBS cluster is launched
```
lesteve: About comments: I would remove this one too.

.travis.yml
```diff
@@ -17,18 +17,26 @@ matrix:
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
+    - python: "2.7"
+      env:
+        #For launching only basic unit tests
```
lesteve: About comments: I don't know about this one. Maybe there could be a comment that explains that "JOBQUEUE=none is for tests that do not need a cluster to run". Maybe also put it just before JOBQUEUE=none to make it more obvious that it is about the next line. Alternatively we could have find a better naming than `none` but I don't find any off the top of my head ...

Python bias related nitpick: I would put a space after the `#` ...

.travis.yml
```diff
@@ -17,18 +17,26 @@ matrix:
...
 before_install:
   - pwd
-  - docker version
-  - docker-compose version
-  # Install miniconda
+  # Install miniconda (only needed for basic unit tests)
   - ./ci/conda_setup.sh
   - export PATH="$HOME/miniconda/bin:$PATH"
-  - conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION dask distributed flake8
+  - conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION dask distributed flake8 pytest docrep
```
lesteve: This can be done in another PR but as you mentioned before, you may not need conda at all for `JOBQUEUE=sge`, so you could actually do this:
* move miniconda setup + conda install to none.sh `jobqueue_before_install`
* move flake8 to none.sh `jobqueue_script` or maybe leave it in .travis.yml  but only run it when `[[ "$JOBQUEUE" == "none" ]].  IIRC Travis will run all the commands in the `script` stage even if one command fails which is handy because you generally want to have all the failures at once.

dask_jobqueue/pbs.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import division
```
lesteve: You can remove all the future imports from this PR because they are in `__init__.py` already.

.travis.yml
```diff
@@ -38,6 +46,7 @@ install:
   - jobqueue_install
 script:
   - flake8 -j auto dask_jobqueue
+  #Do not run py.test outside and inside the docker image, this will fail, so rnu basic tests in "none" jobqueue script
```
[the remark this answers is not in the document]
guillaumeeb: Okay, this is disturbing to look at my comment and see it is so badly written it gets confusing. So you have a point here. But I like your rephrasing, and find it useful. At least I've worked in environments where is would be useful.

ci/none.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -x
+
+function jobqueue_before_install {
+  # Nothing to go
```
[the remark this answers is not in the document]
guillaumeeb: Yes, another point for you.

dask_jobqueue/tests/test_pbs.py
guillaumeeb: Okay I get it, I talk too much for myself in the code 😄

dask_jobqueue/tests/test_pbs.py
guillaumeeb: I kind of like this one, It can helps newcommers like me to understand the code more easily, don't you think?

.travis.yml
guillaumeeb: Finaly, a semi-useful one 😁 !

dask_jobqueue/pbs.py
guillaumeeb: Okay, will look at that and your other comments. Thanks!

dask_jobqueue/tests/test_pbs.py
guillaumeeb: Okay, after checking the code, it is useless.

.travis.yml
guillaumeeb: I'm still not sure about that. I know this is useless, but I like the idea of the flake8 checks running all the time... Don't know why.
But installing conda only for this while these checks are done on the "none" environement seems useless. Another PR?

dask_jobqueue/pbs.py
guillaumeeb: So I removed import __future__ from __init__.py, and put it in the files where it was needed.

.travis.yml
guillaumeeb: Finally when writing the above comment, I decided to implement it here, because keeping the conda install and flake8 check in every build was nonsense.

lesteve:
@guillaumeeb this looks good, thanks for making the extra effort of cleaning up and moving the  conda install to none.sh!

@jhamman or @mrocklin maybe you can have a quick look if you have some spare bandwidth? To sum up:
* this reorganizes the tests a bit so that you can run basic tests without having a cluster. For example, anyone can just run `pytest dask_jobqueue` and make sure nothing is badly broken. 
* if on top of the basic tests, you want to run the tests for a given scheduler you can still use `pytest -E your_cluster_name` as before. For example, `pytest -E sge` is run in Travis.
* JOBQUEUE=none is used in .travis.yml to run only the basic tests + flake8.

mrocklin:
This looks great to me.  I really like the approach.

mrocklin:
In particular I like that we can include both infrastructure-requiring and simple tests in the same testing files.  It seems easy for future developers to use intuitively.

lesteve:
OK thanks for your feedback @mrocklin! I am going to merge this one and we can always iterate and improve the approach later.

Thanks a lot for working on this @guillaumeeb!

PR #34, opened 2018-04-14
Adding basic sphinx docs

jhamman:


jhamman:
I've also setup a read the docs page: https://readthedocs.org/projects/dask-jobqueue

At this point, I'll need to merge this before the docs can be built for the first time.

mrocklin:
We should also include some subset of this in the core dask documentation.

mrocklin:
I can also do that at some point though tomorrow.

jhamman:
Docs are live here: http://dask-jobqueue.readthedocs.io/en/latest/index.html

I'm not 100% sure if the webhook is working yet. It may be that I don't have enough permissions to fully set that up.  @mrocklin - I made you an ower/admin of the readthedocs account as well.

> We should also include some subset of this in the core dask documentation.

That would be great.

guillaumeeb:
Thanks for all that, that's great!

PR #39, opened 2018-04-18
Fix python2 problems

lesteve:
Took the Python 2-related part of #29 with some minor tweaks.

@guillaumeeb let me know if I forgot anything!

### DIFF
--- a/dask_jobqueue/pbs.py
+++ b/dask_jobqueue/pbs.py
@@ -7,9 +7,9 @@ from .core import JobQueueCluster, docstrings
 logger = logging.getLogger(__name__)
 
 
-@docstrings.with_indent(4)
 class PBSCluster(JobQueueCluster):
-    """ Launch Dask on a PBS cluster
+    #Defining docstring this way for python 2 and docrep compatibility. See #35.
+    __doc__ = docstrings.with_indents(""" Launch Dask on a PBS cluster
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


lesteve:
So the CI is not going to tell us anything about the Python 2 fixes but I feel this is better to fix our Python 2 testing in a separate PR.

dask_jobqueue/pbs.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import division
+
```
guillaumeeb: I use some division in slurm.py too. You took it in pbs.py, but not in slurm.py, is this intended?
I can see you've imported that also in the __init__.py file, I'm not sure what it does, and if it is enough for both pbs and slurm then (not coming from pythonic world).

lesteve:
So annoyingly test_sge.test_basic seems to fail quite consistently ... @guillaumeeb in your experience did you find that increasing the timeout helps?

guillaumeeb:
I would not talk about experience, but in my unique try with travis and a timeout of 20 seconds, tests succeeded...
I know that is not particularly helpful.

lesteve:
OK that is the kind of thing I wanted to hear about, so that is still helpful thanks!

lesteve:
An alternative is to wait for https://github.com/Chilipp/docrep/pull/9 to be merged.

I don't feel strongly about how much we care about class docstrings correctness for Python 2.

guillaumeeb:
You're a lot more experienced that I am in this area. I don't know what might be the consequences of non standard docstring use like in pbs.py...

mrocklin:
I ran into this in practice with someone who was using Python 2.  I noticed that they were using some of the parameters to the PBSCluster but not others (they were missing threads, processes, ...)  We pulled up the docstring and I noticed that they weren't included.

A low-tech but maybe sufficient solution to this is just to drop docrep and copy-paste everywhere.  This might not be terrible.

lesteve:
I have to say, I am a bit undecided about `docrep` ... but let's keep this PR about Python 2. I am going to merge this one.

PR #51, opened 2018-05-02
Remove unused SGE run_* files

lesteve:
Looks like they were unused copies of `run-*` files.


lesteve:
CIs are green, merging!

PR #54, opened 2018-05-07
Use released version of distributed.

lesteve:
At the time of writing you need to have both dask and distributed master because of the config overhaul (https://github.com/dask/distributed/pull/1948 and https://github.com/dask/dask/pull/3432). If you only have distributed from master you'll have an error like this:

```
❯ python -c 'from distributed import client'                    
  File "<string>", line 1, in <module>
  File "/home/lesteve/miniconda3/envs/tmp/lib/python3.6/site-packages/distributed/__init__.py", line 3, in <module>
    from . import config
  File "/home/lesteve/miniconda3/envs/tmp/lib/python3.6/site-packages/distributed/config.py", line 13, in <module>
    config = dask.config.config
AttributeError: module 'dask' has no attribute 'config'
```

Also this seems like a better idea to test against released version of distributed
in our CI, as the end user is very likely going to be using a released version of distributed.

Given that there was a relase of distributed 4 days ago, I reckon this is recent enough. We can always revert to testing against distributed if we need bleeding-edge features but I would rather fix master first.

### DIFF
--- a/ci/sge/Dockerfile-master
+++ b/ci/sge/Dockerfile-master
@@ -12,7 +12,6 @@ ENV PATH /opt/anaconda/bin:$PATH
 RUN conda install -n root conda=4.4.11 && conda clean -tipy
 RUN conda install -c conda-forge dask distributed blas pytest mock ipython pip psutil && conda clean -tipy
 RUN pip install --no-cache-dir drmaa
-RUN pip install --no-cache-dir git+https://github.com/dask/distributed.git --upgrade
 
 COPY ./*.sh /
 COPY ./*.txt /
--- a/ci/sge/Dockerfile-slave
+++ b/ci/sge/Dockerfile-slave
@@ -12,7 +12,6 @@ ENV PATH /opt/anaconda/bin:$PATH
 RUN conda install -n root conda=4.4.11 && conda clean -tipy
 RUN conda install -c conda-forge dask distributed blas pytest mock ipython pip psutil && conda clean -tipy
 RUN pip install --no-cache-dir drmaa
-RUN pip install --no-cache-dir git+https://github.com/dask/distributed.git --upgrade
 
 COPY ./setup-slave.sh /
 COPY ./*.sh /


mrocklin:
See also https://github.com/dask/dask-jobqueue/pull/48#issuecomment-386735541

guillaumeeb:
So this seems to be in conflict with current @mrocklin proposal in #48.
I apparently had to use distributed master in the PR about PBS Docker CI, see comment https://github.com/dask/dask-jobqueue/pull/47#issuecomment-385145779.

I'm not against having dependency against master version of distributed or dask for now, as dask-jobqueue is quite new. In the long run, I would recommend as @lesteve to be dependant and test against released versions.

lesteve:
Right I missed this about #48. To be honest, I don't have a strong opinion about this, I would just like master to be fixed.

For some context: I was trying #47 locally and bumped into this problem.

mrocklin:
I think that this PR is correct and that we should go ahead and merge regardless of what happens in #48 .  That discussion could take a while and it shouldn't blog progress in the meantime.  Getting in CI for PBS seems higher priority to me than the config changes.

That being said, I'd love it if people would try out the configuration system and see if it is in an improvement or not.  The more quickly we get feedback like this from downstream projects like dask-jobqueue and dask-kubernetes (see equivalent PR there: https://github.com/dask/dask-kubernetes/pull/75) the more quickly we can solidify things into a release.

guillaumeeb:
And @lesteve, with this fix don't you run into this kind of failure: https://travis-ci.org/dask/dask-jobqueue/jobs/372327569.
This is why I had to stick with distributed master branch in #47.

lesteve:
> And @lesteve, with this fix don't you run into this kind of failure: https://travis-ci.org/dask/dask-jobqueue/jobs/372327569.

The latest distributed release is 4 days old. I am betting on the fact that the bug-fix that was in master when you were working on #47 has been included in the release.

guillaumeeb:
> The latest distributed release is 4 days old. I am betting on the fact that the bug-fix that was in master when you were working on #47 has been included in the release.

You are absolutely right.

PR #59, opened 2018-05-16
fix typo in Slurm walltime

rsignell-usgs:
This typo in wall time (1 minute instead of 1 hour) was my fault, fixing here.


mrocklin:
Thanks for the catch @rsignell-usgs !

Issue #30, opened 2018-04-08
Dask client connects to PBS workers, then rapidly loses them


apatlpo:
The title is intentionally analogous to #20 as I have the feeling the explanation for the observed behavior is similar.

I'm on a PBS cluster whose nodes are made of 2 cpus with 14 cores each.

I was initially calling:
```
cluster = PBSCluster(queue='mpi_1', local_directory=local_dir, interface='ib0', walltime='24:00:00',
                     threads=4, processes=7, memory='10GB', resource_spec='select=1:ncpus=28:mem=100g', 
                     death_timeout=100)
```

This led to the creation of workers but they died after creation.

The following choice seems to fix the issue:
```
threads=14, processes=2, memory='50GB', 
```

Here is a link that describes dask workers:
http://distributed.readthedocs.io/en/latest/worker.html
this may be useful to readers having similar issues

Note that the link between cluster architecture and options that can be passed to PBSCluster is still not entirely clear to me.

So my issue seems to be fixed, but I wanted to put this experience visible to people that may encounter similar issues.

guillaumeeb:
The only difference I see between the two choices is that the second one will only spawn 2 linux processes where the first one tries to spawn 7. Could you ask your cluster admin if there is some limit on the number of spawned processes on the queue or the servers belonging to the queue you are using?

Another thing will be to try to launch worker manually and see what happens...

apatlpo:
Here is the reply that I got from admins:

```
hi aurelien,

spawn means with mpi spawning?

tina

Le 16/04/2018 à 15:55, Aurelien Ponte a écrit :
> Tina,
> Here is there some limit to the number of spawned processes on mpi_1 ?
> thanks
> aurélien 
```

How can I described what type of processes jobqueue is spawning?

guillaumeeb:
I'm not really confident that this could be the issue, but here is what append:
* When starting with 2 processes, 14 threads, there will be 2 linux processes launched in your job, 2 processes running python, using up to 14 cpus with threads.
* With 7 processes: 7 independent linux processes, so things you could see using top or ps. There may potentially be a limit here.

The other thing is that this module does not use MPI at all. The queue you are using seems dedicated to MPI jobs, I don't know if there may be some other things preventing not MPI processes to run well?

jhamman:
@apatlpo - if you have some time, it may be worth giving jobqueue another go. We've had a number of new features land in recent weeks and you may find things easier to work.

apatlpo:
[...]

Issue #11, opened 2018-03-11
Implement a cluster status method, to now if worker are really running


guillaumeeb:
Currently, we only know that we have submitted some jobs to a cluster scheduler. We don't know if these jobs are running or queued, or in any other state.

What do you think of implementing a kind of status method?

In the PBS case for example, it would issue a qstat call, and get the PBS scheduler status of every jobs handled by the Dask cluster.

Not sure if this is really needed, as we are able with the use of Dask Client to know the real size of the cluster (and maybe by some other means). 

Maybe this issue is just about documenting how to retrieve the information about a cluster state, either by job scheduler (eg. PBS) API, either using Dask API.

jhamman:
I think this would be useful when using an autoscaling cluster. Particularly when jobs are waiting in the queue and the adaptive cluster is trying to decide if it should scale up/down.

mrocklin:
I agree that understanding the status would be useful, especially for
things like adaptive scheduling.

jhamman:
Is the following approach a dependable way of querying the cluster for the number of workers?

```Python
min_workers = 10  # for example
while len(cluster.workers) < min_workers:
    sleep(1)
```

mrocklin:
I believe that cluster.workers is only a convention at this point.  It's managed by the cluster object and so represents workers that have been asked for, but not workers that have connected.  For the latter you would want the following:

```python
cluster.scheduler.workers
```

jhamman:
@rabernat - most of the hard work for this functionality is being done in #63. Stay tuned.

andersy005:
This appears to have been addressed:

```python
In [15]: import dask_jobqueue

In [16]: cluster = dask_jobqueue.SLURMCluster(cores=1, processes=1)

In [17]: cluster.worker_spec
Out[17]: {}

In [18]: cluster.workers
Out[18]: {}

In [19]: cluster.scale(2)

In [20]: cluster.workers
Out[20]: {0: <SLURMJob: status=running>, 1: <SLURMJob: status=running>}
```


However, it appears that **the reported information can be misleading sometimes**. For instance, I am using a system which restricts the number of concurrent jobs to 35. When I submit 40 jobs, `dask_jobqueue` reports that all 40 workers are in `running` state even when some of the jobs are in `pending` state according to `squeue`:


- `dask-jobqueue` reports all 40 workers to be in `running` state:

```python
In [22]: cluster.scale(40)

In [23]: cluster.workers
Out[23]:
[...]

PR #63, opened 2018-05-18
Fixes for Adaptive

jhamman:
I'm just starting to work on a clearer naming convention for individual `Worker`s coming from jobqueue. The current behavior is to pass the following string to the `--name` argument of each `dask-worker` call:

```
{JOB_NAME}-{JOB_NUM}
```

and the worker number for each job is appended so we actually end up with: 

- `JOB_NAME` is the name argument given to the Cluster (e.g. 'dask-worker')
- `JOB_NUM` is an integer the job count submitted from each cluster (starting at 1)

```
{JOB_NAME}-{JOB_NUM}-{WORKER_NUM}
```

- `WORKER_NUM` is an integer assigned by distributed when using grouped workers

I'm proposing we add the `JOB_ID` and consider dropping the `JOB_NUM`. So we end up with either:

```
{JOB_NAME}-{JOB_ID}-{JOB_NUM}-{WORKER_NUM}
# or 
{JOB_NAME}-{JOB_ID}-{WORKER_NUM}
```

We could also drop the `JOB_NAME` but maybe that is a step too far.

Edit 6/14/2018:

We ended up with the following name: 

```
{JOB_PREFIX}--{JOB_ID}--[{WORKER_NUM}]
```

Worker num is optional but job prefix and job id are required in all cases. 
-------

This PR includes some small changes to add the `JOB_ID` to the naming convention. Ultimately, this will allow us to create logical mappings from job ids to workers and that will hopefully help elleviate some of the problems like #26 and #30.


dask_jobqueue/core.py
```diff
@@ -232,10 +239,13 @@ def scale_up(self, n, **kwargs):
...
-            names = {v['name'] for v in workers.values()}
-            job_ids = {name.split('-')[-2] for name in names}
-            self.stop_workers(job_ids)
+        if not isinstance(workers, dict):
+            raise ValueError(
+                'Expected dictionary of workers, got %s' % type(workers))
+        names = {v['name'] for v in workers.values()}
+        # This will close down the full group of workers
+        job_ids = {name.split('-')[-2] for name in names}
+        self.stop_workers(job_ids)
```
jhamman: I'm thinking there is a better way to do this. The current behavior to scale down removes the entire job from the system. So if `Adaptive` tells us to remove 1 worker (say we have 10 workers per job), we're going to remove all 10.

@mrocklin - Would it make sense to add logic to `Adaptive` so it knows how to bundle groups of workers? Otherwise, we could bundle here and check to see we're being asked to scale down an entire group.

dask_jobqueue/core.py
```diff
@@ -161,7 +167,8 @@ def job_file(self):
     def start_workers(self, n=1):
         """ Start workers and point them to our local scheduler """
         workers = []
-        for _ in range(n):
+        num_jobs = min(1, math.ceil(n / self.worker_processes))
+        for _ in range(num_jobs):
```
jhamman: This is a breaking change I want to make sure everyone is aware of. The current behavior for a hypothetical setup that includes 10 workers per job would be:

```Python
cluster.start_workers(1)
```

...and get 1 job and 10 workers. 

I'd like to change this so that `start_workers(n)` gives us `n` workers and as many jobs as needed to make that happen.

dask_jobqueue/core.py
mrocklin: The key= parameter to `workers_to_close` (passed through from `retire_workers`) seems relevant here.  I believe that it was made for this purpose.

https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L2525-L2548

dask_jobqueue/core.py
mrocklin: Historically start_workers was a semi-convention between a few projects.  This has decayed, so I have no strong thoughts here.  I do think that we need to be consistent on `scale` though, which seems a bit more standard today.

dask_jobqueue/core.py
```diff
@@ -161,7 +168,8 @@ def job_file(self):
     def start_workers(self, n=1):
         """ Start workers and point them to our local scheduler """
         workers = []
-        for _ in range(n):
+        num_jobs = min(1, math.ceil(n / self.worker_processes))
```
guillaumeeb: Why using `min` here? This would always lead to only one job started if I'm not mistaken.

dask_jobqueue/core.py
guillaumeeb: Will this really help adaptive? Would'nt there still be a problem with starting the worker in a grouped manner?

With your example, calling `cluster.start_workers(1)` will still lead to 1 job and 10 workers!

But this may be well handled by adaptive, I don't know. In this case, this may not be needed to do this breaking change?

dask_jobqueue/core.py
guillaumeeb: Glad to see that grouped worker is handled in adaptive!

Another comment here, not linked to this PR, is that I find the `job_ids` var name misleading. Should be something like `worker_ids`.

dask_jobqueue/core.py
jhamman: good point. I've removed this.

dask_jobqueue/core.py
```diff
@@ -117,7 +122,6 @@ def __init__(self,
         self.worker_threads = threads
         self.name = name
 
-        self.jobs = dict()
```
jhamman: `self.jobs` was a mapping from n-->job_id. However, we were not really using it and it often was not cleaned up when a job ended (so I've removed it).

guillaumeeb: I believe we can get rid of `self.n` and thus `self.jobs` as it is right now. However, for adaptive to work correctly, I believe we should keep track of all submitted jobs and their statuses. If not, don't we risk to continuously submit new jobs that are kept in the scheduler queue?

I'm in favour of keeping a dict mapping job_id --> job-status, e.g as @mrocklin proposed in #11: 'pending', 'running', 'finished' or equivalent. This way, in the scale_up method, we can take that into account.

A maybe simpler solution is to only keep track of the number of workers that are pending or running, and at  least to use this number in scale_up:
`return self.start_workers(n - number_of_pending_or_running_worker)`
But it seems difficult to deal with finished, running and pending worker this way.

jhamman: @guillaumeeb - I think we're in agreement here.

>  we should keep track of all submitted jobs and their statuses.

This would be nice but it might be somewhat difficult to do. We have three/four states that a job might be in:

- Pending - we may be able to combine some form of `qstat job_identifier` with a dictionary of submitted jobs `self.jobs above`
- Running - it is straight forward to determine which workers are attached to the scheduler
- Finished - jobs can exit normally or be killed (e.g. exceeded wall time). When JobQueue culls a worker, its easy to remove that worker from the list of jobs. However, when the queuing system kills a worker, we would need a way to remove that job from the list of running jobs. 

Generally, I think any use of `qstat` is going to be a bit ugly just because repeated queries of the queueing system tend to be quite expensive. For example:

```
$ time qstat 9595055
Job id            Name             User              Time Use S Queue
----------------  ---------------- ----------------  -------- - -----
9595055.chadmin1  dask             jhamman           00:07:22 R premium

real	0m1.030s
user	0m0.012s
sys	0m0.000s
```

So we probably don't want to do this very often. Do others have ideas as to how one would track the status of these jobs in a tractable way?

mrocklin: We could add a plugin to the scheduler that watched for when workers started and compared them against a set of known pending workers: http://distributed.readthedocs.io/en/latest/plugins.html

Something like the following:

```python
from distributed.diagnostics.plugin import SchedulerPlugin

class JobQueuePlugin(SchedulerPlugin):
    def add_worker(self, scheduler, worker=None, name=None, **kwargs):
        job_id = parse(name)
        pending_jobs.remove(job_id)

scheduler.add_plugin(JobQueuePlugin())
```

jhamman: This sounds like what we need. I can implement this as part of this PR.

jhamman: See 92eaf4e for an initial (untested) implementation using the scheduler plugin approach.

dask_jobqueue/core.py
```diff
@@ -1,23 +1,73 @@
...
+            self.running_jobs[job_id] = self.pending_jobs.pop(job_id)
+        self.running_jobs[job_id].update(worker=worker, status='running')
+
+    def remove_worker(self, scheduler=None, worker=None, **kwargs):
+        job_id = _job_id_from_worker_name(worker.name)
+        if job_id not in self.finished_jobs:
+            self.finished_jobs[job_id] = self.running_jobs.pop(job_id)
+        self.finished_jobs[job_id].update(status='finished')
+        if self.finished_jobs[job_id].workers:
+            self.finished_jobs[job_id].workers = []
```
guillaumeeb: If I understand correctly, you are considering the workers are ending all at once, when the first worker corresponding to this job_id is removed. Perhaps we could remove workers one by one, just to be sure? This may be overkill.

dask_jobqueue/core.py
```diff
@@ -144,8 +205,8 @@ def __init__(self,
 
     def job_script(self):
         """ Construct a job submission script """
-        self.n += 1
-        template = self._command_template % {'n': self.n}
+        self._n += 1
+        template = self._command_template % {'n': self._n}
```
guillaumeeb: "-%(n)d" has been removed from self._command_template (l.137), so we don't need this line if I'm not mistaken.

dask_jobqueue/core.py
```diff
@@ -219,29 +278,31 @@ def stop_workers(self, workers):
...
         self._call([self.cancel_command] + list(jobs))
-        for w in workers:
-            with ignoring(KeyError):
-                del self.jobs[w]
 
     def scale_up(self, n, **kwargs):
         """ Brings total worker count up to ``n`` """
-        return self.start_workers(n - len(self.jobs))
+        pending_workers = self.worker_processes * len(self.pending_jobs)
+        active_and_pending = len(self.scheduler.workers) + pending_workers
```
guillaumeeb: A probably rare case, but you may miss starting workers here, when a job just began to start, so is moving from pending to running. We may have some worker process started for a given job_id, but not all of them.
Maybe it is safer to just rely on self.pending_jobs and self.running_jobs, but I'm not sure, we could also miss ending jobs 🙂 ...

dask_jobqueue/core.py
```diff
@@ -219,29 +278,31 @@ def stop_workers(self, workers):
...
+            except KeyError:
+                logger.debug('worker %s is already gone' % w)
+        self.stop_workers(workers)
 
     def __enter__(self):
         return self
 
     def __exit__(self, type, value, traceback):
-        self.stop_workers(self.jobs)
+        self.stop_workers(self.scheduler.workers)
```
guillaumeeb: Here, don't we need to also cancel pending jobs?

dask_jobqueue/tests/test_pbs.py
```diff
@@ -102,14 +102,62 @@ def test_basic(loop):
 
 @pytest.mark.env("pbs")  # noqa: F811
 def test_adaptive(loop):
-    with PBSCluster(walltime='00:02:00', processes=1, threads=2, memory='2GB', local_directory='/tmp',
-                    job_extra=['-V'], loop=loop) as cluster:
+    with PBSCluster(walltime='00:02:00', processes=1, threads=2, memory='2GB',
+                    local_directory='/tmp', ob_extra=['-V'],
```
guillaumeeb: typo in job_extra

dask_jobqueue/tests/test_pbs.py
```diff
@@ -102,14 +102,62 @@ def test_basic(loop):
...
+            start = time()
+            while cluster.pending_jobs or cluster.running_jobs:
+                sleep(0.100)
+                assert time() < start + 10
+
+
+@pytest.mark.env("pbs")  # noqa: F811
+def test_adaptive_grouped(loop):
+    with PBSCluster(walltime='00:02:00', processes=2, threads=1, memory='2GB',
+                    local_directory='/tmp', ob_extra=['-V'],
```
guillaumeeb: typo here too.

dask_jobqueue/core.py
jhamman: Good idea. This was a pretty easy fix so I have included it in 4776892.

jhamman:
This is ready for another round of reviews.

dask_jobqueue/core.py
```diff
@@ -110,15 +178,23 @@ def __init__(self,
...
 
+        # plugin for tracking job status
+        self._scheduler_plugin = JobQueuePlugin()
+        self.cluster.scheduler.add_plugin(self._scheduler_plugin)
+        self.pending_jobs = self._scheduler_plugin.pending_jobs
+        self.running_jobs = self._scheduler_plugin.running_jobs
+        self.finished_jobs = self._scheduler_plugin.finished_jobs
+
+        # counter to keep track of how many jobs have been submitted
+        self._n = 0
```
guillaumeeb: Do we still need to keep this counter?
I feel like we've got all the information we want in pending, running and finished jobs. We could even add some detailed status method wIth all that, maybe in another PR!

dask_jobqueue/core.py
```diff
@@ -1,23 +1,89 @@
...
+
+    def add_worker(self, scheduler, worker=None, name=None, **kwargs):
+        ''' Run when a new worker enters the cluster'''
+        w = scheduler.workers[worker]
+        job_id = _job_id_from_worker_name(w.name)
+
+        # if this is the first worker for this job, move job to running
+        if job_id not in self.running_jobs:
+            self.running_jobs[job_id] = self.pending_jobs.pop(job_id)
+            self.running_jobs[job_id].update(status='running')
```
mrocklin: Are there any implications of doing this part-way through a set of workers starting?

dask_jobqueue/core.py
```diff
@@ -1,23 +1,89 @@
...
+        ''' Run when a worker leaves the cluster'''
+        # the worker may have already been removed from the scheduler so we
+        # need to check in running_jobs for a job that has this worker
+        for job_id, job in self.running_jobs.items():
+            if worker in job.workers:
+                # remove worker from dict of workers on this job id
+                del self.running_jobs[job_id].workers[worker]
+                break
+        else:
+            raise ValueError('did not find a job that owned this worker')
```
mrocklin: This will also run if a worker just restarts, or is temporarily killed by a nanny.  We might not want to remove the job entirely here.

mrocklin: Ah, I see that we just remove one of the workers from the job.  I guess we add the worker back in when it starts back up in the `add_worker` function?

dask_jobqueue/core.py
```diff
@@ -110,15 +178,23 @@ def __init__(self,
 
         self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
 
+        # plugin for tracking job status
+        self._scheduler_plugin = JobQueuePlugin()
+        self.cluster.scheduler.add_plugin(self._scheduler_plugin)
+        self.pending_jobs = self._scheduler_plugin.pending_jobs
+        self.running_jobs = self._scheduler_plugin.running_jobs
+        self.finished_jobs = self._scheduler_plugin.finished_jobs
```
mrocklin: Perhaps these should be properties?  I'm not sure exactly why I'm suggesting this, but it seems like a more common pattern.

dask_jobqueue/core.py
```diff
@@ -133,8 +209,9 @@ def __init__(self,
         if memory is not None:
             self._command_template += " --memory-limit %s" % memory
         if name is not None:
-            self._command_template += " --name %s" % name
-            self._command_template += "-%(n)d" # Keep %(n) to be replaced later
+            # worker names follow this template: {NAME}-{JOB_ID}
+            self._command_template += " --name %s" % name  # e.g. "dask-worker"
+            self._command_template += "-${JOB_ID}"
```
mrocklin: '"Some preference to put this on one line

    " --name %s-${JOB_ID}" % name"

dask_jobqueue/core.py
```diff
@@ -219,29 +294,37 @@ def stop_workers(self, workers):
...
     def scale_down(self, workers):
         ''' Close the workers with the given addresses '''
-        if isinstance(workers, dict):
-            names = {v['name'] for v in workers.values()}
-            job_ids = {name.split('-')[-2] for name in names}
-            self.stop_workers(job_ids)
+        workers = []
+        for w in workers:
+            try:
+                # Get the actual "Worker"
```
mrocklin: I recommend removing the quotes here, and instead use the class name `WorkerState`

dask_jobqueue/core.py
jhamman: The status flag is really just for internal tracking. The pop from pending to running is the real state change here.

dask_jobqueue/core.py
```diff
@@ -219,29 +301,37 @@ def stop_workers(self, workers):
...
+        active_and_pending += self.worker_processes * len(self.pending_jobs)
+        return self.start_workers(n - active_and_pending)
 
     def scale_down(self, workers):
         ''' Close the workers with the given addresses '''
-        if isinstance(workers, dict):
-            names = {v['name'] for v in workers.values()}
-            job_ids = {name.split('-')[-2] for name in names}
-            self.stop_workers(job_ids)
+        workers = []
```
guillaumeeb: After first tests on a real cluster, I've detected a problem in scale_down. I had to look closely for several minutes before seeing the problem: the `workers` argument is erased by a local variable just below!

Scaling up works fine. Did not try adaptive.

As noted before, this is also a small breaking change with before. I need to get used to it, but I believe this is a good change. We will need to explain the behavior somewhere though. It can be weird to have our asked workers rouned up every time we ask for some more.

jhamman: Good catch. Not sure how this crept in.

dask_jobqueue/tests/test_slurm.py
```diff
@@ -6,9 +6,12 @@
 
 from dask_jobqueue import SLURMCluster
 
+pytestmark = pytest.mark.env("slurm")
```
lesteve: Do you need this here? IIRC the consensus was to add `pytest.mark.env` only for decorating functions that need a real cluster to run.

The same thing happens in `test_pbs.py`.

guillaumeeb:
Some more tests today. It's really good to be able to change cluster size only using the `scale` function.

I've observed just one small issue, and I have no advice how to deal with it: if you specify `processes` > 1, then calling `scale(1)` will either launch one job with several workers (which is expected), but on the next call cancel it (that I feel it would be better if it did'nt).

This is probably due to this code https://github.com/dask/distributed/blob/master/distributed/deploy/cluster.py#L93.

jhamman:
@guillaumeeb - thanks for sticking with the testing of this PR. I probably won't have time to get back to it until next week but I'd like to finish it up quickly thereafter.

Quickly, I'll just say that I haven't noticed the behavior you describe and that I'll look into it more when I return.

mrocklin:
@jhamman can I ask about your plans on this?  Is it important to get in before a next release?

jhamman:
I just need to clean up the failing tests. I think the implementation is basically complete. Not important to get into the next release but I thought we were blocked by the next dask/distributed releases.

mrocklin:
> Not important to get into the next release but I thought we were blocked by the next dask/distributed releases

I'm working towards those releases now.  Was loosely planning to coordinate releases with a couple of the downstream packages, notable dask-jobqueue and dask-kubernetes.

jhamman:
Okay, I'll try to wrap this up over the next few days.

guillaumeeb:
On this one, I believe it is important to have this in the next release. It is a breaking change, but I now believe it is a really positive one!

However I reiterate my comment of trying to explain it as clearly as possible in the docs. As english is clearly not my first language, I'm happy to give comment on this part too, but don't think I'm the more qualified to do it, sorry for this.

jhamman:
I am working on the final doc/test edits for this PR. Hopefully we can wrap it up soon and get the next release out ASAP.

dask_jobqueue/tests/test_slurm.py
```diff
@@ -8,7 +8,10 @@
 
 
 def test_header():
-    with SLURMCluster(walltime='00:02:00', processes=4, threads=2, memory='7GB') as cluster:
+    with SLURMCluster(walltime='00:02:00', processes=4, threads=2,
+                      memory='7GB') as cluster:
+
+        print(cluster.job_header)
```
mrocklin: Debug print statements?

docs/index.rst
```diff
@@ -25,7 +25,7 @@ Example
    from dask_jobqueue import PBSCluster
 
    cluster = PBSCluster(processes=6, threads=4, memory="16GB")
-   cluster.start_workers(10)
+   cluster.start_workers(12)  # this equates to 2 PBS jobs, each with 6 workers
```
mrocklin: This is a nicely informative comment!

dask_jobqueue/tests/test_slurm.py
jhamman: yes, I'm using CI to debug a few things at the moment. I'll ping folks here when I have things cleaned up.

mrocklin:
@jhamman anything I can help with here to move this along?

lesteve:
I would be interested in hearing more about what is called "grouped workers" in this PR to make sure I am not misunderstanding. 

* my naive assumption until very recently was that a worker is what you get when you execute `dask-worker` so that may be actually using more than one process under the hood but that is still one worker.
* quickly looking at this PR it feels that what you mean by "grouped workers" represents what you get when you execute `dask-worker` and that workers are the underlying worker processes

jhamman:
This is ready for a final review. 

@mrocklin - if you wouldn't mind looking at the last remaining failing test, I suspect its a test syntax issue but it would be good for you to take a look.

mrocklin:
Happy to take a look in a bit.  You say you suspect that it's a syntax
issue.  Is there anywhere in particular I should look?  Looking at the test
traceback didn't give me any clues.

dask_jobqueue/core.py
```diff
@@ -1,20 +1,95 @@
...
+    def add_worker(self, scheduler, worker=None, name=None, **kwargs):
+        ''' Run when a new worker enters the cluster'''
+        w = scheduler.workers[worker]
+        job_id = _job_id_from_worker_name(w.name)
+
+        # if this is the first worker for this job, move job to running
+        if job_id not in self.running_jobs:
+            if job_id not in self.pending_jobs:
+                raise KeyError(
+                    '%s not in pending jobs: %s' % (job_id, self.pending_jobs))
```
mrocklin: `self.pending_jobs` can be quite large here, right?  It might be unwise to include the entire thing in the exception.

dask_jobqueue/core.py
```diff
@@ -1,20 +1,95 @@
...
+            self.running_jobs[job_id].update(status='running')
+
+        # add worker to dict of workers in this job
+        self.running_jobs[job_id].update(worker=w)
+
+    def remove_worker(self, scheduler=None, worker=None, **kwargs):
+        ''' Run when a worker leaves the cluster'''
+        # the worker may have already been removed from the scheduler so we
+        # need to check in running_jobs for a job that has this worker
+        for job_id, job in self.running_jobs.items():
```
mrocklin: Instead of this loop can we compute the `job_id` from the name and index directly?  I'm thinking of what happens when we close down a cluster with a few thousand workers.  It's probably fine regardless, but if possible it'd be nice to avoid the quadratic (though small) cost

dask_jobqueue/tests/test_pbs.py
```diff
@@ -78,16 +80,17 @@ def test_job_script():
...
 @pytest.mark.env("pbs")  # noqa: F811
 def test_basic(loop):
-    with PBSCluster(walltime='00:02:00', processes=1, threads=2, memory='2GB', local_directory='/tmp',
-                    job_extra=['-V'], loop=loop) as cluster:
+    with PBSCluster(walltime='00:02:00', processes=1, threads=2, memory='2GB',
+                    local_directory='/tmp', job_extra=['-V'],
+                    loop=loop) as cluster:
         with Client(cluster) as client:
-            workers = cluster.start_workers(2)
+            cluster.start_workers(2)
```
mrocklin: Can we `assert cluster.pending_jobs or cluster.running_jobs` immediately?

docs/faq.rst
```diff
@@ -0,0 +1,13 @@
...
+==========================
+
+What is the difference between a Job and a Worker?
+--------------------------------------------------
+
+In dask-distributed, a ``Worker`` is a Python object and node in a dask
+``Cluster`` that serves two purposes, 1) serve data, and 2) perform\
+computations. Jobs are resources submitted to, and managed by, the job queueing
+system (e.g. PBS, SGE, etc.). In dask-jobqueue, a single Job may include one or
+more Workers.
```
mrocklin: We might put this information in the dostring of the most common function that users might see the term "job".

I'm somewhat against having a FAQ if it only has this one question in it.  This seems more like a section of "stuff I think you should know but couldn't find a place to put in the docs" rather than frequently asked questions.

dask_jobqueue/core.py
```diff
@@ -1,20 +1,95 @@
...
+    def update(self, worker=None, status=None):
+        ''' update the status this job'''
+        if worker is not None:
+            self.workers[worker.address] = worker
+        if status is not None:
+            self.status = status
+
+    def __repr__(self):
+        return "<%s: %r, status: %r, workers: %r>" % (
+            self.__class__.__name__, self.job_id, self.status, self.workers)
```
mrocklin: Thoughts on removing this class and just using dictionaries of sets in the plugin?  It seems like we have all of this information implicit in the `*_jobs` attributes already.

```python
self.STATUS_jobs[JOB_ID] = {WORKERS}
```

dask_jobqueue/core.py
```diff
@@ -96,7 +173,10 @@ def __init__(self,
             raise NotImplementedError('JobQueueCluster is an abstract class '
                                       'that should not be instanciated.')
 
-        #This attribute should be overriden
+        if '-' in name:
+            raise ValueError('name (%s) can not include the `-` character')
```
mrocklin: Is this because of the naming convention we're using to grab JOB-IDs?

jhamman: yes, the tricky part is that sometimes the worker names also have an additional `-integer` when we have grouped workers. For now, I just made the straightforward decision to keep it simple. If there are strong opposing thoughts, I'm happy to change this.

dask_jobqueue/core.py
jhamman: Yes, I think we could do this. In an early version of this work, I had some additional functionality on this class but that has all been removed now.

dask_jobqueue/core.py
mrocklin: I wonder if we could further isolate the name by using something like a double-hyphen

    name='dask-worker--{SGE_JOB_ID}`

That might be distinct enough so that we can isolate the job_id more reliably.

guillaumeeb:
> what you mean by "grouped workers" represents what you get when you execute dask-worker and that workers are the underlying worker processes

@lesteve that's what we're talking about here yes.

Launching dask-worker with `--nprocs` option greater than 1 will actually launch several workers that are grouped, because launched together. I had some difficulties to understand that too, especially with my Java/Spark background.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -99,41 +103,93 @@ def test_basic(loop):
             start = time()
             while len(client.scheduler_info()['workers']) > 0:
                 sleep(0.100)
-                assert time() < start + 10
+                assert time() < start + QUEUE_WAIT
 
-            assert not cluster.jobs
+            assert not cluster.running_jobs
```
guillaumeeb: We could also assert cluster.finished_jobs here, don't we?

dask_jobqueue/sge.py
```diff
@@ -56,8 +58,7 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
 
         super(SGECluster, self).__init__(**kwargs)
 
-        header_lines = ['#!/bin/bash']
-
+        header_lines = ['#!/usr/bin/env bash']
```
guillaumeeb: So you don't have a solution for propagating JOB_ID var in sge script?

dask_jobqueue/tests/test_pbs.py
```diff
@@ -99,41 +103,93 @@ def test_basic(loop):
...
+                    loop=loop) as cluster:
         cluster.adapt()
         with Client(cluster) as client:
             future = client.submit(lambda x: x + 1, 10)
-            assert future.result(60) == 11
+            assert cluster.pending_jobs or cluster.running_jobs
+            assert future.result(QUEUE_WAIT) == 11
+
+            start = time()
+            while not len(cluster.pending_jobs):
```
guillaumeeb: Don't we risk to directly have cluster.running_jobs non empty here, without seeing the cluster.pending_jobs step?

dask_jobqueue/tests/test_pbs.py
```diff
@@ -99,41 +103,93 @@ def test_basic(loop):
...
+                sleep(0.100)
+                assert time() < start + QUEUE_WAIT
+
+
+@pytest.mark.env("pbs")  # noqa: F811
+def test_adaptive_grouped(loop):
+    with PBSCluster(walltime='00:02:00', processes=2, threads=1, memory='2GB',
+                    local_directory='/tmp', job_extra=['-V'],
+                    loop=loop) as cluster:
+        cluster.adapt(minimum=1)
```
guillaumeeb: What does minimum=1 means here?

docs/faq.rst
guillaumeeb: And we should maybe point this part of the doc to the example that @mrocklin rightfully pointed to nicely informative!

docs/index.rst
```diff
@@ -42,14 +42,15 @@ resources when not actively computing.
 
 .. code-block:: python
 
-   cluster.adapt(minimum=1, maximum=100)
+   cluster.adapt(minimum=6, maximum=90)
```
guillaumeeb: In the cluster.adapt method, I guess that minimum and maximum are expressed in terms of worker processes, not jobs? Maybe this should be highlighted.

dask_jobqueue/tests/test_pbs.py
guillaumeeb: Maybe one of the test failure comes from this.

guillaumeeb:
Looking at the Slurm tests in Travis, all the jobs are failed, apparently from the start. Something may be wrong in the job script, or there is something Slurm does not like.

docs/faq.rst
jhamman: I've removed the FAQ.

dask_jobqueue/sge.py
jhamman: good catch. I thought I did but I'll have to sort it out.

jhamman: I take that back. SGE uses `JOB_ID` for its environment variable so its already present.

jhamman:
I'm quite close here. 

@mrocklin- question on how the scheduler plugins work. I'm getting the following error. My qualitative/interactive use of the adaptive cluster is that this should work. I'm wondering if there is some latency in the Adaptive that isn't scaling up in time for this assert statement. I'm trying to understand if we should expect a Adaptive to scale up the cluster before `client.submit` returns. Does that make sense? Perhaps I should put this assert statement in a sleep block?

```Python
    def test_adaptive(loop):
        with PBSCluster(walltime='00:02:00', processes=1, cores=2, memory='2GB', local_directory='/tmp',
                        job_extra=['-V'], loop=loop) as cluster:
            cluster.adapt()
            with Client(cluster) as client:
                future = client.submit(lambda x: x + 1, 10)
>               assert cluster.pending_jobs or cluster.running_jobs
E               assert (OrderedDict() or OrderedDict())
E                +  where OrderedDict() = <dask_jobqueue.pbs.PBSCluster object at 0x7fd3766b7a90>.pending_jobs
E                +  and   OrderedDict() = <dask_jobqueue.pbs.PBSCluster object at 0x7fd3766b7a90>.running_jobs
```

mrocklin:
Client.submit finishes as soon as it has placed the message to be sent to the scheduler in the right outbox.  There is a possible delay while it is in transmission.  So your understanding about scheduler plugins remains correct (it'll run at the same time that a task is received) it's just that the task hasn't made it to the scheduler yet.

dask_jobqueue/core.py
```diff
@@ -212,14 +276,12 @@ def job_file(self):
 
     def start_workers(self, n=1):
         """ Start workers and point them to our local scheduler """
-        workers = []
-        for _ in range(n):
+        num_jobs = math.ceil(n / self.worker_processes)
```
mrocklin: Probably need `from __future__ import division` at the top of this file.

jhamman:
Huzzah! Finally, CI is passing here. Thanks @lesteve and @mrocklin for the tips this week at Scipy.

lesteve:
Nice! I'll try to have a closer look today.

guillaumeeb:
Wont be able to test until two weeks for my part. Thanks for the hard work here. A lot of activity from Scipy apparently, I hope I would be able to come in the next years.

mrocklin:
+1 on @guillaumeeb coming to SciPy.  It would be great to meet you!

mrocklin:
I plan to merge this later today if there are no further comments.

PR #64, opened 2018-05-23
DOC: add conda install info

raybellwaves:
Not sure if the conda install info was left out intentionally whilst the API is being developed. Nonetheless, there is a conda build which I just used to install.


jhamman:
Thanks @raybellwaves! Yes, this was just recently added to conda-forge and we hadn't updated the docs accordingly.

PR #66, opened 2018-06-04
Add MoabCluster

spencerkclark:
As suggested in https://github.com/pangeo-data/pangeo/pull/284#discussion_r192569026 this adds a `MoabCluster` class to dask-jobqueue.  I also included an example with a configuration I've used at GFDL.

I'm not sure what we want to do about testing.  Do we need to set up a separate CI environment for testing with Moab, or is it sufficient to assume that if we have robust test coverage for `PBSCluster` that things should work properly?

cc: @guillaumeeb @jhamman

### DIFF
--- a/dask_jobqueue/__init__.py
+++ b/dask_jobqueue/__init__.py
@@ -1,6 +1,7 @@
 # flake8: noqa
 from . import config
 from .core import JobQueueCluster
+from .moab import MoabCluster
 from .pbs import PBSCluster
 from .slurm import SLURMCluster
 from .sge import SGECluster
--- a/dask_jobqueue/moab.py
+++ b/dask_jobqueue/moab.py
@@ -24,8 +24,12 @@ class MoabCluster(PBSCluster):
 
     Examples
     --------
+    >>> import os
     >>> from dask_jobqueue import MoabCluster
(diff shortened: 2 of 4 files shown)


spencerkclark:
Thanks @guillaumeeb, that's a good suggestion.  I updated the docstring.  Let me know if you have any further comments.

guillaumeeb:
I will merge this soon if there is no further comment.

lesteve:
LGTM.

> I'm not sure what we want to do about testing. Do we need to set up a separate CI environment for testing with Moab, or is it sufficient to assume that if we have robust test coverage for PBSCluster that things should work properly?

Just a comment about testing, I think not adding testing for this is fine, especially if it is similar in nature to PBS (I never heard about Moab before so I am not well placed to say). I guess in the future, if you realise that MoabCluster breaks in subtle ways with each new dask-jobqueue release it may make sense to invest time in setting up CI for MoabCluster.

guillaumeeb:
Thanks again @spencerkclark !

rabernat:
Is there a reason that MoabCluster is not visible in the API docs?
http://dask-jobqueue.readthedocs.io/en/latest/api.html

PR #70, opened 2018-06-15
Separate configuration per resource manager

### LINKED ISSUE #68: Reorganize the dask-jobqueue config file
It is not unheard of to have multiple job queue systems running from the machine. For example, the machine I use the most, Cheyenne, uses both PBS and Slurm. For cases like this, it would be nice if we revamped our config file stricture a bit.

Currently we have:

```yml
jobqueue:
  name: dask-worker
  threads: 2
  processes: 4
  memory: 8GB
  interface: null
  death-timeout: 60
  local-directory: null
  extra: ""
  env-extra: []

  queue: null
  project: null
  walltime: '00:30:00'

  pbs:
    resource-spec: null
    job-extra: []

  sge:
    resource-spec: null

  slurm:
    job-cpu: null
    job-mem: null
    job-extra: {}
```

I'd like to propose that we move some of the jobqueue values into the pbs/sge/slurm catagories:

```yml
jobqueue:
  # dask options
  name: dask-worker
  death-timeout: 60

  pbs:
    threads: 2
    processes: 4
    memory: 8GB
    interface: null
    local-directory: null
    extra: ""
    env-extra: []
    queue: null
    project: null
    walltime: '00:30:00'
    resource-spec: null
    job-extra: []

  sge:
    threads: 2
    processes: 4
    memory: 8GB
    interface: null
    local-directory: null
    extra: ""
    env-extra: []
    queue: null
    project: null
    walltime: '00:30:00'
    resource-spec: null

  slurm:
    threads: 2
    processes: 4
    memory: 8GB
    interface: null
    local-directory: null
    extra: ""
    env-extra: []
    queue: null
    project: null
    walltime: '00:30:00'

    job-cpu: null
    job-mem: null
    job-extra: {}
```

mrocklin:
Previously we maintained common configuration elements like name,
threads, and processes in the toplevel jobqueue config namespace.

Now we pull these out into each cluster's config namespace.

This also forces us to pick out config values at object instantiation
time, which should help with changing config values at runtime.

Fixes #68

### DIFF
--- a/dask_jobqueue/pbs.py
+++ b/dask_jobqueue/pbs.py
@@ -58,20 +58,21 @@ class PBSCluster(JobQueueCluster):
     # Override class variables
     submit_command = 'qsub'
     cancel_command = 'qdel'
+    scheduler_name = 'pbs'
 
-    def __init__(self,
-                 queue=dask.config.get('jobqueue.queue'),
-                 project=dask.config.get('jobqueue.project'),
-                 resource_spec=dask.config.get('jobqueue.pbs.resource-spec'),
-                 walltime=dask.config.get('jobqueue.walltime'),
-                 job_extra=dask.config.get('jobqueue.pbs.job-extra'),
-                 **kwargs):
+    def __init__(self, *args, queue=None, project=None, resource_spec=None, walltime=None, job_extra=None, **kwargs):
+        if queue is None:
+            queue = dask.config.get('jobqueue.%s.queue' % self.scheduler_name)
+        if resource_spec is None:
+            resource_spec = dask.config.get('jobqueue.%s.resource-spec' % self.scheduler_name)
+        if walltime is None:
+            walltime = dask.config.get('jobqueue.%s.walltime' % self.scheduler_name)
+        if job_extra is None:
+            job_extra = dask.config.get('jobqueue.%s.job-extra' % self.scheduler_name)
+        project = project or dask.config.get('jobqueue.%s.project' % self.scheduler_name) or os.environ.get('PBS_ACCOUNT')
 
         # Instantiate args and parameters from parent abstract class
-        super(PBSCluster, self).__init__(**kwargs)
-
-        # Try to find a project name from environment variable
-        project = project or os.environ.get('PBS_ACCOUNT')
+        super(PBSCluster, self).__init__(*args, **kwargs)
 
         header_lines = []
         # PBS header build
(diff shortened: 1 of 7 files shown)


dask_jobqueue/pbs.py
```diff
@@ -58,20 +58,21 @@ class PBSCluster(JobQueueCluster):
...
-                 **kwargs):
+    def __init__(self, *args, queue=None, project=None, resource_spec=None, walltime=None, job_extra=None, **kwargs):
+        if queue is None:
+            queue = dask.config.get('jobqueue.%s.queue' % self.scheduler_name)
+        if resource_spec is None:
+            resource_spec = dask.config.get('jobqueue.%s.resource-spec' % self.scheduler_name)
+        if walltime is None:
+            walltime = dask.config.get('jobqueue.%s.walltime' % self.scheduler_name)
+        if job_extra is None:
+            job_extra = dask.config.get('jobqueue.%s.job-extra' % self.scheduler_name)
```
guillaumeeb: There is probably a reason I can't see, but why don't you keep the same one liner syntax here, as in JobQueueCluster class?

mrocklin: I think that we should probably switch to `is None` everywhere so that users can explicitly say things like `job_extra = ''` or `job_extra = []` and have that respected.

jhamman:
I'm generally in favor of the changes I see here. Two of the test builds failed. @mrocklin - can you take a look at those and see if they are meaningful.

mrocklin:
OK, tests pass.  Things seem cleanish to me.

mrocklin:
> Modifications of the file are not propagated at runtime despite what I thought, I had to restart my jupyter kernel to see my changes in the file. Is there a cache mechanism in dask.config?

That's correct.  Today you would have to trigger this manually by calling `dask.config.refresh()`.

guillaumeeb:
Oh yes that's true! I had forgotten about that.  Perfect then.

mrocklin:
OK to merge?

lesteve:
I am fine with the changes (as I mentioned before somewhere I was a big fan of configuration file argument set at import time behaviour).

@jhamman if you can quickly double check that it solves your original issue, it would be great!

Also @mrocklin just curious, was the addition of MoabCluster tests to test_pbs.py meant to be part of this PR?

mrocklin:
Yeah, I was making changes to MoabCluster that wouldn't have been tested
otherwise.  The easiest way seemed to do this seemed to be to parametrize
the PBSCluster tests.

lesteve:
> Yeah, I was making changes to MoabCluster that wouldn't have been tested otherwise.

OK makes sense, thanks!

PR #72, opened 2018-06-20
set `self.worker_memory` to None when appropriate

basnijholt:
We currently have a misconfigured SLURM setup where we can't specify the memory.

Therefore using `dask_jobqueue` always leads to a `Memory specification can not be satisfied` error, even when `memory=None`. This PR fixes that problem.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -112,7 +112,7 @@ class JobQueueCluster(Cluster):
 
         # Keep information on process, threads and memory, for use in
         # subclasses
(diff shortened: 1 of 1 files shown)


jhamman:
Thanks @basnijholt - this seems fine to me. Any objections from others?

mrocklin:
I'm going to go ahead and merge.  Thanks @basnijholt !

I'm fine with ternary if.  It looks like we don't yet include flake8 in our test suite.  I'm inclined to handle this more generally by adding a test to the CI matrix rather than troubling people on each PR.

PR #73, opened 2018-06-21
Dashboard docs

### LINKED ISSUE #71: Unable to view the status dashboard
In today's experiments with Dask + dask-jobqueue, I found that I could not load the Bokeh dashboard that @mrocklin keeps showing me :smile:, which has kept me wondering what exactly is the progress on my simple, embarrassingly parallel task of "loading ~900+ matlab `.mat` matrices into memory".

The URL provided by the client is: `http://172.16.23.102:8787/status`.

I'm able to ping the IP address in there:

```bash
$ ping 172.16.23.102
```

However, I'm unable to access the page in my browser; I get a timeout error.

In terms of network settings, I'm on my work VPN.

Is there something that's blocking access that I'm missing?

ericmjl:
Addresses and closes #71. 

I have done two things here:

1. Added an example for SGE.
2. Added an example on how to access dashboard.

cc: @jhamman @mrocklin.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -99,3 +99,34 @@ SLURM Deployment: Low-priority node usage
                                       'export LC_ALL="en_US.utf8"'],
                            job_extra=['--qos="savio_lowprio"'])
 
+
+Viewing Dask Dashboard
+~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, the Dask dashboard might not be directly accessible via the browser.
+To solve this, you can use SSH tunnelling. Let's say we started with the
+following setup:
+
+.. code-block:: python
+
+    from dask_jobqueue import SGECluster
+    from distributed import Client
+
+    cluster = SGECluster(queue='default.q',
+                         walltime="1500000",
+                         processes=10,
+                         memory='20GB')
+
+    client = Client(cluster)
+
+Say for example, on inspection of the ``client`` object, you see that the
+Dashboard can be viewed at ``http://172.16.23.102:8787/status``. If the webpage
+is not directly accessible in your browser, then the next thing to try would be
+SSH tunneling.
+
+.. code-block:: bash
+
+    # General syntax
+    $ ssh -fN your-login@scheduler-ip-address -L port-number:localhost:port-number
+    # As applied to this example:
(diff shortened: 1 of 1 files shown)


docs/examples.rst
```diff
@@ -87,3 +99,34 @@ SLURM Deployment: Low-priority node usage
...
+Dashboard can be viewed at ``http://172.16.23.102:8787/status``. If the webpage
+is not directly accessible in your browser, then the next thing to try would be
+SSH tunneling.
+
+.. code-block:: bash
+
+    # General syntax
+    $ ssh -fN your-login@scheduler-ip-address -L port-number:localhost:port-number
+    # As applied to this example:
+    $ ssh -fN username@172.16.23.102 -L 8787:localhost:8787
```
mrocklin: We should also say where they should find the tunneled address.  Perhaps something like "Your dashboard will then be accessible at http://localhost:8787/status " or something similar.

docs/examples.rst
```diff
@@ -48,12 +48,24 @@ can be used, called ``MoabCluster``:
...
+
+.. code-block:: python
+
+    from dask_jobqueue import SGECluster
+
+
+    cluster = SGECluster(queue='default.q',
+                         walltime="1500000",
+                         processes=10,
+                         memory='20GB')
```
mrocklin: I'm curious, you don't specify anything else when launching SGECluster?  This section is intended to be filled with real-world examples, so we encourage you not to strip things out for simplicity's sake.

ericmjl:
Btw, you're awfully early, @mrocklin!

mrocklin:
East coast time, It's 8:30am

docs/examples.rst
mrocklin: Alright.  Thanks for sharing! 

In case you haven't already read through it, http://dask.pydata.org/en/latest/setup/hpc.html might be helpful.  I've found that it's common for people to want to specify network interfaces and local directories in particular.

ericmjl:
Looks like the test failed on a piece of code. Anything I could help with here, @mrocklin?

mrocklin:
Not sure.  Maybe the SLURM cluster took a while to fire up?  I suspect that any of the other maintainers has more experience here than I do.

mrocklin:
Thanks @ericmjl !

PR #77, opened 2018-06-26
Find dask-worker through the interpreter rather than assume a location

### LINKED ISSUE #75: Assumption made about location of dask-worker
In https://github.com/dask/dask-jobqueue/blob/774f2610f7e7befd44e04d972c42e299b942fa5e/dask_jobqueue/core.py#L15
the assumption is made that the `dask-worker` script is in the same place as the python executable. This is not true in our case since dask is available through an environment module (so the availability of the script is handled via the `$PATH` envvar).

Would it be possible to add a check before https://github.com/dask/dask-jobqueue/blob/774f2610f7e7befd44e04d972c42e299b942fa5e/dask_jobqueue/core.py#L127 to verify that this default actually exists and, if not, use the location found in `$PATH` (otherwise die gracefully)? Something like
```
if not os.path.isfile(os.path.join(dirname, 'dask-worker'):
    if os.path.isfile(os.path.abspath('dask-worker'):
        dirname = os.path.join(os.path.abspath('dask-worker'), '..')
    else:
        # Trigger a graceful death
```

ocaisa:
Fixes #75 (at least for our system)

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -12,7 +12,10 @@ from distributed import LocalCluster
 from distributed.deploy import Cluster
 from distributed.utils import get_ip_interface, ignoring, parse_bytes, tmpfile
 
-dirname = os.path.dirname(sys.executable)
+worker_bin_path = (
+    '%(python)s -m distributed.cli.dask_worker'
+    % dict(python=sys.executable)
+)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
@@ -124,8 +127,8 @@ class JobQueueCluster(Cluster):
         self._env_header = '\n'.join(env_extra)
 
         # dask-worker command line build
-        self._command_template = os.path.join(
-            dirname, 'dask-worker %s' % self.scheduler.address)
+        self._command_template = ' '.join([
+            worker_bin_path, '%s' % self.scheduler.address])
         if threads is not None:
             self._command_template += " --nthreads %d" % threads
         if processes is not None:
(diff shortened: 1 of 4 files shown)


lesteve:
To fix the CI failures, I think we do look for `python -m distributed.cli.dask_worker` and not `dask-worker` in the script.

ocaisa:
Thanks, they are passing again now. I left out `python` just to leave some flexibility there.

lesteve:
Thanks a lot for fixing the tests!

> I left out python just to leave some flexibility there.

Can you elaborate what you mean by "some flexibility"? I would be in favour of putting `python` back because I can't see a good reason for leaving it out.

Other than this the change looks fine to me. @mrocklin any opinions on this?

ocaisa:
It assumes that `sys.executable` is `python` as opposed to `python2` or `python3` but maybe that's not relevant for the tests. NP, I'll put it back in.

lesteve:
Ah good point, look for `{} -m distributed.cli.dask_worker'.format(sys.executable)` in the tests then.

lesteve:
I pushed some minor tweaks, hopefully you don't mind @ocaisa! Let's see whether the CIs are happy with that.

lesteve:
Small tip for next time @ocaisa while I am at it: best practice is to use a different branch than master to create a PR. Basically the only reason you want to touch your master branch is to bring it up-to-date with master.

ocaisa:
Sure, improvements welcome!

ocaisa:
@lesteve Yes, sorry, I was being lazy, I didn't think it would be getting merged, it was initially just a proof of concept

lesteve:
I am going to merge this one, thanks a lot @ocaisa!

Issue #4, opened 2018-03-01
Add LSF


mrocklin:
It might be valuable to extend this repository with a solution for LSF.  My hope is that this is relatively easy for someone with modest LSF experience.  Looking at the current solutions for [PBS](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/pbs.py) or [SLURM](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/slurm.py) might be helpful (they're about 100 lines, mostly docstrings)

jakirkham:
Happy to give this a go. We use LSF at work. So this should be pretty useful.

lzamparo:
@jakirkham if it helps, I tried making a [dask-distributed LSF](https://github.com/dask/distributed/issues/1260) script a while back.  I encountered an error I couldn't fix, but maybe you can? Here's a [gist](https://gist.github.com/lzamparo/1ec0275efac635b103725412808319e9).

lesteve:
Just to split the task in smaller chunks:
* add a LSFCluster that inherits from `dask_jobqueue.core.JobQueueCluster`
* define `LSFCluster.submit_command` (probably `bsub`?) and `LSFCluster.cancel_command` (probably `bkill`?)
* implement `LSFCluster._job_id_from_submit_output` that takes a string (stdout output from the `submit_cmd`) and turns into into a job identifier.

Once implemented, it would be great to test it minimally on a LSFCluster that you have at your disposal.

jakirkham:
Sorry have been busy with other things lately. If someone has interest and time, they should feel free to go ahead.

lesteve:
> If someone has interest and time, they should feel free to go ahead.

Yep, that was what I had in mind by specifying the steps in more details.

lzamparo:
@lesteve I can give this a go; thanks for the direction.  Is there a specific fixture-based test implemented for other cluster methods, or am I free to try some of my own tasks?

lesteve:
It'd be great if you could give this a go! For the first iteration, I think you can try to put together and run a small snippet, e.g. something along these lines and get it to work on your local LSFCluster:

```py
from dask_jobqueue import LSFCluster, Client

cluster = LSFCluster(...)  # use some arguments that make sense on your local cluster
client = Client(cluster)
result = client.map(lambda x: x + 1, range(10))
client.gather(result)
```

[...]

PR #78, opened 2018-06-27
Add lsf

raybellwaves:
working on https://github.com/dask/dask-jobqueue/issues/4

Thanks to @jhamman for his (in person) help with this. Hopefully I can get this squared away by the end of the week.

FYI I'm testing on python 3.6 at pegasus (University of Miami's HPC) and I was getting an un-obvious `psutil` error (see below). After installing the dependencies I installed a conda version of `psutil` (`conda install -c conda-forge psutil`) and it went away.

```
In [1]: from dask_jobqueue import LSFCluster
ImportError                               Traceback (most recent call last)
<ipython-input-1-8ba34d16bb88> in <module>()
----> 1 from dask_jobqueue import LSFCluster

~/PYTHON3/dask-jobqueue/dask_jobqueue/__init__.py in <module>()
      1 # flake8: noqa
      2 from . import config
----> 3 from .core import JobQueueCluster
      4 from .moab import MoabCluster
      5 from .pbs import PBSCluster

~/PYTHON3/dask-jobqueue/dask_jobqueue/core.py in <module>()
      8 import dask
      9 import docrep
---> 10 from distributed import LocalCluster
     11 from distributed.deploy import Cluster
     12 from distributed.utils import get_ip_interface, ignoring, parse_bytes, tmpfile

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/distributed/__init__.py in <module>()
      3 from . import config
      4 from dask.config import config
----> 5 from .core import connect, rpc
      6 from .deploy import LocalCluster, Adaptive
      7 from .diagnostics import progress

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/distributed/core.py in <module>()
     23                    unparse_host_port, get_address_host_port)
     24 from .metrics import time
---> 25 from .system_monitor import SystemMonitor
     26 from .utils import (get_traceback, truncate_exception, ignoring, shutting_down,
     27                     PeriodicCallback, parse_timedelta)

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/distributed/system_monitor.py in <module>()
      2 
      3 from collections import deque
----> 4 import psutil
      5 
      6 from .compatibility import WINDOWS

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/psutil/__init__.py in <module>()
     97     PROCFS_PATH = "/proc"
     98 
---> 99     from . import _pslinux as _psplatform
    100 
    101     from ._pslinux import IOPRIO_CLASS_BE  # NOQA

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/psutil/_pslinux.py in <module>()
     24 from . import _common
     25 from . import _psposix
---> 26 from . import _psutil_linux as cext
     27 from . import _psutil_posix as cext_posix
     28 from ._common import ENCODING

ImportError: /nethome/rxb826/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/psutil/_psutil_linux.cpython-36m-x86_64-linux-gnu.so: undefined symbol: __intel_sse4_strncpy
```

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -157,7 +157,7 @@ class JobQueueCluster(Cluster):
         else:
             host = socket.gethostname()
 
-        self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+        self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
 
         # Keep information on process, threads and memory, for use in subclasses
         self.worker_memory = parse_bytes(memory)
@@ -237,7 +237,7 @@ class JobQueueCluster(Cluster):
     @property
     def scheduler(self):
         """ The scheduler of this cluster """
-        return self.cluster.scheduler
+        return self.local_cluster.scheduler
 
     def _calls(self, cmds):
         """ Call a command using subprocess.communicate
@@ -309,7 +309,7 @@ class JobQueueCluster(Cluster):
 
     def __exit__(self, type, value, traceback):
         self.stop_workers(self.jobs)
-        self.cluster.__exit__(type, value, traceback)
+        self.local_cluster.__exit__(type, value, traceback)
 
     def _job_id_from_submit_output(self, out):
         raise NotImplementedError('_job_id_from_submit_output must be '
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -159,7 +159,7 @@ class JobQueueCluster(Cluster):
 
         self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
 
-        # Keep information on process, cores and memory, for use in subclasses
+        # Keep information on process, cores, and memory, for use in subclasses
         self.worker_memory = parse_bytes(memory)
 
         self.worker_processes = processes
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -103,7 +103,6 @@ class JobQueueCluster(Cluster):
                  env_extra=None,
                  walltime=None,
                  threads=None,
-                 shell=False,
                  **kwargs
                  ):
         """ """
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -143,8 +143,6 @@ class JobQueueCluster(Cluster):
     _adaptive_options = {
         'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
 
-    popen_shell = False
-
     def __init__(self,
                  name=None,
                  cores=None,
@@ -282,7 +280,7 @@ class JobQueueCluster(Cluster):
                 f.write(self.job_script())
             yield fn
 
-    def submit_job(self, script_filename):
+    def _submit_job(self, script_filename):
         return self._call(shlex.split(self.submit_command) + [script_filename])
 
     def start_workers(self, n=1):
@@ -291,7 +289,7 @@ class JobQueueCluster(Cluster):
         num_jobs = math.ceil(n / self.worker_processes)
         for _ in range(num_jobs):
             with self.job_file() as fn:
-                out = self.submit_job(fn)
+                out = self._submit_job(fn)
                 job = self._job_id_from_submit_output(out.decode())
                 logger.debug("started job: %s" % job)
                 self.pending_jobs[job] = {}
@@ -301,7 +299,7 @@ class JobQueueCluster(Cluster):
         """ The scheduler of this cluster """
         return self.local_cluster.scheduler
 
-    def _calls(self, cmds):
+    def _calls(self, cmds, **kwargs):
         """ Call a command using subprocess.communicate
 
         This centralzies calls out to the command line, providing consistent
@@ -327,9 +325,9 @@ class JobQueueCluster(Cluster):
         for cmd in cmds:
             logger.debug(' '.join(cmd))
             procs.append(subprocess.Popen(cmd,
-                                          shell=self.popen_shell,
                                           stdout=subprocess.PIPE,
-                                          stderr=subprocess.PIPE))
+                                          stderr=subprocess.PIPE,
+                                          **kwargs))
 
         result = []
         for proc in procs:
@@ -339,9 +337,9 @@ class JobQueueCluster(Cluster):
             result.append(out)
         return result
 
-    def _call(self, cmd):
+    def _call(self, cmd, **kwargs):
         """ Singular version of _calls """
-        return self._calls([cmd])[0]
+        return self._calls([cmd], **kwargs)[0]
 
     def stop_workers(self, workers):
         """ Stop a list of workers"""
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -93,7 +93,7 @@ class JobQueueCluster(Cluster):
 
     # Required for excuting commands through the shell in the subprocess module.
     # It handles the shell input redirection e.g. bsub < script_filename.sh
-    # and does not consider it a less than operator.    
+    # and does not consider '<' as a 'less than' operator.
     shell = False
 
     def __init__(self,
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -54,6 +54,7 @@ class JobQueuePlugin(SchedulerPlugin):
         logger.debug("adding worker %s" % worker)
         w = scheduler.workers[worker]
         job_id = _job_id_from_worker_name(w.name)
+
         logger.debug("job id for new worker: %s" % job_id)
         self.all_workers[worker] = (w.name, job_id)
 
@@ -345,9 +346,6 @@ class JobQueueCluster(Cluster):
     def _call(self, cmd):
         """ Singular version of _calls """
         return self._calls([cmd])[0]
-
-    def kill_jobs(self, jobs):
-        return self._call([self.cancel_command] + list(jobs))
 
     def stop_workers(self, workers):
         """ Stop a list of workers"""
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -93,7 +93,7 @@ class JobQueueCluster(Cluster):
 
     # Required for excuting commands through the shell in the subprocess module.
     # It handles the shell input redirection e.g. bsub < script_filename.sh
-    # and does not consider '<' as a 'less than' operator.
+    # and does not consider '<' as a command, file or directory.
     shell = False
 
     def __init__(self,
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -66,6 +66,9 @@ class JobQueueCluster(Cluster):
     cancel_command: str
         Abstract attribute for job scheduler cancel command,
         should be overriden
+    scheduler_name: str
+        Abstract attribute for job scheduler name,
+        should be overriden
 
     See Also
     --------
@@ -100,6 +103,7 @@ class JobQueueCluster(Cluster):
                  env_extra=None,
                  walltime=None,
                  threads=None,
+                 shell=False,
                  **kwargs
                  ):
         """ """
@@ -153,9 +157,9 @@ class JobQueueCluster(Cluster):
         else:
             host = socket.gethostname()
 
-        self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+        self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
 
-        # Keep information on process, cores, and memory, for use in subclasses
+        # Keep information on process, threads and memory, for use in subclasses
         self.worker_memory = parse_bytes(memory)
 
         self.worker_processes = processes
@@ -210,12 +214,21 @@ class JobQueueCluster(Cluster):
                 f.write(self.job_script())
             yield fn
 
+    def submit_job(self, script_filename):
+        """ Sumbits job and handles lsf exception """
+        if self.scheduler_name == 'lsf':
+            piped_cmd = [self.submit_command + '<\"' + script_filename + '\"']
+            self.shell = True
+            return self._call(piped_cmd)
+        else:
+            return self._call(shlex.split(self.submit_command) + [script_filename])
+
     def start_workers(self, n=1):
         """ Start workers and point them to our local scheduler """
         workers = []
         for _ in range(n):
             with self.job_file() as fn:
-                out = self._call(shlex.split(self.submit_command) + [fn])
+                out = self.submit_job(fn)
                 job = self._job_id_from_submit_output(out.decode())
                 self.jobs[self.n] = job
                 workers.append(self.n)
@@ -224,7 +237,7 @@ class JobQueueCluster(Cluster):
     @property
     def scheduler(self):
         """ The scheduler of this cluster """
-        return self.local_cluster.scheduler
+        return self.cluster.scheduler
 
     def _calls(self, cmds):
         """ Call a command using subprocess.communicate
@@ -251,6 +264,7 @@ class JobQueueCluster(Cluster):
         for cmd in cmds:
             logger.debug(' '.join(cmd))
         procs = [subprocess.Popen(cmd,
+                                  shell=self.shell,
                                   stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE)
                  for cmd in cmds]
@@ -273,6 +287,7 @@ class JobQueueCluster(Cluster):
             return
         workers = list(map(int, workers))
         jobs = [self.jobs[w] for w in workers]
+        self.shell = False
         self._call([self.cancel_command] + list(jobs))
         for w in workers:
             with ignoring(KeyError):
@@ -294,7 +309,7 @@ class JobQueueCluster(Cluster):
 
     def __exit__(self, type, value, traceback):
         self.stop_workers(self.jobs)
-        self.local_cluster.__exit__(type, value, traceback)
+        self.cluster.__exit__(type, value, traceback)
 
     def _job_id_from_submit_output(self, out):
         raise NotImplementedError('_job_id_from_submit_output must be '
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -90,6 +90,8 @@ class JobQueueCluster(Cluster):
     submit_command = None
     cancel_command = None
     scheduler_name = ''
+    
+    shell = False # Required for submitting jobs with LSF
 
     def __init__(self,
                  name=None,
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -1,17 +1,22 @@
-from contextlib import contextmanager
+from __future__ import absolute_import, division, print_function
+
 import logging
+import math
 import shlex
 import socket
 import subprocess
 import sys
 import warnings
+from collections import OrderedDict
+from contextlib import contextmanager
 
 import dask
 import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
-from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
-                               format_bytes)
+from distributed.diagnostics.plugin import SchedulerPlugin
+from distributed.utils import (
+    format_bytes, get_ip_interface, parse_bytes, tmpfile)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
@@ -26,6 +31,54 @@ Please specify job size with the following keywords:
 -  memory: total memory per job, across all processes
 -  processes: number of processes to launch, splitting the quantities above
 """.strip()
+
+
+def _job_id_from_worker_name(name):
+    ''' utility to parse the job ID from the worker name
+
+    template: 'prefix--jobid--suffix'
+    '''
+    _, job_id, _ = name.split('--')
+    return job_id
+
+
+class JobQueuePlugin(SchedulerPlugin):
+    def __init__(self):
+        self.pending_jobs = OrderedDict()
+        self.running_jobs = OrderedDict()
+        self.finished_jobs = OrderedDict()
+        self.all_workers = {}
+
+    def add_worker(self, scheduler, worker=None, name=None, **kwargs):
+        ''' Run when a new worker enters the cluster'''
+        logger.debug("adding worker %s" % worker)
+        w = scheduler.workers[worker]
+        job_id = _job_id_from_worker_name(w.name)
+        logger.debug("job id for new worker: %s" % job_id)
+        self.all_workers[worker] = (w.name, job_id)
+
+        # if this is the first worker for this job, move job to running
+        if job_id not in self.running_jobs:
+            logger.debug("this is a new job")
+            self.running_jobs[job_id] = self.pending_jobs.pop(job_id)
+
+        # add worker to dict of workers in this job
+        self.running_jobs[job_id][w.name] = w
+
+    def remove_worker(self, scheduler=None, worker=None, **kwargs):
+        ''' Run when a worker leaves the cluster'''
+        logger.debug("removing worker %s" % worker)
+        name, job_id = self.all_workers[worker]
+        logger.debug("removing worker name (%s) and"
+                     "job_id (%s)" % (name, job_id))
+
+        # remove worker from this job
+        del self.running_jobs[job_id][name]
+
+        # once there are no more workers, move this job to finished_jobs
+        if not self.running_jobs[job_id]:
+            logger.debug("that was the last worker for job %s" % job_id)
+            self.finished_jobs[job_id] = self.running_jobs.pop(job_id)
 
 
 @docstrings.get_sectionsf('JobQueueCluster')
@@ -90,6 +143,8 @@ class JobQueueCluster(Cluster):
     submit_command = None
     cancel_command = None
     scheduler_name = ''
+    _adaptive_options = {
+        'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
 
     popen_shell = False
 
@@ -160,15 +215,17 @@ class JobQueueCluster(Cluster):
 
         self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
 
-        # Keep information on process, cores, and memory, for use in subclasses
-        self.worker_memory = parse_bytes(memory)
-
+        # Keep information on process, threads and memory, for use in
+        # subclasses
+        self.worker_memory = parse_bytes(memory) if memory is not None else None
         self.worker_processes = processes
         self.worker_cores = cores
         self.name = name
 
-        self.jobs = dict()
-        self.n = 0
+        # plugin for tracking job status
+        self._scheduler_plugin = JobQueuePlugin()
+        self.local_cluster.scheduler.add_plugin(self._scheduler_plugin)
+
         self._adaptive = None
 
         self._env_header = '\n'.join(env_extra)
@@ -184,10 +241,8 @@ class JobQueueCluster(Cluster):
         mem = format_bytes(self.worker_memory / self.worker_processes)
         mem = mem.replace(' ', '')
         self._command_template += " --memory-limit %s" % mem
+        self._command_template += " --name %s--${JOB_ID}--" % name
 
-        if name is not None:
-            self._command_template += " --name %s" % name
-            self._command_template += "-%(n)d" # Keep %(n) to be replaced later
         if death_timeout is not None:
             self._command_template += " --death-timeout %s" % death_timeout
         if local_directory is not None:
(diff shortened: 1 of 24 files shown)


dask_jobqueue/jobqueue.yaml
```diff
@@ -79,3 +79,25 @@ jobqueue:
...
+    local-directory: null
+
+    # LSF resource manager options
+    queue: null
+    project: null
+    walltime: '00:30:00'
+    extra: ""
+    env-extra: []
+    job-cpu: null
+    job-mem: null
```
mrocklin: I wonder, can we use the `memory` and `threads * processes` entries to remove the need for these entries?  This is probably also a question for other dask-jobqueue maintainers, as I know that these appear in other configurations.

mrocklin:
It would be useful to have a basic test, even if it only performs sanity-checks on the header.  Here is an example for SLURM

https://github.com/dask/dask-jobqueue/blob/4cd24dc5b1dd4bb9eda1cf27eeaf453787faf873/dask_jobqueue/tests/test_slurm.py#L11-L21

mrocklin:
Thanks for working on this!

jakirkham:
Could you please share how this should be tested? In particular, would be good to have a short script to try starting up Dask on LSF and running some simple computation with it.

mrocklin:
I would expect the following to work for any job-queue cluster

```python
from dask_jobqueue import LSFCluster
cluster = LSFCluster()

client = Client(cluster)

assert client.submit(lambda x: x + 1, 10).result() == 11
```

raybellwaves:
Hitting a wall here.
https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/slurm.py is probably closest to the lsf submit.
There is some code in slurm.py which explicitly sets n to 1 then sets ncpus using `#SBATCH --cpus-per-task` but there is no such command is LSF see [here](https://slurm.schedmd.com/rosetta.pdf)
https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/slurm.py#L88-L94
not sure if I need to set -n as 1 at the start?

When I run `client.submit(lambda x: x + 1, 10).result()` at the moment is just sits.

guillaumeeb:
>There is some code in slurm.py which explicitly sets n to 1 then sets ncpus using #SBATCH --cpus-per-task but there is no such command is LSF see here

You probably want to use the following options to indicate the use of one node with several processes:
````
#BSUB -R "span[hosts=1]"
#BSUB -n 8
````
see https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.3/lsf_admin/span_string.html or https://www.hpc.dtu.dk/?page_id=1401

File: dask_jobqueue/lsf.py [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+        header_lines.append('#BSUB -M %s' % memory_string)
+        header_lines.extend(['#BSUB %s' % arg for arg in job_extra])
+
+        # Declare class attribute that shall be overriden
+        self.job_header = '\n'.join(header_lines)
+
+        logger.debug("Job script: \n %s" % self.job_script())
+
+    def _job_id_from_submit_output(self, out):
+        return out.split('.')[0].strip()
```
guillaumeeb: Here you need to either do a parsing of the output string of the job, which seems to be something like:
````
Job is submitted to <cpp> project.
Job <16633929> is submitted to default queue <general>.
````
But you more likely want to find the correct option for just outputing the job ID after the bsub command. Unfortunatly I did not find such an option after a few minutes Google search.

File: dask_jobqueue/lsf.py [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+    >>> client = Client(cluster)
+
+    This also works with adaptive clusters.  This automatically launches and
+    kill workers based on load.
+
+    >>> cluster.adapt()
+    """, 4)
+
+    # Override class variables
+    submit_command = 'bsub'
```
guillaumeeb: From what I understand from your last comment, the `submit_command` should be `bsub <` instead of just `bsub`.

raybellwaves:
Thanks for the comments @guillaumeeb!

I think `submit_command = 'bsub'` was the sticking point from me. `submit_command = 'bsub<'` seems to be doing something now! I had `submit_command = 'bsub <'` previously but https://github.com/dask/dask-jobqueue/blob/d1cd730cd3da7b3b21023596a89516ca957b85d4/dask_jobqueue/core.py#L185 was splitting the command. `bsub<` works the same as `bsub <` anyway.

Someone is hammering the general queue so i'll report back on progress later today or tomorrow.

```
In [1]: from dask_jobqueue import LSFCluster
In [2]: cluster = LSFCluster(walltime='00:02', processes=2, threads=1, memory='4GB')
In [3]: cluster.job_header
Out[3]: '#BSUB -J dask-worker\n#BSUB -e dask-worker.err\n#BSUB -o dask-worker.out\n#BSUB -W 00:02\n#BSUB -R "span[hosts=1]"\n#BSUB -n 2\n#BSUB -M 7630'
In [4]: cluster.job_script()
Out[4]: '#!/bin/bash\n\n#BSUB -J dask-worker\n#BSUB -e dask-worker.err\n#BSUB -o dask-worker.out\n#BSUB -W 00:02\n#BSUB -R "span[hosts=1]"\n#BSUB -n 2\n#BSUB -M 7630\n\n\n\n/nethome/rxb826/local/bin/miniconda3/envs/d-jq-test/bin/python -m distributed.cli.dask_worker tcp://10.10.0.14:38155 --nthreads 1 --nprocs 2 --memory-limit 4GB --name dask-worker-2 --death-timeout 60\n'
In [5]: from dask.distributed import Client
In [6]: client = Client(cluster)
In [7]: client.submit(lambda x: x + 1, 10).result()

# In another terminal
[rxb826@pegasus ~]$ bjobs
JOBID     USER    STAT  QUEUE      FROM_HOST   EXEC_HOST   JOB_NAME   SUBMIT_TIME
16639606  rxb826  PEND  general    login4                  *sk-worker Jun 28 15:07
[rxb826@pegasus ~]$ bqueues
QUEUE_NAME      PRIO STATUS          MAX JL/U JL/P JL/H NJOBS  PEND   RUN  SUSP
general         100  Open:Active       -    -    -    - 14178 13446   708     0
```

raybellwaves:
Well my hack with `submit_command` in `lsf.py` and the `subprocess.Popen` in `core.py` broke `cluster.stop_workers(workers)` so I need to think of a solution for `bsub <` without butchering `core.py`.

guillaumeeb:
See how this is done in ipyparallel: https://github.com/ipython/ipyparallel/blob/6.1.1/ipyparallel/apps/launcher.py#L1397. Maybe you could try to use a similar syntax.

There is still the problem of `shell=True` keyword.

Maybe we should change the way core.py is written and have dedicated start and stop functions, that could be easily overridden if needed. Or just add a `launch_command()` function that the other one will call, and that just do the Popen part. @jhamman, @mrocklin, @lesteve any thoughts?

lesteve:
> Maybe we should change the way core.py is written and have dedicated start and stop functions, that could be easily overridden if needed. Or just add a launch_command() function that the other one will call, and that just do the Popen part. @jhamman, @mrocklin, @lesteve any thoughts?

Not sure what the best way is, but it looks like we need some special treatment for LSF indeed since it takes stdin and not the script name. Maybe something like this (I am guessing this is similar to your `launch_command` suggestion):

```py
# in JobQueueCluster
def submit_job(self, script_filename):
    return self._call(shlex.split(self.submit_command) + [script_filename])

def start_workers(self, n=1):
    ...
    with self.job_file() as fn:
        out = self.submit_job(fn)
        ...
```

```py
# in LSFCluster
def submit_job(self, filename):
    return self._call(shlex.split(self.submit_command) + ['<', script_filename], popen_kwargs={'shell': True})
```

raybellwaves:
Thanks for the discussion on this. I'll wait until a decision is made. Unfortunately, LSF seems to be the black sheep and it looks as though it will have to handled specifically. A `launch_command` may work that way I won't break `cluster.stop_workers(workers)` which expects the normal `subprocess.Popen`.

I can work on the Docker files in the meantime. I've not used docker before so if anyone can point me to some resources for setting up LSF in docker that would be appreciated.

Lastly, I'll play catch up with the latest PR once the method is agreed.

mrocklin:
Something like what @lesteve proposes seems reasonable to me.

@raybellwaves what do you think we should do?

raybellwaves:
Thanks for the suggestion @lesteve. I believe i've got it working for myself and hopefully it should still work for others. I'll have to test tomorrow though as the queue is jammed.

lesteve:
Great to hear, let us know when you think this is ready for review!

mrocklin:
Thanks!  Someone was asking me about LSF support just yesterday, so I know
that this will be a welcome change!

dask_jobqueue/core.py
```diff
@@ -210,12 +214,21 @@ def job_file(self):
...
             yield fn
 
+    def submit_job(self, script_filename):
+        """ Sumbits job and handles lsf exception """
+        if self.scheduler_name == 'lsf':
+            piped_cmd = [self.submit_command + '<\"' + script_filename + '\"']
+            self.shell = True
+            return self._call(piped_cmd)
+        else:
+            return self._call(shlex.split(self.submit_command) + [script_filename])
```
mrocklin: Would it be possible to keep the core solution simple and instead put the LSF-specific implementation on the `LSFCluster` class?

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,117 @@
...
+        if project is None:
+            project = dask.config.get('jobqueue.%s.project' % self.scheduler_name)
+        if ncpus is None:
+            ncpus = dask.config.get('jobqueue.%s.ncpus' % self.scheduler_name)
+        if mem is None:
+            memory = dask.config.get('jobqueue.%s.mem' % self.scheduler_name)
+        if walltime is None:
+            walltime = dask.config.get('jobqueue.%s.walltime' % self.scheduler_name)
+        if job_extra is None:
+            job_extra = dask.config.get('jobqueue.lsf.job-extra')
```
mrocklin: Should this also use the `self.scheduler_name` pattern as above?

raybellwaves: Yep

dask_jobqueue/core.py
raybellwaves: Ah I see what @lesteve was suggesting

lesteve:
Failures on SGE and PBS seem genuine, can you have a look? If I had to guess I would say that `.shell` does not exist in the base class and it probably should. Here is the error on SGE for completeness: 
```
=================================== FAILURES ===================================
__________________________________ test_basic __________________________________
loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7feb64d688d0>
    def test_basic(loop):  # noqa: F811
        with SGECluster(walltime='00:02:00', cores=8, processes=4, memory='28GB',
                        loop=loop) as cluster:
            with Client(cluster, loop=loop) as client:
>               workers = cluster.start_workers(2)
dask_jobqueue/tests/test_sge.py:16: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
dask_jobqueue/core.py:226: in start_workers
    out = self.submit_job(fn)
dask_jobqueue/core.py:219: in submit_job
    return self._call(shlex.split(self.submit_command) + [script_filename])
dask_jobqueue/core.py:277: in _call
    return self._calls([cmd])[0]
dask_jobqueue/core.py:265: in _calls
    for cmd in cmds]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.0 = <list_iterator object at 0x7feb6417cfd0>
    procs = [subprocess.Popen(cmd,
                              shell=self.shell,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
>            for cmd in cmds]
E   AttributeError: 'SGECluster' object has no attribute 'shell'
dask_jobqueue/core.py:265: AttributeError
================ 1 failed, 15 passed, 6 skipped in 1.66 seconds ================
```

raybellwaves:
Not sure if I need to add it to `moab.py` or it'll pick it up from the `PBSCluster`.

lesteve:
Maybe one slightly simpler way of doing it would be to use a class variable `shell = False` in LocalCluster and to set the class variable `shell = True` in LSFCluster.

dask_jobqueue/core.py
```diff
@@ -87,6 +90,8 @@ class JobQueueCluster(Cluster):
     submit_command = None
     cancel_command = None
     scheduler_name = ''
+    
+    shell = False # Required for submitting jobs with LSF
```
mrocklin: Ideally we wouldn't mention LSF here in the superclass.  We would say "required for submitting jobs that require full shell access" or something else like that (I don't actually understand the issue behind `shell`, so I recommend finding better language.)

mrocklin:
@guillaumeeb @lesteve ok to merge?

dask_jobqueue/core.py
```diff
@@ -273,6 +286,7 @@ def stop_workers(self, workers):
             return
         workers = list(map(int, workers))
         jobs = [self.jobs[w] for w in workers]
+        self.shell = False
```
lesteve: You probably don't need this.

dask_jobqueue/sge.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
         super(SGECluster, self).__init__(**kwargs)
 
         header_lines = ['#!/bin/bash']
-
+        # SGE header build
```
lesteve: General piece of advice: do your best to keep a PR focussed and avoid unrelated changes. It's fine to keep it like this for this PR.

dask_jobqueue/core.py
```diff
@@ -88,6 +91,11 @@ class JobQueueCluster(Cluster):
     cancel_command = None
     scheduler_name = ''
 
+    # Required for excuting commands through the shell in the subprocess module.
+    # It handles the shell input redirection e.g. bsub < script_filename.sh
+    # and does not consider '<' as a command, file or directory.
+    shell = False
```
lesteve: Personally I would not use a comment in the base class.

Also I would go for a more informative variable name, maybe popen_shell = True (better name more than welcome)?

dask_jobqueue/core.py
```diff
@@ -210,12 +218,16 @@ def job_file(self):
                 f.write(self.job_script())
             yield fn
 
+    def submit_job(self, script_filename):
+        """ Sumbits job """
```
lesteve: Personally I am in favour of not having a docstring for this function since this is just a one-liner. More generally I think if your docstring is just repeating the function name, remove it.

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,127 @@
...
+        # Declare class attribute that shall be overriden
+        self.job_header = '\n'.join(header_lines)
+
+        logger.debug("Job script: \n %s" % self.job_script())
+
+    def _job_id_from_submit_output(self, out):
+        return out.split('<')[1].split('>')[0].strip()
+
+    def submit_job(self, script_filename):
+        """ Sumbits job and handles lsf exception """
```
lesteve: Remove this docstring too, same typo interestingly :-)

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,127 @@
...
+        piped_cmd = [self.submit_command + '<\"' + script_filename + '\"']
+        return self._call(piped_cmd)
+
+
+def lsf_format_bytes_ceil(n):
+    """ Format bytes as text
+    LSF expects megabytes
```
lesteve: I don't really know what LSF expects but for MB you have to divide by 1024**2 but by 1000**2

guillaumeeb:
On vacation right now, no computer access (and don't want one :) ). Thanks for the work @raybellwaves.

raybellwaves:
@lesteve I realized I have to switch between `shell_popen = True` for `start_workers` and `shell_popen = False` for `stop_workers`. e.g.:

```
In [1]: import subprocess
In [5]: proc = subprocess.Popen(['bsub<"tmp.sh"'])
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-5-601b6e3aded3> in <module>()
----> 1 proc = subprocess.Popen(['bsub<"tmp.sh"'])

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/subprocess.py in __init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags, restore_signals, start_new_session, pass_fds, encoding, errors)
    707                                 c2pread, c2pwrite,
    708                                 errread, errwrite,
--> 709                                 restore_signals, start_new_session)
    710         except:
    711             # Cleanup if the child failed starting.

~/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/subprocess.py in _execute_child(self, args, executable, preexec_fn, close_fds, pass_fds, cwd, env, startupinfo, creationflags, shell, p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite, restore_signals, start_new_session)
   1342                         if errno_num == errno.ENOENT:
   1343                             err_msg += ': ' + repr(err_filename)
-> 1344                     raise child_exception_type(errno_num, err_msg, err_filename)
   1345                 raise child_exception_type(err_msg)
   1346 

FileNotFoundError: [Errno 2] No such file or directory: 'bsub<"tmp.sh"': 'bsub<"tmp.sh"'

In [6]: proc = subprocess.Popen(['bsub<"tmp.sh"'], shell=True)

In [7]: Job is submitted to <cpp> project.
Job <16818485> is submitted to default queue <general>.

In [8]: proc = subprocess.Popen(['bkill', '16818485'], shell=True)

In [9]: Job ID or one of '-m', '-u', '-q', '-J', '-g', '-sla' and '-app' must be specified.
Usage: bkill   [-h] [-V] [-l] [-b] [-m host_name] [-q queue_name] [-g job_group] [-sla service_class] [-app application_profile]
               [-u user_name | -u all] [0] [-r | -s sigval] [-J job_name]
               [jobId | "jobId[index_list]"]...

In [9]: proc = subprocess.Popen(['bkill', '16818485'])

In [10]: Job <16818485>: Job has already finished
```
so I created a `kill_jobs` function. That way I don't have the `self.popen_shell = False` in `core.py` twice.

raybellwaves:
Lastly, a very minor thing (i'm inclined to add this to an issue and get help from other LSF users once it's merged) is to find a way to suppress the output line saying which project the job has been submitted to.

Currently when I do `cluster.scale(n)` I get the output:
```
In [3]: cluster.scale(5)

Job is submitted to <cpp> project.

Job is submitted to <cpp> project.

Job is submitted to <cpp> project.

Job is submitted to <cpp> project.

Job is submitted to <cpp> project.

In [4]: 
```
It just adds clutter to the screen.

This is because when you do `bsub < script_filename.sh` you get the output
```
$ bsub < tmp.sh
Job is submitted to <PROJECT> project.
Job <JOBID> is submitted to default queue <QUEUE>.
```

and https://github.com/dask/dask-jobqueue/blob/5bf3e01c5db2509a49eb5f0afc48aefb73eefc07/dask_jobqueue/core.py#L218 is capturing the second line in my case. Which it why https://github.com/raybellwaves/dask-jobqueue/blob/462110a529895f986feda62863ae132a6cc57f8e/dask_jobqueue/lsf.py#L113 works in getting the job id.

mrocklin:
@jhamman are you able to help with the merge conflicts here?  I suspect that you are the most familiar with these changes due to the recent large PR that just went in.

jhamman:
@raybellwaves - I just pushed a merge resolution to your branch.

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,131 @@
...
+        if mem is not None:
+            memory_string = lsf_format_bytes_ceil(mem)
+            header_lines.append('#BSUB -M %s' % memory_string)
+        if walltime is not None:
+            header_lines.append('#BSUB -W %s' % walltime)
+        header_lines.extend(['#BSUB %s' % arg for arg in job_extra])
+
+        # Declare class attribute that shall be overriden
+        self.job_header = '\n'.join(header_lines)
+
```
jhamman: you'll need to add a line here to set the `JOB_ID` environment variable. I think this will come from the lsf variable: `LSB_JOBID`. The PBSCluster has a good example that you can follow.

dask_jobqueue/sge.py
```diff
@@ -58,7 +58,8 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
 
         super(SGECluster, self).__init__(**kwargs)
 
-        header_lines = ['#!/usr/bin/env bash']
+        header_lines = ['#!/bin/bash']
```
jhamman: can you change this back to `#!/usr/bin/env bash`? I must have missed this in the merge conflict resolution.

jhamman:
@raybellwaves, regarding the project log messages you see above, you may want to see if lsf has a command line option to limit its verbosity. Either that or we need a mechanism to capture this stream. You may do some simple tests on the command line to figure out if this message is coming from stdout or stderr. 

Finally, now that I've merged master into your branch, you'll probably want to spend some time testing again on your system.

mrocklin:
Sorry for the slow turnaround time on this @raybellwaves !  I'd like to get
this in soon so that we can avoid future merge conflicts.

raybellwaves:
@jhamman thanks. Although i'm having issues getting the `job_id`. Am I missing something?

When I do
```
In [2]: cluster.start_workers(2)
Job is submitted to <cpp> project.

Job is submitted to <cpp> project.
```
I get the below error message after about two seconds:
```
distributed.scheduler - ERROR - ''
  File "/nethome/rxb826/local/bin/miniconda3/envs/d-jq-test/lib/python3.6/site-packages/distributed/scheduler.py", line 1266, in add_worker
    plugin.add_worker(scheduler=self, worker=address)
  File "/nethome/rxb826/PYTHON3/dask-jobqueue/dask_jobqueue/core.py", line 66, in add_worker
    self.running_jobs[job_id] = self.pending_jobs.pop(job_id)
KeyError: ''
```
I was printing `w` at https://github.com/dask/dask-jobqueue/blob/f7c565a2fe87e8829acba3c752a4da8d196c92ca/dask_jobqueue/core.py#L55
and it was showing `tcp://10.10.107.15:54369`. `w.name` is then returning empty.

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,136 @@
...
+            mem = self.worker_memory
+            logger.info("mem specification for LSF not set, "
+                        "initializing it to %s" % mem)
+        if mem is not None:
+            memory_string = lsf_format_bytes_ceil(mem)
+            header_lines.append('#BSUB -M %s' % memory_string)
+        if walltime is not None:
+            header_lines.append('#BSUB -W %s' % walltime)
+        header_lines.extend(['#BSUB %s' % arg for arg in job_extra])
+        header_lines.append('JOB_ID=${LSF_JOBID%.*}')
```
jhamman: isn't this supposed to be: `LSB_JOBID`

xref: https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.2/lsf_config_ref/lsf_envars_job_exec.html

raybellwaves:
@jhamman good spot. Everything is working again.
I'll spend a little time on suppressing the `Job is submitted to <PROJECT> project.` output before another review.

raybellwaves:
The `Job is submitted to <PROJECT> project.` was coming from the `err` https://github.com/dask/dask-jobqueue/blob/f7c565a2fe87e8829acba3c752a4da8d196c92ca/dask_jobqueue/core.py#L330

I worked out how to suppress it by creating a `_error_capture()` function and create a check in `lsf.py`: https://github.com/raybellwaves/dask-jobqueue/blob/c1f85fd27b76d59687bc42f971cc3c5509d97833/dask_jobqueue/lsf.py#L132 the solution came from some SO answers. Happy to hear cleaner suggestions for my work around.

Ready for review.

raybellwaves:
Pinging @mrocklin @jhamman. Sorry to bother you. This is ready for review when you have time.

dask_jobqueue/core.py
```diff
@@ -280,13 +282,16 @@ def job_file(self):
                 f.write(self.job_script())
             yield fn
 
+    def submit_job(self, script_filename):
+        return self._call(shlex.split(self.submit_command) + [script_filename])
```
mrocklin: This should probably be a private method so that users don't get the wrong idea that they should use it to submit jobs.

dask_jobqueue/jobqueue.yaml
```diff
@@ -83,3 +83,25 @@ jobqueue:
...
+
+    # LSF resource manager options
+    queue: null
+    project: null
+    walltime: '00:30'
+    extra: ""
+    env-extra: []
+    ncpus: null
+    mem: null
+    job-extra: []
```
mrocklin: Just checking in, does LSF need/use all of these options?  I'm slightly concerned that as we copy configs from different systems we may accrue more than is necessary.

dask_jobqueue/lsf.py
mrocklin: @raybellwaves any response to this comment?

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,142 @@
...
+    def _job_id_from_submit_output(self, out):
+        return out.split('<')[1].split('>')[0].strip()
+
+    def submit_job(self, script_filename):
+        """ set `self.popen_shell = True` and redirect the `script_filename` to bsub.
+        Output of bsub normally looks like:
+        `Job is submitted to <PROJECT> project.` which is the stderr and
+        `Job <JOBID> is submitted to (default) queue <QUEUE>.` which is the stdout.
+        Supress the stderr by redirecting it to nowhere.
+        The `piped_cmd` looks like ['bsub < tmp.sh 2> /dev/null'] """
```
mrocklin: This docstring looks a bit wonky.  It looks more like a developer comment than a documentation string for users.  If it *is* supposed to be a docstring then you might want to follow the numpydoc standard, or take a look at the dask developer notes on docstrings: http://dask.pydata.org/en/latest/develop.html#docstrings

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,142 @@
...
+        self.popen_shell = True
```
mrocklin: It seems odd to set this here again.  It was already set above on the class, right?

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,142 @@
...
+        piped_cmd = [self.submit_command+' < '+script_filename+' 2> /dev/null']
```
mrocklin: There are some minor style issues.  I recommend running flake8 on the codebase.

```
mrocklin@carbon:~/workspace/dask-jobqueue$ flake8 dask_jobqueue
dask_jobqueue/lsf.py:5:1: F401 'os' imported but unused
dask_jobqueue/lsf.py:69:13: F841 local variable 'memory' is assigned to but never used
dask_jobqueue/lsf.py:124:41: E226 missing whitespace around arithmetic operator
dask_jobqueue/lsf.py:124:47: E226 missing whitespace around arithmetic operator
dask_jobqueue/lsf.py:124:63: E226 missing whitespace around arithmetic operator
```

mrocklin: I've opened https://github.com/dask/dask-jobqueue/issues/106 to discuss including this in the CI

dask_jobqueue/lsf.py
mrocklin: @raybellwaves any response to this comment?

mrocklin: Also it would be good to format this docstring like http://dask.pydata.org/en/latest/develop.html#docstrings

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,142 @@
...
+        return self._call(piped_cmd)
+
+    def stop_jobs(self, jobs):
+        """ set `self.popen_shell = False` """
+        self.popen_shell = False
```
mrocklin: Ah, I see, you're using the class state to sneak a parameter into the `._call` method.  I think it's probably better to pass extra keyword arguments into the `._call` method directly and avoid the extra state.  I'll write a comment on this in the core.py file.

dask_jobqueue/core.py
```diff
@@ -322,6 +327,7 @@ def _calls(self, cmds):
         for cmd in cmds:
             logger.debug(' '.join(cmd))
             procs.append(subprocess.Popen(cmd,
+                                          shell=self.popen_shell,
```
mrocklin: I recommend that we pass `**kwargs`, or at least `shell=` from the `_call` method into this function call and avoid the state entirely.

```python
def _call(self, cmd, **kwargs):
    return self._calls([cmd], **kwargs)

def _calls(self, cmds, **kwargs):
    ...
    procs.append(subprocess.Popen(cmd,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE,
                                  **kwargs))
```

Then we can call this like `self._call(cmd, shell=True)` and avoid mucking about with the `popen_shell` state.

mrocklin:
Then lets drop them

dask_jobqueue/lsf.py
```diff
@@ -0,0 +1,137 @@
...
+    def _submit_job(self, script_filename):
+        piped_cmd = [self.submit_command + ' ' + script_filename + ' 2> /dev/null']
+        return self._call(piped_cmd, shell=True)
+
+    def stop_jobs(self, jobs):
+        """ Stop a list of jobs"""
+        logger.debug("Stopping jobs: %s" % jobs)
+        if jobs:
+            jobs = list(jobs)
+            self._call([self.cancel_command] + list(set(jobs)), shell=False)
```
guillaumeeb: As there is no state anymore, and `shell=False` is default with `popen`, you probably don't need it here, and thus you probably don't need to redefine `stop_jobs` in Lsf implementation.

raybellwaves: Thanks. Good spot.

mrocklin:
This now seems fine to me.  @raybellwaves can you verify that things work well on your LSF cluster?

raybellwaves:
screen shot 2018-07-31 at 4 45 32 pm

raybellwaves:
We can also get feedback from folks in https://github.com/dask/dask-jobqueue/issues/4 after merging

mrocklin:
Merging this tomorrow if there are no further comments.  If anyone gets to this before I do and are ok with things I encourage you to merge if you think it's ready.

jhamman:
Thanks @raybellwaves for sticking with this. The only thing I see this needing is a few docs.

See the {index,examples,configurations,api}.rst files for some good places to document the LSFCluster. Bare minimum would be `api.rst`.

mrocklin:
Looks like the api.rst change would be a single line.  I'm in favor of that
:)  It'd be nice to get this in.

mrocklin:
This is in.  Thank you @raybellwaves for implementing this!  I think that it will be very valuable

lesteve:
Very nice!

PR #79, opened 2018-06-27
Rename JobQueueCluster cluster attribute to local_cluster

lesteve:
This avoids things like `client.cluster.cluster` to access the `LocalCluster` object (yes `.cluster` appears twice, not a typo) which is a bit confusing.

I used the same attribute name as in dask-drmaa [here](https://github.com/dask/dask-drmaa/blob/c909d153ab3eafe354f3a55a5eb696246b937643/dask_drmaa/core.py#L161).

I reckon this is a bit of an implementation detail and is unlikely to affect anyone but I may be wrong about this so do share you concerns if you have some!

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -126,7 +126,7 @@ class JobQueueCluster(Cluster):
         else:
             host = socket.gethostname()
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Fine with me

lesteve:
Merging this one, since there seems to be consensus.

PR #86, opened 2018-06-28
Add cores and total_memory parameters

mrocklin:
These overlap with (and might possibly replace) the threads and memory
parameters.  I believe that they are more consistent with typcial use.

Now we ask people to specify the cores and total memory of a node.  This
should remain relatively static over time.  Then we allow them to tune
processes as they like independently of other parameters.

### Example

```python
# Before
cluster = PBSCluster(threads=4, processes=9, memory='10GB')
cluster = PBSCluster(threads=12, processes=3, memory='30GB')
```
```python
# After
cluster = PBSCluster(cores=36, total_memory='90GB', processes=9)
cluster = PBSCluster(cores=36, total_memory='90GB', processes=3)
# or just
cluster = PBSCluster(cores=36, total_memory='90GB')
```

I've found that novices are more comfortable answering the questions on the bottom rather than the questions on the top.  Answering the questions on the top means that you already need to understand how dask-worker manages its keywords.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -138,7 +138,11 @@ class JobQueueCluster(Cluster):
 
         if cores is None:
             raise ValueError("You must specify how many cores to use per job "
-                             "with the cores= parameter")
+                             "like ``cores=8``")
+
+        if memory is None:
+            raise ValueError("You must specify how much memory to use per job "
+                             "like ``memory='24 GB'``")
 
         #This attribute should be overriden
         self.job_header = None
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -4,12 +4,14 @@ import socket
 import subprocess
 import sys
 from contextlib import contextmanager
+import multiprocessing
 
 import dask
 import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
-from distributed.utils import get_ip_interface, ignoring, parse_bytes, tmpfile
+from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
+        format_bytes)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
@@ -26,13 +28,12 @@ class JobQueueCluster(Cluster):
     ----------
     name : str
         Name of Dask workers.
-    threads : int
-        Number of threads per process.
+    cores : int
+        Total number of cores on a node / job
+    total_memory: str
+        Total amount of memory on a node / job
     processes : int
         Number of processes per node.
-    memory : str
-        Bytes of memory that the worker can use. This should be a string
-        like "7GB" that can be interpretted both by PBS and Dask.
     interface : str
         Network interface like 'eth0' or 'ib0'.
     death_timeout : float
@@ -45,6 +46,19 @@ class JobQueueCluster(Cluster):
         Other commands to add to script before launching worker.
     kwargs : dict
         Additional keyword arguments to pass to `LocalCluster`
+    memory : str
+        Bytes of memory that the worker can use. This should be a string
+        like "7GB" that can be interpretted both by PBS and Dask.
+
+        This is redundant with total_memory and processes.  If specifying
+        total_memory you should not specify memory, which should interpretted
+        as memory per process.
+    threads : int
+        Number of threads per process.
+
+        This is redundant with cores and proceses.  If specifying cores you
+        should not specify threads, which should be interpretted as threads per
+        process.
 
     Attributes
     ----------
@@ -78,15 +92,17 @@ class JobQueueCluster(Cluster):
 
     def __init__(self,
                  name=None,
-                 threads=None,
+                 cores=None,
+                 total_memory=None,
                  processes=None,
-                 memory=None,
                  interface=None,
                  death_timeout=None,
                  local_directory=None,
                  extra=None,
                  env_extra=None,
                  walltime=None,
+                 threads=None,
+                 memory=None,
                  **kwargs
                  ):
         """ """
@@ -100,6 +116,10 @@ class JobQueueCluster(Cluster):
 
         if name is None:
             name = dask.config.get('jobqueue.%s.name' % self.scheduler_name)
+        if cores is None:
+            cores = dask.config.get('jobqueue.%s.cores' % self.scheduler_name)
+        if total_memory is None:
+            total_memory = dask.config.get('jobqueue.%s.total_memory' % self.scheduler_name)
         if threads is None:
             threads = dask.config.get('jobqueue.%s.threads' % self.scheduler_name)
         if processes is None:
@@ -117,6 +137,12 @@ class JobQueueCluster(Cluster):
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
 
+        if cores is not None and threads is not None and processes is not None:
+            raise ValueError("You can only specify two of cores, threads, and processes")
+
+        if total_memory is not None and memory is not None:
+            raise ValueError("You can only specify one of total_memory and memory")
+
         #This attribute should be overriden
         self.job_header = None
 
@@ -130,9 +156,15 @@ class JobQueueCluster(Cluster):
 
         # Keep information on process, threads and memory, for use in
         # subclasses
-        self.worker_memory = parse_bytes(memory) if memory is not None else None
+        if total_memory is not None:
+            self.worker_memory = parse_bytes(total_memory) / processes
+        elif memory is not None:
+            self.worker_memory = parse_bytes(memory)
+        else:
+            self.worker_memory = None
+
         self.worker_processes = processes
-        self.worker_threads = threads
+        self.worker_threads = cores / processes if cores and processes else threads
         self.name = name
 
         self.jobs = dict()
@@ -145,12 +177,12 @@ class JobQueueCluster(Cluster):
         dask_worker_command = (
             '%(python)s -m distributed.cli.dask_worker' % dict(python=sys.executable))
         self._command_template = ' '.join([dask_worker_command, self.scheduler.address])
-        if threads is not None:
-            self._command_template += " --nthreads %d" % threads
+        if self.worker_threads is not None:
+            self._command_template += " --nthreads %d" % self.worker_threads
         if processes is not None:
             self._command_template += " --nprocs %d" % processes
-        if memory is not None:
-            self._command_template += " --memory-limit %s" % memory
+        if self.worker_memory is not None:
+            self._command_template += " --memory-limit %s" % format_bytes(self.worker_memory).replace(' ', '')
         if name is not None:
             self._command_template += " --name %s" % name
             self._command_template += "-%(n)d" # Keep %(n) to be replaced later
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -1,9 +1,10 @@
+from contextlib import contextmanager
 import logging
 import shlex
 import socket
 import subprocess
 import sys
-from contextlib import contextmanager
+import warnings
 
 import dask
 import docrep
@@ -14,6 +15,17 @@ from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
+
+
+threads_deprecation_message = """
+The threads keyword has been removed and the memory keyword has changed.
+
+Please specify job size with the following keywords:
+
+-  cores: total cores per job, across all processes
+-  memory: total memory per job, across all processes
+-  processes: number of processes to launch, splitting the quantities above
+""".strip()
 
 
 @docstrings.get_sectionsf('JobQueueCluster')
@@ -87,6 +99,7 @@ class JobQueueCluster(Cluster):
                  extra=None,
                  env_extra=None,
                  walltime=None,
+                 threads=None,
                  **kwargs
                  ):
         """ """
@@ -94,6 +107,9 @@ class JobQueueCluster(Cluster):
         # This initializer should be considered as Abstract, and never used
         # directly.
         # """
+        if threads is not None:
+            raise ValueError(threads_deprecation_message)
+
         if not self.scheduler_name:
             raise NotImplementedError('JobQueueCluster is an abstract class '
                                       'that should not be instanciated.')
@@ -116,6 +132,9 @@ class JobQueueCluster(Cluster):
             extra = dask.config.get('jobqueue.%s.extra' % self.scheduler_name)
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
+
+        if dask.config.get('jobqueue.%s.threads', None):
+            warnings.warn(threads_deprecation_message)
 
         if cores is None:
             raise ValueError("You must specify how many cores to use per job "
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -4,7 +4,6 @@ import socket
 import subprocess
 import sys
 from contextlib import contextmanager
-import multiprocessing
 
 import dask
 import docrep
@@ -29,11 +28,11 @@ class JobQueueCluster(Cluster):
     name : str
         Name of Dask workers.
     cores : int
-        Total number of cores on a node / job
+        Total number of cores per job
     total_memory: str
-        Total amount of memory on a node / job
+        Total amount of memory per job
     processes : int
-        Number of processes per node.
+        Number of processes per job
     interface : str
         Network interface like 'eth0' or 'ib0'.
     death_timeout : float
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -10,7 +10,7 @@ import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
 from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
-        format_bytes)
+                               format_bytes)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
@@ -29,7 +29,7 @@ class JobQueueCluster(Cluster):
         Name of Dask workers.
     cores : int
         Total number of cores per job
-    total_memory: str
+    memory: str
         Total amount of memory per job
     processes : int
         Number of processes per job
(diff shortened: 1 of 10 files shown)


mrocklin:
This also changes the default processes/threads in the config file to `null` to imply that we don't have any information.  This seems like a more accurate default.

Alternatively, we could also use `multiprocessing.cpu_count()` for cores and something similar for memory from `psutil`.  This would assume that the login node is similar to the worker nodes though, I'm not sure how common this is.

mrocklin:
OK, I've also added some non-trivial documentation changes here.  Review of all of this would be welcome.  I like these changes, but they are a significant deviation from current behavior.

lesteve:
In terms of API I agree that this feels more natural to someone already familiar with cluster stuff (and not so familiar with dask). I'll try to look at this one in more details but here is a quick comment:

> Alternatively, we could also use multiprocessing.cpu_count() for cores and something similar for memory from psutil. This would assume that the login node is similar to the worker nodes though, I'm not sure how common this is.

I don't see a good way to set the default here so I am fine with null for threads/processes. In particular: the login node has a significantly smaller processing power than the worker nodes (at least in all the setups I have come across). Even if we could somehow use cpu_count and psutil on the worker node, they are not aware of the scheduler constraints passed into the submit command.

mrocklin:
When the user doesn't provides cores/memory should we err informatively or
should we not include a resource spec?  Dask will use sane defaults when it
gets to the worker node (using psutil then, which will presumably make more
sense) but the job scheduler won't know what we're asking for.

mrocklin:
Also, if we felt like making a larger change, we could also replace the use of the term `memory` from "memory of a single worker" to "memory of a single node".  This would be a confusing breaking change but, given the small number of users today, this might be acceptable.

This would make things, I think, more clear.

```python
cluster = PBSCluster(cores=36, memory="100GB", processes=4)
```

I would also remove the threads term entirely.

guillaumeeb:
A few comments on the discussion above before review:

>the login node has a significantly smaller processing power than the worker nodes

In our case, login nodes are used also for dev purpose and graphical display, so they have bigger processing power 🙂. Anyway, I agree with @lesteve, no good way to setting defaults here.

>When the user doesn't provides cores/memory should we err informatively 

My point of view is that we should err informatively. Upcomming errors could be much more difficult to analyse. On our cluster, default resources spec (when nothing is specified) is `select=1;ncpus=1;mem=5GB`. If we start a worker with these resources spec, and it begin using all the cores of the node it runs on, it will get killed few seconds after. Hard to debug for day to day user.

> we could also replace the use of the term memory from "memory of a single worker" to "memory of a single node". This would be a confusing breaking change 

I'm not sure about that, can't decide yet. Could we at least keep the old `memory` keyword as a new one for a while? `memory_per_worker` for example.

> I would also remove the threads term entirely

This would prevent from doing "over threading":
```python
cluster = PBSCluster(cores=24, memory="100GB", processes=12, threads=4)
```
don't know if it is used a lot though.

guillaumeeb:
> This would prevent from doing "over threading"

I am probably mistaken here, this will still be possible using resource_spec.

mrocklin:
> Could we at least keep the old memory keyword as a new one for a while? memory_per_worker for example

We could, but I'm inclined to make a clean break if it's possible.  However, if there is a use case for `memory_per_worker` then we should keep it.  Is there such a use case?

dask_jobqueue/core.py
```diff
@@ -117,6 +137,12 @@ def __init__(self,
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
 
+        if cores is not None and threads is not None and processes is not None:
+            raise ValueError("You can only specify two of cores, threads, and processes")
+
+        if total_memory is not None and memory is not None:
+            raise ValueError("You can only specify one of total_memory and memory")
+
```
guillaumeeb: Hm, I think I'm beginning to be incline to remove `threads` and `memory`. At least use `memory` keyword instead of `total_memory` (with a warning comment).

dask_jobqueue/jobqueue.yaml
```diff
@@ -3,12 +3,13 @@ jobqueue:
     name: dask-worker
 
     # Dask worker options
-    threads: 2
-    processes: 4
-    memory: 8GB
-    interface: null
-    death-timeout: 60
-    local-directory: null
+    cores: null                 # Total number of cores on a worker node
```
guillaumeeb: Same remark for the _node_ in comment. Should'nt we use the term _job_?

dask_jobqueue/pbs.py
```diff
@@ -83,15 +83,16 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
             header_lines.append('#PBS -q %s' % queue)
         if project is not None:
             header_lines.append('#PBS -A %s' % project)
-        if resource_spec is None:
+        if resource_spec is None and self.worker_processes and self.worker_threads:
```
guillaumeeb: As discussed, I believe we should fail fast if processes and cores are not provided, so these checks could be removed. Nevertheless, they don't hurt, so we can keep them!

dask_jobqueue/jobqueue.yaml
guillaumeeb: More importantly, I think we should provide a default cores number, especially if we fail if not given.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -34,7 +34,7 @@ def test_header(Cluster):
 
         assert '#PBS -j oe' not in cluster.job_header
         assert '#PBS -N' in cluster.job_header
-        assert '#PBS -l select=1:ncpus=' in cluster.job_header
+        # assert '#PBS -l select=1:ncpus=' in cluster.job_header
```
guillaumeeb: If we fail fast if enough information not provided, and if we give a default cores number, this assert should be ok.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -43,26 +43,27 @@ def test_header(Cluster):
...
 
 @pytest.mark.parametrize('Cluster', [PBSCluster, MoabCluster])
 def test_job_script(Cluster):
-    with Cluster(walltime='00:02:00', processes=4, threads=2, memory='7GB') as cluster:
+    with Cluster(walltime='00:02:00', processes=4, cores=8, total_memory='28GB') as cluster:
 
         job_script = cluster.job_script()
         assert '#PBS' in job_script
         assert '#PBS -N dask-worker' in job_script
+        assert '--nthreads 2' in job_script
```
guillaumeeb: Redundant with l.66

dask_jobqueue/tests/test_slurm.py
```diff
@@ -35,19 +35,22 @@ def test_header():
...
 def test_job_script():
-    with SLURMCluster(walltime='00:02:00', processes=4, threads=2, memory='7GB') as cluster:
+    with SLURMCluster(walltime='00:02:00', processes=4, cores=8,
+                      total_memory='28GB') as cluster:
 
         job_script = cluster.job_script()
         assert '#SBATCH' in job_script
         assert '#SBATCH -J dask-worker' in job_script
+        assert '--nthreads 2 ' in job_script
+        assert '--memory-limit 7.00GB ' in job_script
```
guillaumeeb: Redundant with l64

docs/configurations.rst
```diff
@@ -0,0 +1,42 @@
+Configuration Examples
+======================
+
+We include configuration files for known supercomputers.
+Hopefully these help both other users that use those machines and new users who
+want to see examples for similar clusters.
+
+Additional examples from other cluster welcome `here <https://github.com/dask/dask-jobqueue/issues/40>`_.
+
+Cheynne
```
guillaumeeb: typo here

docs/index.rst
```diff
@@ -2,74 +2,162 @@
 Dask-Jobqueue
 =============
 
-*Easy deployment of Dask Distributed on job queuing systems like
-PBS, Slurm, and SGE.*
+*Easily deploy Dask on job queuing systems like PBS, Slurm, MOAB, and SGE.*
```
guillaumeeb: Just curious, why did you remove _distributed_, and only kept Dask?

docs/index.rst
```diff
@@ -2,74 +2,162 @@
...
 
 .. code-block:: python
 
    from dask_jobqueue import PBSCluster
+   cluster = PBSCluster()
+   cluster.scale(10)         # Ask for ten worker nodes
 
-   cluster = PBSCluster(processes=6, threads=4, memory="16GB")
-   cluster.start_workers(10)
+   # wait for jobs to arrive, depending on the queue, this may take some time
```
guillaumeeb: Wait can be done after the client creation (as you very well known), I often run a cell with client on Jupyter to know when the worker are connected. But maybe it is better for user understanding to put it here.

docs/index.rst
```diff
@@ -2,74 +2,162 @@
...
 
    from dask.distributed import Client
-   client = Client(cluster)
+   client = Client(cluster)  # Connect Dask to the worker nodes
```
guillaumeeb: Here again I take it the comment is a simplification that you believe is more appropriate for common users?

docs/index.rst
```diff
@@ -2,74 +2,162 @@
...
 .. code-block:: python
 
    cluster.adapt(minimum=1, maximum=100)
 
+
+Configuration
+-------------
+
+Dask-jobqueue should be configured to your cluster so that it knows how many
+resources to request of each node and how to break up those resources.  You can
```
guillaumeeb: Again the node vs job point of view. Not sure what we should say. At least mention the term job somewhere?

dask_jobqueue/core.py
```diff
@@ -26,13 +28,12 @@ class JobQueueCluster(Cluster):
     ----------
     name : str
         Name of Dask workers.
-    threads : int
-        Number of threads per process.
+    cores : int
+        Total number of cores on a node / job
+    total_memory: str
+        Total amount of memory on a node / job
```
guillaumeeb: > if there is a use case for memory_per_worker then we should keep it. Is there such a use case?

I can't find such a use case. However this makes me think we should put in the comment here that the total_memory will be splitted among worker processes. There won't be a process handling all this memory alone, I think it is better if it is clear.

dask_jobqueue/jobqueue.yaml
mrocklin: What should the default cores number be?  Currently I'm somewhat inclined to fail and ask them to specify with a well worded error message.

docs/index.rst
mrocklin: The name of the project is Dask.  It happens to have a local and a distributed scheduler but these are, from my perspective, not part of the project name.  The fact that there is a repository named `distributed` is a historical artifact and something that I prefer to hide.

docs/index.rst
mrocklin: I've moved this line down below `client = Client(...)`.  I started writing "If you are in a Jupyter notebook then you might want to render `cluster` so that you can see the status of the cluster" but then I realized that this is the header example of the entire project, and so I figured that this was too much detail upfront.

docs/index.rst
mrocklin: Yes

mrocklin: Alternatives welcome

docs/index.rst
mrocklin: Fixed

docs/configurations.rst
mrocklin: Fixed

dask_jobqueue/tests/test_slurm.py
mrocklin: Fixed

dask_jobqueue/tests/test_pbs.py
mrocklin: Fixed

dask_jobqueue/core.py
```diff
@@ -26,13 +28,12 @@ class JobQueueCluster(Cluster):
     ----------
     name : str
         Name of Dask workers.
-    threads : int
-        Number of threads per process.
+    cores : int
+        Total number of cores on a node / job
```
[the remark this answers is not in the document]
mrocklin: Fixed, now using the job term everywhere

dask_jobqueue/core.py
```diff
@@ -4,12 +4,14 @@
 import subprocess
 import sys
 from contextlib import contextmanager
+import multiprocessing
```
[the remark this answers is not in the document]
mrocklin: I added it originally for cpu_count, but then decided to ask first to see if that was a good idea.  Removed it

mrocklin:
OK, I've made the node->job switch.  I'm waiting on agreement behind threads/total_memory to make those changes.

docs/index.rst
guillaumeeb: Right, I totally support this. distributed not really being part of dask is something I've been told when evangelizing dask around me. Much better if they are really shown as one.

mrocklin:
I've tenatively removed threads and memory.  I know that we haven't agreed on this yet, but I wanted to see how it would change things.  I've also added an informative error message.

dask_jobqueue/core.py
```diff
@@ -145,12 +171,12 @@ def __init__(self,
...
-        if threads is not None:
-            self._command_template += " --nthreads %d" % threads
+        if self.worker_threads is not None:
+            self._command_template += " --nthreads %d" % self.worker_threads
         if processes is not None:
             self._command_template += " --nprocs %d" % processes
-        if memory is not None:
-            self._command_template += " --memory-limit %s" % memory
+        if self.worker_memory is not None:
+            self._command_template += " --memory-limit %s" % format_bytes(self.worker_memory).replace(' ', '')
```
jhamman: can we strip the space chars in `format_bytes`?

docs/index.rst
```diff
@@ -2,74 +2,162 @@
...
 
 .. code-block:: python
 
    cluster.adapt(minimum=1, maximum=100)
 
+
+Configuration
+-------------
+
+Dask-jobqueue should be configured to your cluster so that it knows how many
```
jhamman: configured ***for*** your cluster?

docs/index.rst
```diff
@@ -2,74 +2,162 @@
...
+        processes=6,          # Number of Python processes to cut up each job
+        local_directory='$TMPDIR',  # Location to put temporary data if necessary
+        # Job scheduler specific keywords
+        resource_spec='select=1:ncpus=24:mem=100GB',
+        queue='regular',
+        project='my-project',
+        walltime='02:00:00',
+   )
+
+Note that the cores and total_memory keywords above correspond not to your
```
jhamman: suggest using syntax highlighting `cores`/`total_memory` as keywords

docs/index.rst
mrocklin: Fixed.  Thanks

docs/index.rst
mrocklin: Fixed.  Thanks

dask_jobqueue/core.py
mrocklin: Maybe.  This will have to happen in dask/dask regardless.

mrocklin: I'm also not sure that we always want to.  Often when rendering bytes for human readers having a space is nicer.  I'm inclined to leave this as an explicit step.

mrocklin:
I would like to merge this soon if possible.

dask_jobqueue/jobqueue.yaml
```diff
@@ -3,12 +3,13 @@ jobqueue:
...
 
     # Dask worker options
-    threads: 2
-    processes: 4
-    memory: 8GB
-    interface: null
-    death-timeout: 60
-    local-directory: null
+    cores: null                 # Total number of cores per job
+    memory: null                # Total amount of memory per job
```
guillaumeeb: Don't like the memory default to null and thus None in the code. But it is much more difficult to set to a defaut value when it is a global one. 
I would give default core and memory values here, or fail properly if memory is None, as for cores. Values are tightly linked to each other.

dask_jobqueue/pbs.py
```diff
@@ -83,15 +83,18 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
             header_lines.append('#PBS -q %s' % queue)
         if project is not None:
             header_lines.append('#PBS -A %s' % project)
-        if resource_spec is None:
+        if resource_spec is None and self.worker_threads:
             # Compute default resources specifications
             ncpus = self.worker_processes * self.worker_threads
```
guillaumeeb: Should'nt we store cores and memory in jobqueue abstract class. So that we don't have to perform the inverse computation as in https://github.com/dask/dask-jobqueue/pull/86/files#diff-12a0c7384462f5473600598f508cac74R161 ?

I think PBS or Slurm part only need the cores and memory part.

dask_jobqueue/jobqueue.yaml
mrocklin: If we choose defaults then what should they be?  

I'd prefer to raise with informative errors if memory isn't specified.  I'd be happy to make that change.

dask_jobqueue/pbs.py
mrocklin: Can do

dask_jobqueue/jobqueue.yaml
mrocklin: Done

dask_jobqueue/pbs.py
mrocklin: Done

mrocklin:
Any further comments?

dask_jobqueue/pbs.py
```diff
@@ -85,13 +85,15 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
             header_lines.append('#PBS -A %s' % project)
         if resource_spec is None:
             # Compute default resources specifications
-            ncpus = self.worker_processes * self.worker_threads
-            memory = self.worker_memory * self.worker_processes
-            memory_string = pbs_format_bytes_ceil(memory)
-            resource_spec = "select=1:ncpus=%d:mem=%s" % (ncpus, memory_string)
+            resource_spec = "select=1:ncpus=%d" % self.worker_cores
+            if self.worker_memory:
+                memory = self.worker_memory * self.worker_processes
```
guillaumeeb: Same as `self.worker_cores`, I propose to just record and use a `self.worker_total_memory`, or `self.memory`. 
Except if you've done that because of the conversion from string to int and the other way around.

mrocklin: Ah indeed.  That was an oversight.  I've now just changed the definition of worker_memory to mean total memory.

mrocklin:
Thanks for the review.  It's been very helpful

guillaumeeb:
My pleasure, thanks for the work here (and for dask more generally).

jhamman:
> I'd be fine to switch to that. I wouldn't mind it being done in a separate PR though, maybe even just #63 

+1

> I would like to merge this soon if possible.

+1. I'm good with merging this now.

guillaumeeb:
Found some time to test, works perfectly, nice errors when not properly configured.

mrocklin:
Great.  Thanks for the review all.  Merging!

PR #93, opened 2018-07-14
Add doe arm

josephhardinee:
Added the Department of Energy - Atmosheric Radiation Measurement stratus cluster to configuration examples.

### DIFF
--- a/docs/configurations.rst
+++ b/docs/configurations.rst
@@ -44,7 +44,7 @@ Cheyenne
 ARM Stratus
 -----------
 
-`Department of Energy Atmospheric Radiation Measurement (DOE-ARM) Stratus <https://adc.arm.gov/tutorials/cluster/stratusclusterquickstart.html>`_.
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @josephhardinee .  This looks great.  Should we wait to learn if this applies to all of Titan or should we merge sooner?

josephhardinee:
I think merging for now is fine and we'll just call it instructions for ARM's stratus cluster as it's open to quite a few users of dask as well.

mrocklin:
OK, if you do find that this applies to Titan as well then an update would be well appreciated.  Presumably there are users of Titan that won't recognize this name.

Merging.

bstrdsmkr:
No no, these look fine, mostly just wanted to leave a record in case it suddenly stopped working for someone (we're shooting for a timescale in days).

What's the best way to get in touch aside from highjacking @josephhardinee 's PR thread? :)

mrocklin:
We tend to do most of our communication direclty on Github.  I recommend
opening a new issue at https://github.com/dask/dask-jobqueue/issues/new

PR #94, opened 2018-07-14
Add nersc cori configuration example

josephhardinee:
This PR adds a config file example for NERSC's Cori system. We ran into some issues with the filesystem and network IO so we had to disable the swapping to disk, and also added some instructions about running the scheduler on a worker node itself. 

I'll try to touch base with NERSC IT next week to see if they have suggestions on fixing both issues.

### DIFF
--- a/docs/configurations.rst
+++ b/docs/configurations.rst
@@ -41,3 +41,33 @@ Cheyenne
 
        resource-spec: select=1:ncpus=36:mem=109G
 
+NERSC Cori
+--------
+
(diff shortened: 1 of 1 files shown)


mrocklin:
He's already doing this, but we're having trouble with the scratch
directory that we're given

mrocklin:
I'm going to merge this for now.  We can edit it later when we find out about the scratch space.

mrocklin:
Thanks @josephhardinee !  I appreciate your effort here.  I suspect that these instructions will be quite useful for others.

@stefanv if you're interested it might be nice to have someone with access to Cori try things out and verify that things work for someone else.  I know that this isn't your normal system these days, but perhaps you know someone?

PR #97, opened 2018-07-17
more adaptive scaling fixes

### LINKED ISSUE #112: Edge cases of Cluster.scale()
Consider a simple jobqueue cluster:

```Python
cluster = FooCluster(cores=1, processes=1)
```

What should the behavior of the following cases be:

```Python
cluster.scale(2)
# run immediately
cluster.scale(0)
```

Currently, pending jobs are not removed from the cluster and eventually come on line. This is producing some buggy behavior. 

or 

```Python
cluster.scale_up(4)
cluster.scale_up(2)
```

Currently, the second scale_up call is a no-op but 4 workers will come online. Should we update the target number of workers or inform the user that scaling up to beyond 2 is already pending?

jhamman:
@lesteve submitted a review on #63 after we merged. The changes he requested were mostly small so I'm just addressing them here.

closes #112

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -1,17 +1,22 @@
-from contextlib import contextmanager
+from __future__ import absolute_import, division, print_function
+
 import logging
+import math
 import shlex
 import socket
 import subprocess
 import sys
 import warnings
+from collections import OrderedDict
+from contextlib import contextmanager
 
 import dask
 import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
-from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
-                               format_bytes)
+from distributed.diagnostics.plugin import SchedulerPlugin
+from distributed.utils import (
+    format_bytes, get_ip_interface, parse_bytes, tmpfile)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
@@ -26,6 +31,54 @@ Please specify job size with the following keywords:
 -  memory: total memory per job, across all processes
 -  processes: number of processes to launch, splitting the quantities above
 """.strip()
+
+
+def _job_id_from_worker_name(name):
+    ''' utility to parse the job ID from the worker name
+
+    template: 'prefix--jobid--suffix'
+    '''
+    _, job_id, _ = name.split('--')
+    return job_id
+
+
+class JobQueuePlugin(SchedulerPlugin):
+    def __init__(self):
+        self.pending_jobs = OrderedDict()
+        self.running_jobs = OrderedDict()
+        self.finished_jobs = OrderedDict()
+        self.all_workers = {}
+
+    def add_worker(self, scheduler, worker=None, name=None, **kwargs):
+        ''' Run when a new worker enters the cluster'''
+        logger.debug("adding worker %s" % worker)
+        w = scheduler.workers[worker]
+        job_id = _job_id_from_worker_name(w.name)
+        logger.debug("job id for new worker: %s" % job_id)
+        self.all_workers[worker] = (w.name, job_id)
+
+        # if this is the first worker for this job, move job to running
+        if job_id not in self.running_jobs:
+            logger.debug("this is a new job")
+            self.running_jobs[job_id] = self.pending_jobs.pop(job_id)
+
+        # add worker to dict of workers in this job
+        self.running_jobs[job_id][w.name] = w
+
+    def remove_worker(self, scheduler=None, worker=None, **kwargs):
+        ''' Run when a worker leaves the cluster'''
+        logger.debug("removing worker %s" % worker)
+        name, job_id = self.all_workers[worker]
+        logger.debug("removing worker name (%s) and"
+                     "job_id (%s)" % (name, job_id))
+
+        # remove worker from this job
+        del self.running_jobs[job_id][name]
+
+        # once there are no more workers, move this job to finished_jobs
+        if not self.running_jobs[job_id]:
+            logger.debug("that was the last worker for job %s" % job_id)
+            self.finished_jobs[job_id] = self.running_jobs.pop(job_id)
 
 
 @docstrings.get_sectionsf('JobQueueCluster')
@@ -87,6 +140,8 @@ class JobQueueCluster(Cluster):
     submit_command = None
     cancel_command = None
     scheduler_name = ''
+    _adaptive_options = {
+        'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
 
     def __init__(self,
                  name=None,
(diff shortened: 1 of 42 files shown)


mrocklin:
Some interesting failures here

jhamman:
> Some interesting failures here

I agree. I plan to dig into them further tomorrow. I think the switch to using `scale()` has triggered some additional race conditions that we'll want to chase down.

jhamman:
@lesteve - may I ask for a review here? 

If anyone has any ideas on how to move forward here, I would certainly appreciate it. It seems that my change to `scale()` instead of `stop_workers()` in our tests has exposed a bug somewhere.

dask_jobqueue/core.py
```diff
@@ -356,29 +354,29 @@ def stop_workers(self, workers):
...
-        logger.debug("Found %d active/pending workers." % active_and_pending)
+        logger.debug("Found %d active/pending workers.", active_and_pending)
         self.start_workers(n - active_and_pending)
 
     def scale_down(self, workers):
         ''' Close the workers with the given addresses '''
-        logger.debug("Scaling down. Workers: %s" % workers)
+        logger.debug("Scaling down. Workers: %s", workers)
         worker_states = []
         for w in workers:
```
jhamman: @mrocklin - I've been working on debugging the failures we're seeing and I may have stumbled onto a real bug. I'll lay out what I can understand.

- when we call `scale(0)`, we correctly end up in the scale down method
- however, `workers` is of type (`_asyncio.Future`) and is not a list of workers
- `workers.result()` yields an empty list. 

From here, I've tried a bunch of things including:

- removing the `loop.add_callback` calls in `scale`
- tweaking the remove/close_workers in `retire_workers`

But alas, I'm pretty lost. Any pointers would be appreciated.

mrocklin: Yup, lost with reason.  That looks like a pretty clear bug.  I'm not sure how that ever would have worked.

For context.  `Scheduler.retire_workers` is a coroutine

```python
    @gen.coroutine
    def retire_workers(self, comm=None, workers=None, remove=True,
                       close_workers=False, **kwargs):
```

This means that it is meant to be called within *other* coroutines, with a yield statment

```python
@gen.coroutine
def f():
    yield scheduler.retire_workers(...)
```

This is true of any internal dask scheduler method that communicates with other things or does anything that might take any non-trivial amount of time.

In this way, coroutines are a bit viral, which in our case is bad, because we don't want `scale` to be a coroutine, because users use it, and users get confused by coroutines.

Another approach is to call coroutines with `add_callback` which says "run this whenever you have a moment, but don't lets deal with it now, because it's a bit messy".

So probably the thing to do here is in `distributed/deploy/cluster.py::Cluster.scale` we want to call `self.scheduler.workers_to_close` which is thankfully just a normal method and then we want to add `self.scheduler.retire_workers` as a callback with those workers to be run in just a moment, then we want to call `scale_down` as we do currently (which is also as a callback, just in case the cluster object implements it as a coroutine as well.

I can do this, but I'd also be very happy to guide someone else through it in the interests of spreading some of this knowledge around.

mrocklin: To be clear the problem here is that we were calling it as a normal function

```python
def scale(...):
    ...

    to_close = self.scheduler.retire_workers(...)
```

*without a yield* within a function that *was not a coroutine*.  In this case it returns an opaque asyncio/torando Future object.  This is never a good thing to see and in our case almost always signifies a bug.

jhamman: @mrocklin - I'm happy to push this forward if you can provide some hand holding. https://github.com/jhamman/distributed/commit/ad0256006bc7d5018277da0d6c686224ffaaa8fe is a first attempt based on my reading of what is above. This does cause some tests to fail but I think it is worth sharing in its current form for initial feedback.

dask_jobqueue/core.py
```diff
@@ -363,33 +361,84 @@ def stop_workers(self, workers):
...
+            else:
+                n_to_close = active_and_pending - n
+
+                if n_to_close < self._count_pending_workers():
+                    # We only need to kill some pending jobs
+                    to_kill = int(n_to_close / self.worker_processes)
+                    jobs = list(self.pending_jobs.keys())[to_kill:]
+                    self.stop_jobs(jobs)
+                else:
+                    # we need to retire some workers (and maybe pending jobs)
```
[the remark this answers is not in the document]
jhamman: Yes, this comment is a bit misleading. What I meant to say was that all pending jobs will be killed too.

dask_jobqueue/core.py
```diff
@@ -363,33 +361,84 @@ def stop_workers(self, workers):
...
+        -------
+        >>> cluster.scale(10)  # scale cluster to ten workers
+        See Also
+        --------
+        Cluster.scale_up
+        Cluster.scale_down
+        """
+        with log_errors():
+            active_and_pending = self._count_active_and_pending_workers()
+            if n >= active_and_pending:
```
jhamman: these two lines are why I am currently overloading the scale method. We need to know if we want to scale up or scale down and that depends on the number of pending workers. The current `Cluster.scale()` method does not have a way to evaluate pending workers.

guillaumeeb: This part is currently done in `scale_up`, maybe not correctly, but what prevents us from doing it there?

Something like this:
````python
    def scale_up(self, n, **kwargs):
            """ Brings total worker count up to ``n`` """
            active_and_pending = self._count_active_and_pending_workers()
            if n >= active_and_pending:
                 self.start_workers(n - self._count_active_and_pending_workers())
            else:
                n_to_close = active_and_pending - n
                if n_to_close < self._count_pending_workers():
                    # We only need to kill some pending jobs
                    to_kill = int(n_to_close / self.worker_processes)
                    jobs = list(self.pending_jobs.keys())[to_kill:]
                    self.stop_jobs(jobs)
                else:
                    raise something, we should never be there in scale_up.
````

guillaumeeb:
I will give a go to this one probably next week as discussed in #130. Thanks @jhamman.

guillaumeeb:
Gave a try today to this PR. Works fine! Actually, I don't have any problem on master branch when using scale and adaptive, and this one works good too (no noticable change from user perspective).

Just one case in which I se some faulty behaviour (but seeing it too on master) : 
````python
# Create some cluster
cluster = PBSCluster(processes=2, cores=4, memory="20GB")
# Scale workers up
cluster.scale(8)
# Correct number of jobs and worker showing up
# Do things, then scale down
cluster.scale(4)
# Only one job and two workers left in running jobs :
cluster.running_jobs
 Out:   OrderedDict([('6732209',
              {'dask-worker--6732209---0': <Worker 'tcp://10.135.36.116:40331', memory: 0, processing: 0>,
               'dask-worker--6732209---1': <Worker 'tcp://10.135.36.116:35941', memory: 0, processing: 0>})])
````

One way or another, it looks like something's wrong with
````python
_adaptive_options = {
        'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
````

guillaumeeb:
Could we enable some debug logs in the test here, to have more insights of why tests are failing?

jhamman:
@guillaumeeb - thanks for taking a look at this. My attentions have obviously been elsewhere lately so I appreciate the persistence on your part.

guillaumeeb:
So I propose this new way without scale. And if we manage to correct upstream issue https://github.com/dask/distributed/issues/2257, this should be a good step.

Not sure how to handle the test `test_basic_scale_edge_case` though.

guillaumeeb:
So as discussed previously, the `scale(0)` operation here https://github.com/dask/dask-jobqueue/pull/97/files#diff-660047dbe8333ae717d8f94cc4162529R127 is basically a noop as cluster state is evaluated before `scale(2)` has done any modification to the current state. This is why the associated test fails.

In order to advance here, I propose to disable the test here for the time being, and to merge this PR as it provides some nice enhancement to dask-jobqueue.

I propose then to implement https://github.com/dask/distributed/issues/2257 in another PR directly in dask-jobqueue, meaning overload scale or even upstream Cluster object. This should hopefully fix the failing test here. As discussed in #130, this should provide some interesting lessons for https://github.com/dask/distributed/issues/2235.

jhamman:
Thanks @guillaumeeb. I like the plan you've laid out. Happy to see this merged as is.

Issue #90, opened 2018-07-11
Allow global access to dashboard by default


mrocklin:
When using interfaces like infiniband it's often the case that one can't get the dashboard to show up.  This is because the bokeh server is expecting to be called by its IP address on the infiniband network, not the normal network.

One solution to this  would be to provide `services={('bokeh', ('', 8787)): BokehScheduler}` to LocalCluster.

lesteve:
Interesting, I do seem to remember to have bumped into this problem ...

lesteve:
Hmmm I think I got something like you suggested working, but the `_repr_html_` is not showing the right port for some reason ...

lesteve:
Just to put a bit more context, here is my diff (I hard-code the BokehScheduler port to 12345):
```
diff --git a/dask_jobqueue/core.py b/dask_jobqueue/core.py
index ab4b602..9bc467e 100644
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -9,6 +9,7 @@ import warnings
 import dask
 import docrep
 from distributed import LocalCluster
+from distributed.bokeh.scheduler import BokehScheduler
 from distributed.deploy import Cluster
 from distributed.utils import (get_ip_interface, ignoring, parse_bytes, tmpfile,
                                format_bytes)
@@ -153,7 +154,8 @@ class JobQueueCluster(Cluster):
         else:
             host = socket.gethostname()
 
-        self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+        self.local_cluster = LocalCluster(n_workers=0, ip=host,
+                                          services={('bokeh', ('', 12345)): BokehScheduler}, *kwargs)
 
         # Keep information on process, cores, and memory, for use in subclasses
         self.worker_memory = parse_bytes(memory)
```

I run this ssh proxy (redirecting the 12345 port to 9898 locally):
```
ssh -N lesteve@sequoia -L 9898:sequoia:12345 # dashboard
```

and then I can access the dashboard at http://localhost:9898/status.

The widget for `client.cluster` shows the dashboard at 8787 though:
```
Dashboard: http://10.149.255.254:8787/status
```

I made sure I had the latest release of dask (0.18.1) and distributed (1.22).

mrocklin:
Indeed.  It looks like this is broken in dask/distributed.  We just do a hard overwrite of the bokeh key: 

[...]

PR #107, opened 2018-07-27
forward 'ip' to LocalCluster

twoertwein:
JobQueueCluster forwards any additional keyword arguments to LocalCluster. However, it fails to forward 'ip' to the LocalCluster as it is explicitly set. This commit allows the user to forward 'ip' to the LocalCluster.

Context:
My cluster head has two network interfaces (one facing only the workers and one facing the internet). I want the scheduler to only listen on the worker facing interface. The 'interface' argument would be well suited for this task but JobQueueCluster also forwards this interface to the workers. Unfortunately, not all my cluster nodes have the same interface names. Therefore, I have to use 'ip' to force the scheduler to listen on the worker facing interface (without forcing the workers itself to use a (potentially non-existing) interface).
Personally, I would prefer to not force the interface of the workers (in that case I could use 'interface' to set the interface of only the scheduler) but I assume you have a reason for doing so :)

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -202,13 +202,16 @@ class JobQueueCluster(Cluster):
         #This attribute should be overriden
         self.job_header = None
 
+        # Find the IP address and (if requested) set interface for workers
+        localhost_kwargs = {}
         if interface:
-            host = get_ip_interface(interface)
+            localhost_kwargs['ip'] = get_ip_interface(interface)
             extra += ' --interface  %s ' % interface
         else:
-            host = socket.gethostname()
+            localhost_kwargs['ip'] = socket.gethostname()
+        localhost_kwargs.update(kwargs)
 
-        self.local_cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+        self.local_cluster = LocalCluster(n_workers=0, **localhost_kwargs)
 
         # Keep information on process, threads and memory, for use in
         # subclasses
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -41,4 +41,3 @@ def test_forward_ip():
     with PBSCluster(walltime='00:02:00', processes=4, cores=8, memory='28GB',
                     name='dask-worker') as cluster:
         assert cluster.local_cluster.scheduler.ip == default_ip
-
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -1,4 +1,5 @@
 import pytest
+import socket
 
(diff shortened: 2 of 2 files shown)


lesteve:
I am not an expert in this kind of thing but this seems like a problem related to https://github.com/dask/dask-drmaa/issues/75. It would be great if you could add a test similar to the one in https://github.com/dask/dask-drmaa/pull/77.

> Unfortunately, not all my cluster nodes have the same interface names.

Interesting, can I ask whether there is a good reason for this?

> Personally, I would prefer to not force the interface of the workers (in that case I could use 'interface' to set the interface of only the scheduler) but I assume you have a reason for doing so :)

Not sure ... maybe @mrocklin has some inputs on this.

twoertwein:
> I am not an expert in this kind of thing but this seems like a problem related to dask/dask-drmaa#75. It would be great if you could add a test similar to the one in dask/dask-drmaa#77.

Yes, that looks similar. I will look trough how you are doing unit tests and then add one to verify the IP address.

> Interesting, can I ask whether there is a good reason for this?

There is no (good) reason for that. I wish the interfaces across all nodes had the same name. It is a cluster from a university research lab.

> Those changes seem fine to me, thanks. I was searching a way to do that with less 'if' but I'm still not pythonic enough to find one.

This solution would have less if's and it doesn't change the input arguments:

```py
# Find the IP address and (if requested) set interface for workers
locahost_kwargs = {}
if interface:
    locahost_kwargs['ip'] = get_ip_interface(interface)
    extra += ' --interface  %s ' % interface
else:
    locahost_kwargs['ip'] = socket.gethostname()

locahost_kwargs.update(kwargs)
self.local_cluster = LocalCluster(n_workers=0, **locahost_kwargs)
```

twoertwein:
@lesteve about the test case, there are some differences compared to dask/dask-drmaa#77
1) JobQueueCluster should not directly be instantiated. I could add a SLURM test case, but that seems to be the wrong place for it;
2) The requested IP address in dask/dask-drmaa#77 is the same one as JobQueueCluster would use by default, i.e., it wouldn't test anything. I could request an IP address which 'probably' does not coincide with the default IP address by requesting the (probably not-existing) IP address `127.0.0.2` (surprisingly, it does work to insatiate a SLURMCluster with this IP address, even though that machine doesn't have an interface with that ip).

I don't think that using a SLURMCluster (wrong component) with a probably not-existing IP address is a good way to test this PR. Do you have any suggestions?

I will change the PR to reflect @guillaumeeb suggestion - it looks much nicer :)

guillaumeeb:
I'm adding basic tests here, to merge if everything is OK so that we can move on to solve #90.

guillaumeeb:
Getting this in quiclky to propose another PR that solves #90 and allows #111.

Thanks @twoertwein.

PR #109, opened 2018-07-27
Flake8 and source code cleanup

### LINKED ISSUE #38: flake8 igore configuration
I was wondering whether there is a good motivation for the flake8 ignore settings and whether people would be open to use the flake8 defaults instead, or maybe a smaller set of ignores instead.

I am aware this is a bit of a bikeshedding issue, but I think that this is better to discuss this when the project is young than later down the line.

I ran flake8 with the default settings and there are eight flake8 violations which are super easy to fix (if you are curious the winner is `E265 block comment should start with '# '` which to me is not very controversial).

I had a quick look and from what I can tell, the flake8 ignore settings of some projects in the dask organization projects are copied from the dask/dask one. I found this was introduced in https://github.com/dask/dask/pull/1464 and it does not look like there is a strong motivated reason behind it. Some projects have drifted a bit off like dask-ml and dask-kubernetes.

### LINKED ISSUE #106: Add flake8 to tests or black to CI
Currently it looks like PRs can pass tests without passing flake8 tests.  Do we want to start enforcing style?

jhamman:
closes #38, closes #106 

This uses the flake8 config from distributed and adds some boilerplate future imports.

### DIFF
--- a/dask_jobqueue/tests/__init__.py
+++ b/dask_jobqueue/tests/__init__.py
@@ -1,2 +1,4 @@
+# flake8: noqa
+from __future__ import absolute_import, division, print_function
 
 QUEUE_WAIT = 60  # seconds
--- a/dask_jobqueue/moab.py
+++ b/dask_jobqueue/moab.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, division, print_function
+
 from .core import docstrings
 from .pbs import PBSCluster
 
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, division, print_function
+
 import pytest
 
 from dask_jobqueue import JobQueueCluster
--- a/dask_jobqueue/tests/test_pbs.py
+++ b/dask_jobqueue/tests/test_pbs.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, division, print_function
+
 import sys
 from time import sleep, time
 
--- a/dask_jobqueue/tests/test_sge.py
+++ b/dask_jobqueue/tests/test_sge.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, division, print_function
+
 from time import sleep, time
 
 import pytest
--- a/dask_jobqueue/tests/test_slurm.py
+++ b/dask_jobqueue/tests/test_slurm.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, division, print_function
+
 import sys
 from time import sleep, time
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ license-file = LICENSE.txt
 # https://flake8.readthedocs.io/en/latest/user/error-codes.html
 
 # Note: there cannot be spaces after comma's here
-exclude = __init__.py
+exclude = __init__.py,distributed/_concurrent_futures_thread.py
 ignore =
     # Extra space in brackets
     E20,
@@ -21,8 +21,29 @@ ignore =
     E721,
     # Assigning lambda expression
     E731,
+    # continuation line under-indented for hanging indent
+    E121,
+    # continuation line over-indented for hanging indent
+    E126,
+    # continuation line over-indented for visual indent
+    E127,
+    # E128 continuation line under-indented for visual indent
+    E128,
+    # multiple statements on one line (semicolon)
+    E702,
+    # line break before binary operator
+    W503,
+    # visually indented line with same indent as next logical line
+    E129,
+    # unexpected indentation
+    E116,
+    # redefinition of unused 'loop' from line 10
+    F811,
+    # local variable is assigned to but never used
+    F841,
     # Ambiguous variable names
     E741
+
 max-line-length = 120
 
 [versioneer]
@@ -40,3 +61,6 @@ test = pytest
 default_section=THIRDPARTY
 known_first_party=dask_jobqueue
 multi_line_output=4
+
+[bdist_wheel]
+universal = 1


lesteve:
Personally as I tried to convey in #38, I would be in favour of stripping down the flake8 ignore we already have in setup.cfg rather than adding more. Here are the default ignore option from `flake8 --help`:

```
 --ignore=errors       Comma-separated list of errors and warnings to ignore
                        (or skip). For example, ``--ignore=E4,E51,W234``.
                        (Default: E121,E123,E126,E226,E24,E704,W503,W504)
 ```

With the default flake8 warnings: I get only two flake8 violations which are quite easy to fix and do not seem very controversial to me:

```
❯ flake8 dask_jobqueue --show-source --ignore='E121,E123,E126,E226,E24,E704,W503,W504'
dask_jobqueue/sge.py:80:49: E231 missing whitespace after ','
                  'resource_spec': resource_spec,}
                                                ^
dask_jobqueue/core.py:202:9: E265 block comment should start with '# '
        #This attribute should be overriden
```

For completeness sake is the diff I would be in favour of (+ fixing the 2 flake8 violations):
```diff
diff --git a/setup.cfg b/setup.cfg
index a42170f..d3913b0 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,22 +7,6 @@ license-file = LICENSE.txt
 # https://flake8.readthedocs.io/en/latest/user/error-codes.html
 
 # Note: there cannot be spaces after comma's here
-exclude = __init__.py
-ignore =
-    # Extra space in brackets
-    E20,
-    # Multiple spaces around ","
-    E231,E241,
-    # Comments
-    E26,
-    # Import formatting
-    E4,
-    # Comparing types instead of isinstance
-    E721,
-    # Assigning lambda expression
-    E731,
-    # Ambiguous variable names
-    E741
 max-line-length = 120
 
 [versioneer]
```

guillaumeeb:
If there is no strong reason to ignore all this, and no standard ignores defined for all the dask modules, then 👍 to @lesteve proposal.

mrocklin:
Fine by me

jhamman:
@lesteve - sounds great. This is my preference as well. I will update accordingly.

lesteve:
I am fine with keeping the origin max-line-length = 120. At the moment you have some flake8 failures because some lines are over 79 characters (flake8 violations don't get the Travis build to fail, but #110 should fix that).

guillaumeeb:
Oh yes please, max-line-length to 120 characters. It's not like we've got low-res small screens anymore.

mrocklin:
Checking in.  What is the status on this?

jhamman:
provided the tests pass here, I think this is now ready to merge.

docs/configurations.rst
```diff
@@ -28,18 +30,30 @@ Cheyenne
...
+       cores: 1                    # Total number of cores per job
+       memory: '25 GB'             # Total amount of memory per job
+       processes: 1                # Number of Python processes per job
 
        interface: ib0
-       local-directory: $TMPDIR
 
-       queue: regular
-       project: null  # TODO, change me
+       project: PXYZ123
```
mrocklin: I anticipate that people who don't pay attention won't change this.  I slightly prefer leaving this as `null` so that they get an appropriate message about it.

docs/configurations.rst
```diff
@@ -28,18 +30,30 @@ Cheyenne
...
-       cores: 36
-       memory: 108GB
-       processes: 4
+       name: dask-worker
+       cores: 36                   # Total number of cores per job
+       memory: '109 GB'            # Total amount of memory per job
+       processes: 9                # Number of Python processes per job
+       interface: ib0              # Network interface to use like eth0 or ib0
+
+       queue: premium
```
mrocklin: Is this definitely the right queue for everyone?  I'm inclined to leave these things out if they're not universally useful.

guillaumeeb: Sounds good to me to have an as much as possible real example, so all except the project detail.

jhamman:
I would actually prefer sticking with 80 chars but 1) I'll get over it since there seems to be consensus among the other devs and 2) I've rolled back all line breaks that were between 80 and 120 chars in bad50ac. This may just be for demonstration or we can just go with it.

mrocklin:
FWIW my operational line length in most personal projects tends to be "80 characters except when it's inconvenient".

mrocklin:
No comment here though.  I'm happy to stay out of this conversation :)

guillaumeeb:
Apparently I am the only one that prefers long lines, @lesteve was just fine keeping `max-line-length = 120` for some lines, which looks like @mrocklin comment. 

It is really a minor concern, so lets just keep it your way @jhamman!

For the test failures here, I had some problems with this file and the last line when pushing to #107.

jhamman:
This seems ready to merge. Since there were a few minor disagreements along the way on this one, I'll let @guillaumeeb or @lesteve have a final look.

dask_jobqueue/core.py
```diff
@@ -313,14 +306,13 @@ def scheduler(self):
     def _calls(self, cmds, **kwargs):
         """ Call a command using subprocess.communicate
 
-        This centralzies calls out to the command line, providing consistent
-        outputs, logging, and an opportunity to go asynchronous in the future
+        This centralzies calls out to the command line, providing consistent outputs, logging, and an opportunity
```
guillaumeeb: Old typo in centralizes

dask_jobqueue/pbs.py
```diff
@@ -60,7 +57,8 @@ class PBSCluster(JobQueueCluster):
     cancel_command = 'qdel'
     scheduler_name = 'pbs'
 
-    def __init__(self, queue=None, project=None, resource_spec=None, walltime=None, job_extra=None, **kwargs):
+    def __init__(self, queue=None, project=None, resource_spec=None,
+                 walltime=None, job_extra=None, **kwargs):
```
guillaumeeb: You still have some line breaks here and there, but this is not important.

dask_jobqueue/tests/__init__.py
```diff
@@ -1,2 +1,4 @@
+# flake8: noqa
```
guillaumeeb: Why is there a noqa here? Do we need this import?

PR #110, opened 2018-07-31
Get flake8 errors to fail the build on Travis

lesteve:
This adds `set -e` to all bash scripts. My understanding is that using `set -e` in bash scripts is always a good idea: if one of the command fails you want to stop the script and exit with a non-zero exit code.

As mentioned in #106, flake8 is run in the JOBQUEUE=none build but does not fail the build.
Look at this [Travis build](https://travis-ci.org/dask/dask-jobqueue/jobs/406081574#L915) for an example where the flake8 has some errors but the Travis build stays green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,20 @@ env:
     - DOCKER_COMPOSE_VERSION=1.6.0
 
 before_install:
+  - set -e
   - pwd
   # Init jobqueue environment: load init and test methods
   - source ci/${JOBQUEUE}.sh
   - jobqueue_before_install
+  - set +e
 install:
+  - set -e
   - jobqueue_install
+  - set +e
 script:
+  - set -e
   - jobqueue_script
+  - set +e
 after_script:
   - jobqueue_after_script
 
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -e
-
 function jobqueue_before_install {
   # Install miniconda
   ./ci/conda_setup.sh
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 function jobqueue_before_install {
   # Install miniconda
   ./ci/conda_setup.sh
--- a/ci/pbs.sh
+++ b/ci/pbs.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 function jobqueue_before_install {
     docker version
     docker-compose version
--- a/ci/pbs.sh
+++ b/ci/pbs.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -e
-
 function jobqueue_before_install {
     docker version
     docker-compose version
--- a/ci/pbs/build.sh
+++ b/ci/pbs/build.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e
+
 cd /src/pbspro
 ./autogen.sh
 ./configure -prefix=/opt/pbs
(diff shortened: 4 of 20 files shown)


lesteve:
The commits are a bit convoluted on purpose to double-check that this PR is doing what it says on the tin. Please select squash+merge if you are merging this PR! 

If I got this right this is what should happen:
- Travis status for the first commit should be green although I added a flake8 violation. This should reproduce the problem that indeed there is a problem with our CI setup.
- Travis status for the second commit should be red: flake8 violation should be detected
- Travis status for the third commit should be green: I reverted the flake8 violation

lesteve:
> Please select squash+merge if you are merging this PR!

To be perfectly honest, I would be in favour of only enabling squash+merge in the web UI!

lesteve:
At the moment, I can't get the Travis build to fail for some reason (second commit). Need to debug this a bit more ...

lesteve:
OK I think this seems to be failing as expected now. See this [Travis build](https://travis-ci.org/dask/dask-jobqueue/jobs/410327547#L937).

The problem was me trying to outsmart Travis by pushing multiple commits and stopping/restarting the Travis builds manually (related to the auto-cancel functionality being probably turned on for dask-jobqueue) ...

guillaumeeb:
Looks like there are side effects on sge builds 🙁.

I'm relieved that it did nothing to pbs and slurm ones though.

lesteve:
> In a perfect scripting world it would probably be true, but I'm a little afraid of what could happen here: failing docker images or so one.

Do you have an example where using `set -e` could be a problem? In my experience it is better to fail early and this is what `set -e` is doing. The only time I found `set -e` slightly annoying was when I had to do a grep in a script that potentially would not match anything (and so would exit with an exit code equal to 1) but there are work-arounds for this.

> Looks like there are side effects on sge builds slightly_frowning_face.

Hmmm good point I'll have a look.

guillaumeeb:
> Do you have an example where using set -e could be a problem?

Not right here, I can just tell that if I had used it in all the scripts I've written, a good part of them would probably fail 🙂, and this is probably not a good thing!

guillaumeeb:
Hm, it seems to trigger random failing in PBS too. 

From what I can see, there is not always files in /var/spool/pbs/spool/. Which is normal because PBS must remove or move those if everything is OK I think. It is probably a matter of how fast previous command are ran and if PBS had the time to do its internal things. Those commands were mainly to have some more informations if tests are failing though.

If we want to be OK with the `set -e` command, we probably need to use a `if directory not empty` before that, or remove the line. One example of problem we may encounter here.

lesteve:
Although Travis is green, I don't think this one should be merged quite yet.

Basically using `set -e` has unintended consequences in Travis. One that I bumped into with the random PBS failure: after_script failure should not mark the build as fail but it does when you use `set -e`. I was able to reproduce this unexpected behaviour on a toy project. This is probably loosely related to https://github.com/travis-ci/travis-ci/issues/3586#issuecomment-274585174.

lesteve:
I went for a less intrusive change, basically using `set -e` and `set +e` around the 
before_install, install and script steps directly in the .travis.yml.

About the impact of a "global" `set -e` in Travis I found similar comments:

https://github.com/travis-ci/docs-travis-ci-com/issues/1672#issuecomment-363230601
> It would be good to mention that using set -e inside .travis.yml itself may break things +1 Or to remind people to do set +e as soon as possible after set -e

https://github.com/travis-ci/docs-travis-ci-com/issues/1672#issuecomment-363234263
> warning Don't use set -e in .travis.yml, otherwise the build may stop unexpectedly because of benign non-zero status in internal travis scripts (see issue #1672).

lesteve:
I think we are good. Adding `set -e/set +e` in .travis.yml feels like the simplest solution.

guillaumeeb:
Merging this then, the soonner we have flake, the better!

PR #113, opened 2018-08-01
Add a cluster __repr__

### LINKED ISSUE #91: Cluster needs a repr
We look great in a Jupyter Notebook, but not so great in IPython.  It would be useful to include the number of running/pending jobs and maybe the total amount of CPU/Memory if there is space?

Found with @josephhardinee

guillaumeeb:
Closes #91 

As mentionned in #91, I started from the format found in [LocalCluster](https://github.com/dask/distributed/blob/master/distributed/deploy/local.py#L143), in order to have some consistency:
````python
In [9]: cluster
Out[9]: PBSCluster('tcp://10.120.42.23:32961', workers=6, cores=12/16, memory=60.00 GB/80.00 GB, jobs=3/4, finished_jobs=0)

In [12]: cluster
Out[12]: PBSCluster('tcp://10.120.42.23:32961', workers=4, cores=8/8, memory=40.00 GB/40.00 GB, jobs=2/2, finished_jobs=1)
````

Please give me some return on the information displayed and the way it is printed.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -245,6 +245,21 @@ class JobQueueCluster(Cluster):
         if extra is not None:
             self._command_template += extra
 
+    def __repr__(self):
+        running_workers = sum(len(value) for value in self.running_jobs.values())
+        running_cores = running_workers * self.worker_threads
+        total_jobs = len(self.pending_jobs) + len(self.running_jobs)
+        total_cores = total_jobs * self.worker_cores
+        running_memory = running_workers * self.worker_memory / self.worker_processes
+        total_memory = total_jobs * self.worker_memory
+
+        return (self.__class__.__name__ +
+                '(%r, workers=%d, cores=%d/%d, memory=%s/%s, jobs=%d/%d, finished_jobs=%d)' %
+            (self.scheduler.address, running_workers, running_cores, total_cores,
+             format_bytes(running_memory), format_bytes(total_memory),
+             len(self.running_jobs), total_jobs, len(self.finished_jobs))
+            )
+
     @property
     def pending_jobs(self):
         """ Jobs pending in the queue """

### CODE REVIEW

**Comment by mrocklin** [comment]
This looks good to me.  I think that we should think about how best to use horiztonal space, and how to prioritize information.  For example, I mostly care about number of cores and amount of memory.  I care less about the scheduler address and finished jobs.  I also think that there is probably some redundancy with workers/cores/jobs that we might be able to remove.  Ideally things would fit in one 80-character line.

Going in the other extreme, maybe the following:

    PBSCluster(cores=12, memory=60 GB, jobs=3/4)

Here I've dropped scheduler address, workers, and finished jobs.  I've only included the ratio for jobs, letting the user infer the rest.  I've removed `.00` from the memory string (this should probably happen upstream)

**Comment by mrocklin** [comment]
To be clear, I'm intentionally going to one extreme with the example above.  I like it but I suspect that the right answer is somewhere in the middle.

**Comment by mrocklin** [comment]
Maybe we only include ratios when the cluster allocation is not complete?

**Comment by guillaumeeb** [comment]
I was also giving a lot of information just to be able to chose what to keep, but that's true that I like to have all these information :slightly_smiling_face:.

 Maybe we could go with
```
PBSCluster(cores=12, memory=60 GB, workers=6/8, jobs=3/4)
```
I feel that worker number is important, as this is what we use in the `scale()` function.

**Comment by mrocklin** [comment]
Ah, good point

**Comment by guillaumeeb** [comment]
So with the last commit, we should have something simplified:

```
PBSCluster(cores=0, memory=0 B, workers=0/0, jobs=0/0)
```

**Comment by mrocklin** [comment]
In principle that seems fine to me.

**Comment by guillaumeeb** [comment]
Real world example are working good:

```
PBSCluster(cores=12, memory=30.00 GB, workers=6/18, jobs=3/9)
PBSCluster(cores=36, memory=90.00 GB, workers=18/18, jobs=9/9)
```

mrocklin:
This looks good to me.  My apologies for the delay in response. 

I recommend though that we do have a small test for this.  It shouldn't necessarily be too strict, but it would be good to make sure that these lines are covered.  I might test the following:

1. The name of the cluster class is present
2.  The number of cores is present
3.  The total amount of memory is present

I probably wouldn't test much else because most of this is subjective.

guillaumeeb:
I added a very basic test, with no worker launched. Do you want to add some tests with a launched cluster? Inside PBS or Slurm specific tests for example?

PR #114, opened 2018-08-01
Adding stop_jobs and close methods to JobQueueCluster

### LINKED ISSUE #83: Stopping worker and stopping whole jobqueue cluster
The question is what is the right way to do this?

Currently, when I want to stop my JobQueue cluster to restart a fresh one (with new parameters for instance), I use
````
cluster.scheduler.close()
````
For stopping workers, as the scale method is not currently working, I use something like 
````
cluster.stop_wokers(cluster.jobs)
````
(I don't remember the exact syntax, and not eager to look for it).

Are these the right things to do?

Should we expose some more explicit methods, for example:
````
cluster.stop()
````
and
````
cluster.stop_all_workers()
````

### PR DESCRIPTION
**Pull request by guillaumeeb**
Closes #83.

I'm wondering if we should use `self.scale(0)` into `stop_all_jobs()` function.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -342,7 +342,7 @@ class JobQueueCluster(Cluster):
         logger.debug("Stopping workers: %s" % workers)
         if not workers:
             return
-        jobs = self._stop_pending_jobs()  # stop pending jobs too
+        jobs = self._del_pending_jobs()  # stop pending jobs too
         for w in workers:
             if isinstance(w, dict):
                 jobs.append(_job_id_from_worker_name(w['name']))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: dask_jobqueue/core.py [hunk 1]
```diff
@@ -342,7 +342,7 @@ def stop_workers(self, workers):
         logger.debug("Stopping workers: %s" % workers)
         if not workers:
             return
-        jobs = self._stop_pending_jobs()  # stop pending jobs too
+        jobs = self._del_pending_jobs()  # stop pending jobs too
```
mrocklin: I'm actually a bit confused by this line of code.  We always stop/delete pending jobs whenever we call `stop_workers`?  What if someone wants to stop a particular job even while other jobs are coming online?  

I know that this isn't related to this PR.  My apologies for the off-topic conversation.

guillaumeeb: I was a really surprised at first too. But actually, if I understand correctly, `scale_down` will only be called if we want less workers that we currently have, so less worker that what is already online. This means that all pending jobs are useless and can be safely cancelled. 

On the other hand, we could call scale_up, but find that we already have enough pending jobs to fulfil the requested number. In this case we do nothing. Maybe we should even stop some of pending jobs if the new requested scale is less than running + pending.

I don't know if this is the correct approach regarding dask behaviour upstream. Maybe we should always consider the potential amount of worker as online ones... That doesn't sound really good though.

Any thought?

guillaumeeb: And for argumentation about the name change, it is because _stop_pending_jobs is not actually stopping anything, so I wanted to make that clearer. Any suggestion on the name of the method welcome.

guillaumeeb:
> It looks to me like it needs a decision on whether or not we want to include it.

Yes exactly. Ideally, if this is not something you find useful, we should write some documentation on how to properly stop and restart a cluster in a notebook.

guillaumeeb:
cc @jhamman and @lesteve, any advice on this PR? I'm fine with it not being merged if you don't find it useful.

guillaumeeb:
Merging this, I believe it would be of use for new users, and it is good not to depend on scale.

Issue #111, opened 2018-07-31
Include nbserverproxy or describe how to use it in the docs


guillaumeeb:
See https://github.com/dask/dask-jobqueue/issues/80#issuecomment-400797727.

To prevent from doing a lot of ssh tunneling, using [`nbserverproxy`]( https://github.com/jupyterhub/nbserverproxy ) could be helpful!

guillaumeeb:
Cross referencing the work of @jhamman on https://github.com/pangeo-data/pangeo/pull/304 here, this already work in progress!

jakirkham:
Probably worth linking [this comment]( https://github.com/dask/dask-jobqueue/issues/80#issuecomment-400947329 ) as well since that has a little more details about how to set it up. Not that there is much to setup at all.

guillaumeeb:
@jakirkham, I tried nbserverproxy with no luck, as mentionned in pangeo thread. Could you look at it to see if you're doing things differently? I'll give it another go when I have some time.

raybellwaves:
@guillaumeeb I was the same. I ended up opening an issue over there https://github.com/jupyterhub/nbserverproxy/issues/45 but closed it once I had the `ssh` solution working.

guillaumeeb:
Okay, so I've found the problem with my setup : Dask dashboard was bind to a given network interface, but apparently nbserverproxy was not able to proxy to this one, it was probably trying another interface, don't know why though.

So in order to make nbserverproxy work, I add to use the trick described in #90: bind the dashboard to all interfaces.

guillaumeeb:
@raybellwaves, could you give another go to nbserverproxy solution, now that #119 is in master?

raybellwaves:
I haven't actually used jupyterhub before so you may want to ask someone with more experience than I. I still gave it a go though and my build notes from scratch are below. I made it as far as to 'sign up with github button' but it wasn't really responding after that

```
$ conda create -n djq_nbs python=3.6
$ conda activate dqj_nbs
$ pip install git+https://github.com/dask/dask-jobqueue.git
$ conda install -c conda-forge psutil jupyterhub nbserverproxy bokeh
$ pip install install oauthenticator
$ jupyter serverextension enable --py nbserverproxy
$ jupyterhub --generate-config

# Not sure if now needed with github login method
$ openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -subj '/CN=XX.XXX.XXX.XXX'
[...]

PR #119, opened 2018-08-07
Default bind scheduler and dashboard to all network adresses

### LINKED ISSUE #90: Allow global access to dashboard by default

guillaumeeb:
Closes #90.
Solve my issues in #111.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -14,8 +14,7 @@ import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
 from distributed.diagnostics.plugin import SchedulerPlugin
-from distributed.utils import (
-    format_bytes, parse_bytes, tmpfile)
+from distributed.utils import (format_bytes, parse_bytes, tmpfile)
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


dask_jobqueue/core.py
```diff
@@ -16,7 +15,7 @@
 from distributed.deploy import Cluster
 from distributed.diagnostics.plugin import SchedulerPlugin
 from distributed.utils import (
-    format_bytes, get_ip_interface, parse_bytes, tmpfile)
+    format_bytes, parse_bytes, tmpfile)
```
jhamman: I don't think this line break is necessary anymore.

guillaumeeb:
Okay, just a random test failure here. Ready to merge if no further comment.

lesteve:
Nice! Thanks to this I can now access run a Jupyter notebook and access the dashboard on the same host at `http://localhost/proxy/<bokeh_server_port>/status`.

PR #121, opened 2018-08-10
Fix link to Cori

wgustafson:
The link to Cori was pointing to Cheyenne. Here's a fix.


guillaumeeb:
Thanks for spotting this one! Merging.

PR #129, opened 2018-08-16
Use correct scheduler address to connect to when using interface keyword

### LINKED ISSUE #126: Adaptive prematurely killing workers
I think this may be related to #122 but not sure so I'll post it as it's own issue.

When I start up a PBS cluster with adaptive, it starts up the minimum number of workers, but then kills all of them after about 10 seconds. This is not the case if I scale without using adaptive. I'm attaching a notebook to show a simple example and other information. 

[Adaptive Example](https://gist.github.com/josephhardinee/23594e44d943a6f97fc689842d501952)

guillaumeeb:
Closes #126.

I'm not completely convinced about the solution here. The thing is, we need to keep binding to all network interface, at least for Bokeh Dashboard, as demanded here: #90. Keeping listening to all network interface for distributed scheduler is nice too I think, but not mandatory.

So another solution from what is proposed here could be to go with a solution as proposed by @mrocklin in https://github.com/dask/dask-jobqueue/issues/90#issue-340311408, and stick to the use of self.scheduler.address as a URL to point workers to.

Any advice on this welcome.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -6,6 +6,7 @@ import shlex
 import subprocess
 import sys
 import warnings
+import socket
 from collections import OrderedDict
 from contextlib import contextmanager
 
@@ -14,7 +15,7 @@ import docrep
 from distributed import LocalCluster
 from distributed.deploy import Cluster
 from distributed.diagnostics.plugin import SchedulerPlugin
-from distributed.utils import (format_bytes, parse_bytes, tmpfile)
+from distributed.utils import (format_bytes, parse_bytes, tmpfile, get_ip_interface)
 
 logger = logging.getLogger(__name__)
 docstrings = docrep.DocstringProcessor()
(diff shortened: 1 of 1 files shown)


lesteve:
> So another solution from what is proposed here could be to go with a solution as proposed by @mrocklin in #90 (comment), and stick to the use of self.scheduler.address as a URL to point workers to.

@guillaumeeb I opened #135 which is closer to https://github.com/dask/dask-jobqueue/issues/90#issue-340311408. I am not sure this is a better option to be honest, let me know if you have any thoughts on it!

Side-comment: it'd be great to have tests for this but I don't really know how hard it is to have a IP address setup in docker containers that looks similar to the one of a typical cluster ...

guillaumeeb:
No idea for more complete testing, there is some though already in test_jobqueue_core.py.

Closing this one in favor of #135 that I found cleaner.

PR #131, opened 2018-08-22
WIP: Use regex for job-id lookup

willirath:
There's pbs schedulers returning with more verbose returns like: "Request 123456.asdfqwer submitted to queue: standard."

- [x] ~~f0566c8 should be fully compatible to the former version.~~
- [x] ~~It's probably a good thing to make the regex configurable.  I'll go ahead and add this as well.~~
- [x] Moving the regex matching in core.py
- [x] Making it a static string in the code, non configurable, with a value of r'\d+'
- [x] Deleting all _job_id_from_submit_output from JobQueueCluster subclasses
- [x] Adding tests
- [x] Updating trouble shooting section of the docs

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -409,16 +409,20 @@ class JobQueueCluster(Cluster):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        msg = ('Could not parse job id from submission command output. Job id '
-               'regexp is {}, submission command output is: {}'.format(
-                    self.job_id_regexp, out))
+        no_match_msg = ('Could not parse job id from submission command '
+                        'output. Job id regexp is {}, submission command '
+                        'output is: {}'.format(self.job_id_regexp, out))
 
         match = re.search(self.job_id_regexp, out)
         if match is None:
-            raise ValueError(msg)
+            raise ValueError(no_match_msg)
 
-        job_id = match.group('job_id')
+        no_job_id_msg = ("You need to use a `job_id` named group, e.g. "
+                         "'(?P<job_id>\d+),' in your regexp. Your regexp was: "
+                         "{}".format(self.job_id_regexp))
+
+        job_id = match.groupdict().get('job_id')
         if job_id is None:
-            raise ValueError(msg)
+            raise ValueError(no_job_id_msg)
 
         return job_id
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import logging
 import math
+import re
 import shlex
 import subprocess
 import sys
@@ -407,5 +408,4 @@ class JobQueueCluster(Cluster):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        raise NotImplementedError('_job_id_from_submit_output must be implemented when JobQueueCluster is '
-                                  'inherited. It should convert the stdout from submit_command to the job id')
+        return re.findall(r'\d+', out)[0]
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -139,6 +139,7 @@ class JobQueueCluster(Cluster):
     cancel_command = None
     scheduler_name = ''
     _adaptive_options = {'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
+    job_id_regexp = r'\d+'
 
     def __init__(self,
                  name=None,
@@ -407,6 +408,12 @@ class JobQueueCluster(Cluster):
             del self.pending_jobs[job_id]
         return jobs
 
-    @staticmethod
-    def _job_id_from_submit_output(out):
-        return re.findall(r'\d+', out)[0]
+    def _job_id_from_submit_output(self, out):
+        regexp = r'(?P<job_id>{}'.format(self.job_id_regexp)
+        match = re.search(regexp, out)
+        job_id = match.group('job_id')
+        if job_id is None:
+            raise ValueError('Could not parse job id from submission command'
+                             'output. Job id regexp is {}, submission command'
+                             'output is: {}'.format(self.job_id_regexp, out))
+        return job_id
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -409,10 +409,16 @@ class JobQueueCluster(Cluster):
         return jobs
 
     def _job_id_from_submit_output(self, out):
+        msg = ('Could not parse job id from submission command output. Job id '
+               'regexp is {}, submission command output is: {}'.format(
+                    self.job_id_regexp, out))
+
         match = re.search(self.job_id_regexp, out)
+        if match is None:
+            raise ValueError(msg)
+
         job_id = match.group('job_id')
         if job_id is None:
-            raise ValueError('Could not parse job id from submission command'
-                             'output. Job id regexp is {}, submission command'
-                             'output is: {}'.format(self.job_id_regexp, out))
+            raise ValueError(msg)
+
         return job_id
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -409,7 +409,7 @@ class JobQueueCluster(Cluster):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        regexp = r'(?P<job_id>{}'.format(self.job_id_regexp)
+        regexp = r'(?P<job_id>{})'.format(self.job_id_regexp)
         match = re.search(regexp, out)
         job_id = match.group('job_id')
         if job_id is None:
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -409,20 +409,18 @@ class JobQueueCluster(Cluster):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        no_match_msg = ('Could not parse job id from submission command '
-                        'output. Job id regexp is {}, submission command '
-                        'output is: {}'.format(self.job_id_regexp, out))
-
         match = re.search(self.job_id_regexp, out)
         if match is None:
-            raise ValueError(no_match_msg)
-
-        no_job_id_msg = ("You need to use a `job_id` named group, e.g. "
-                         "'(?P<job_id>\d+),' in your regexp. Your regexp was: "
-                         "{}".format(self.job_id_regexp))
+            msg = ('Could not parse job id from submission command '
+                   "output.\nJob id regexp is {!r}\nSubmission command "
+                   'output is:\n{}'.format(self.job_id_regexp, out))
+            raise ValueError(msg)
 
         job_id = match.groupdict().get('job_id')
         if job_id is None:
-            raise ValueError(no_job_id_msg)
+            msg = ("You need to use a 'job_id' named group in your regexp, e.g. "
+                   "r'(?P<job_id>\d+)', in your regexp. Your regexp was: "
+                   "{!r}".format(self.job_id_regexp))
+            raise ValueError(msg)
 
         return job_id
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -139,7 +139,7 @@ class JobQueueCluster(Cluster):
     cancel_command = None
     scheduler_name = ''
     _adaptive_options = {'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
-    job_id_regexp = r'(?P<job_id>{})'
+    job_id_regexp = r'(?P<job_id>\d+)'
 
     def __init__(self,
                  name=None,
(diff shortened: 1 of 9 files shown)


willirath:
> Out of curiosity, what PBs versions do you work with?

It's _NQSII_. I'm running this on an NEC Linux cluster:

```bash
$ qstat -V
NQSII CUI Version R03.00 / API Version R03.00 (linux)
```

willirath:
> It's NQSII. I'm running this on an NEC Linux cluster:

Once I have a fully working setup, I can add something to the example configs.

lesteve:
Not familiar with PBS but is there no command line argument you can pass to PBS qsub that just outputs the job-id (there was one for SGE qsub which is why I am asking).

I am not convinced about making the regex configurable, but maybe I am grossly underestimating the variability of possible qsub outputs (maybe each cluster defines its own qsub wrapper which prints more information to the stdout, or maybe it depends on the PBS version)

guillaumeeb:
@willirath, will you be OK to modify your PR according to what has been discussed in #132:
* Moving the regex matching in core.py
* Making it a static string in the code, non configurable, with a value of `r'\d+'`
* Deleting all `_job_id_from_submit_output` from JobQueueCluster subclasses

willirath:
I'll do that.

willirath:
> Making it a static string in the code, non configurable, with a value of r'\d+'

Just to clarify:  We expect this regex to be working for **all** cluster?  This makes it much simpler, then.

guillaumeeb:
> Just to clarify: We expect this regex to be working for all cluster? This makes it much simpler, then.

Yes, that's the idea, and it should work.

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -44,3 +46,14 @@ def test_forward_ip():
     with PBSCluster(walltime='00:02:00', processes=4, cores=8, memory='28GB',
                     name='dask-worker') as cluster:
         assert cluster.local_cluster.scheduler.ip == default_ip
+
+
+@pytest.mark.parametrize('qsub_return_string',
+                         ['Request {jobid}.asdf was sumbitted to queue 12.',
+                          '{jobid}',
+                          '  <{jobid}>  ',
+                          '{jobid}; asdf'])
```
guillaumeeb: Nice!

Might I ask you to use complete output as examples provided in https://github.com/dask/dask-jobqueue/issues/132#issuecomment-415306612?

dask_jobqueue/core.py
```diff
@@ -294,7 +295,7 @@ def start_workers(self, n=1):
         for _ in range(num_jobs):
             with self.job_file() as fn:
                 out = self._submit_job(fn)
-                job = self._job_id_from_submit_output(out.decode())
+                job = _job_id_from_submit_output(out.decode())
```
guillaumeeb: Are we sure we don't want to leave it as a class method?

Just asking, I don't really have an opinion yet, and I am not a really experienced python user.

If one day we want to be able to add the regex to config file, then won't it be a problem?

dask_jobqueue/tests/test_jobqueue_core.py
willirath: Yes, I'll do this.

dask_jobqueue/core.py
willirath: I also don't have any strong opinion on this.  Just went for the minimal amount of code necessary to implement (and test) this.

willirath: > - HTCondor (taken somewhere on internet):
> ```
> Submitting job(s)con.
> Logging submit event(s).
> 1 job(s) submitted to cluster 6075.
> 
> # but there is the -terse option that should make it like
> 10744.0 - 10744.2
> ```

Looks like a strong case for leaving the parser a static method of the class.  (This way, it would be easy to create an `HTCondorCluster` with a different parser.

I'll change this.

willirath:
I think I have addressed everything. Should I rebase this into a nicer PR?

dask_jobqueue/core.py
```diff
@@ -406,6 +407,6 @@ def _del_pending_jobs(self):
             del self.pending_jobs[job_id]
         return jobs
 
-    def _job_id_from_submit_output(self, out):
-        raise NotImplementedError('_job_id_from_submit_output must be implemented when JobQueueCluster is '
-                                  'inherited. It should convert the stdout from submit_command to the job id')
+    @staticmethod
```
guillaumeeb: Is this annotation really required?

I'm not sure how it will be handled if a subclass needs to override it?

docs/index.rst
```diff
@@ -285,5 +285,4 @@ problems are the following:
   We use submit command stdout to parse the job_id corresponding to the
   launched group of worker. If the parsing fails, then dask-jobqueue won't work
   as expected and may throw exceptions. You can have a look at the parsing
-  function in every ``JobQueueCluster`` implementation, see
-  ``_job_id_from_submit_output`` function.
+  function in the ``JobQueueCluster._job_id_from_submit_output`` function.
```
guillaumeeb: Duplicated _function_ here, maybe we should formulate differently.

dask_jobqueue/core.py
```diff
@@ -406,6 +407,6 @@ def _del_pending_jobs(self):
...
+    def _job_id_from_submit_output(out):
+        return re.findall(r'\d+', out)[0]
```
lesteve: A few comments about this:
* I am in favour of using a class variable for the regexp pattern. This way the regexp pattern can be tweaked more easily in derived classes. It can also be tweaked in the case where the cluster config is a bit special if needed:
```py
cluster = MyCluster(...)
cluster.job_id_regexp = my_special_regexp
```
* I think the best way is to have a regexp with a named match: r`(?<job_id>:\d+)` by default. This way you can use the name of the match which is more robust:
```py
match = re.search(self.job_id_regexp, out)
job_id = match.groupdict.get('job_id')
```
* There should be better error treatment
```py
if job_id is None:
    raise ValueError('Could not parse job id from submission command output. Job id regexp is {}, submission command output is: {}'.format(self.job_id_regexp, out)
```

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -44,3 +44,19 @@ def test_forward_ip():
     with PBSCluster(walltime='00:02:00', processes=4, cores=8, memory='28GB',
                     name='dask-worker') as cluster:
         assert cluster.local_cluster.scheduler.ip == default_ip
+
+
+@pytest.mark.parametrize(
+    'qsub_return_string',
+    ['{jobid}.admin01',
```
lesteve: Nice test, could you use `job_id` for consistency reaons?

willirath:
f72a7f1 should address all your points @lesteve. Let me know what you think.

willirath:
I went for only having `r'\d+'` in the class var and then constructing the named group regexp in the function.  Should make it easier for a user to override at runtime.

dask_jobqueue/core.py
```diff
@@ -138,6 +139,7 @@ class JobQueueCluster(Cluster):
     cancel_command = None
     scheduler_name = ''
     _adaptive_options = {'worker_key': lambda ws: _job_id_from_worker_name(ws.name)}
+    job_id_regexp = r'\d+'
```
lesteve: Put the full regexp here, i.e. `(?P<job_id>\d+)`. This way this is fully customisable in derived class. Currently you are assuming that the regexp starts with `(?<job_id>` in `_job_id_from_submit_output` which may not be appropriate.

dask_jobqueue/core.py
```diff
@@ -407,5 +409,11 @@ def _del_pending_jobs(self):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        raise NotImplementedError('_job_id_from_submit_output must be implemented when JobQueueCluster is '
-                                  'inherited. It should convert the stdout from submit_command to the job id')
+        regexp = r'(?P<job_id>{})'.format(self.job_id_regexp)
+        match = re.search(regexp, out)
```
lesteve: I think to cover all the cases that can error with a not very helpful error message, you need to do something like this:
```py
msg = 'Could not parse ...'

if match is None:
    raise ValueError(msg)

job_id = match.groupdict().get('job_id')

if job_id is None:
    raise ValueError(msg)
```

It would be great to have a test for error messages: you can use something like:
```py
with pytest.raises(ValueError, match=error_message_regexp):
    ...
```

dask_jobqueue/core.py
```diff
@@ -407,5 +409,10 @@ def _del_pending_jobs(self):
         return jobs
 
     def _job_id_from_submit_output(self, out):
-        raise NotImplementedError('_job_id_from_submit_output must be implemented when JobQueueCluster is '
-                                  'inherited. It should convert the stdout from submit_command to the job id')
+        match = re.search(self.job_id_regexp, out)
+        job_id = match.group('job_id')
+        if job_id is None:
+            raise ValueError('Could not parse job id from submission command'
```
lesteve: Add a space at the end of each line otherwise the text will be something like "submission commandoutput"

willirath: Done.

dask_jobqueue/core.py
willirath: Done.

dask_jobqueue/core.py
willirath: Done.

dask_jobqueue/core.py
willirath: The `match` keyword gets only a part of the expected error message. Not sure if this is enough. Looks similar to similar tests in `dask-ml`, though.

willirath:
@lesteve What do you think?

dask_jobqueue/core.py
```diff
@@ -407,5 +409,16 @@ def _del_pending_jobs(self):
...
-                                  'inherited. It should convert the stdout from submit_command to the job id')
+        msg = ('Could not parse job id from submission command output. Job id '
+               'regexp is {}, submission command output is: {}'.format(
+                    self.job_id_regexp, out))
+
+        match = re.search(self.job_id_regexp, out)
+        if match is None:
+            raise ValueError(msg)
+
+        job_id = match.group('job_id')
```
lesteve: Can you use `job_id = match.groupdict().get('job_id')` to have a better error as I mentioned (probably folded away in an outdated diff).

Thinking about this it would be great to have a better error in the case when there is a match but no `job_id` named group, i.e. something like:

You need to use a `job_id` named group in your regexp, e.g. '(?P<job_id>\d+)'. Your regexp was: {}

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -44,3 +44,50 @@ def test_forward_ip():
...
+@pytest.mark.parametrize('Cluster', [PBSCluster, MoabCluster, SLURMCluster,
+                                     SGECluster, LSFCluster])
+@pytest.mark.parametrize(
+    'qsub_return_string',
+    ['Request {job_id}.asdf was sumbitted to queue: standard.',
+     'sbatch: Submitted batch job {job_id}',
+     '{job_id};cluster',
+     'Job <{job_id}> is submitted to default queue <normal>.',
+     '{job_id}',
+     pytest.param('{job_id}.admin01', marks=pytest.mark.xfail)])
```
lesteve: I forgot you could do something like this with parametrize, nice! Can you explain why it is xfail though?

willirath: In this case `{job_id}` expands to `XXXXX` (see <https://github.com/dask/dask-jobqueue/pull/131/files/9c83c296c33421c2ec496580b67a81feaa58f6b1#diff-6715b450c58d2e22119574769a68a1d6R92>).  Then, the `01` in `admin01` would be interpreted as the `job_id`.

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -44,3 +44,50 @@ def test_forward_ip():
...
+    qsub_return_string = qsub_return_string.format(job_id=original_job_id)
+    with Cluster(walltime='00:02:00', processes=4, cores=8, memory='28GB',
+                 name='dask-worker') as cluster:
+        assert (original_job_id
+                == cluster._job_id_from_submit_output(qsub_return_string))
+
+
+@pytest.mark.parametrize('Cluster', [PBSCluster, MoabCluster, SLURMCluster,
+                                     SGECluster, LSFCluster])
+@pytest.mark.parametrize(
```
lesteve: General remark, for the error cases I would remove this parametrize and just test with something like this:
* an output without match: e.g. `'there is no number here'`
* a regexp that match but that does not have a `job_id` named group, e.g. `cluster.job_id_regexp = r'\d+'` and an output `'Job <12345> is submitted to default queue <normal>.`

willirath:
@lesteve I think, I've addressed everything, now.

(I won't probably get back to this before Wednesday. So feel free to push stuff to this branch. :) )

lesteve:
I pushed some small tweaks, merging, thanks a lot @willirath!

jhamman:
I just want to say, I am a fan of the general approach here. I'm also glad to see some tests for parsing these job ids. Thanks to all of you for ticking this one off.

PR #135, opened 2018-08-22
Alternative way of listening on all interfaces for the bokeh server

### LINKED ISSUE #126: Adaptive prematurely killing workers

lesteve:
This is an alternative to #129. Not sure which one is preferable to be completely honest ... one difference that may matter is that the scheduler does not listen on all interfaces. 

This is more in the spirit of https://github.com/dask/dask-jobqueue/issues/90#issue-340311408. I discovered through looking at the source code, that you can pass a `(address, port)` address fine to the `diagnostics_port` despite its name. Thanks to this, the bokeh diagnostics server will listen on all interfaces.

Close #126.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -199,7 +199,7 @@ class JobQueueCluster(Cluster):
             extra += ' --interface  %s ' % interface
             kwargs.setdefault('ip', get_ip_interface(interface))
         else:
-            kwargs.setdefault('ip', socket.gethostname())
+            kwargs.setdefault('ip', '')
 
         # Bokeh diagnostics server should listen on all interfaces
         diagnostics_ip_and_port = ('', 8787)
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -8,13 +8,14 @@ import sys
 import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
+import socket
 
 import dask
 import docrep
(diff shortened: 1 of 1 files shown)


guillaumeeb:
And by the way, should we update distributed keyword name? This could break other things I imagine.

lesteve:
> I think I prefer this way, except I would still bind scheduler itself on all interfaces by default if no interface is provided. This should prove compatible with more network setup as default, even if only on edge cases.

Can you give a quick example of a cluster setup when it is useful to have the scheduler listen on all interfaces?

guillaumeeb:
No I cannot :), I just suspect it could happen. But in such a case workaround would be to pass ip kwargs when initializing cluster object. So nevermind.

guillaumeeb:
@lesteve feel free to revert your last commit, as discussed in previous comment, I don't have a strong argument for listening on all interfaces for scheduler, and if it is needed it can easily be done with kwargs.

guillaumeeb:
I will merge this once travis is OK.

PR #136, opened 2018-08-24
Adding info for scaling with cores and memory

### LINKED ISSUE #130: Use number of cores as a parameter to scale()
We've recently switched from number of jobs to number of worker processes as an argument to `scale` method. This is good for working with adaptive cluster as defined in distributed, but somehow not really easy to deal with while using dask-jobqueue, at least for me.

Ideally, we would use unitary job that spawn an entire node of our cluster. But in real life, I often find myself changing cores and process parameters between use cases, or to better fit with the current use of my cluster, when there is other users' job that are not using full nodes for example.

And at the end, when using scale method, what I want is to scale to a given number of CPU cores, no matter the number of jobs or workers or process or threads. So I'm left doing the (simple) maths of translating between number of cores and number of processes given my JobqueueCluster initialization.

It would be really easier if we could have the option to specify a cores kwarg or something like this to scale method.

guillaumeeb:
Closes #130. But should wait for https://github.com/dask/distributed/pull/2209 to be merged before going in.

An implementation of worker information to be able to scale with `cores` or `memory` kwargs.

This works fine on my setup.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -204,7 +204,7 @@ class JobQueueCluster(Cluster):
 
         # Keep information on process, threads and memory, for use in
         # subclasses
-        self.worker_memory = parse_bytes(memory) if memory is not None else None
+        self.worker_memory = parse_bytes(memory)
         self.worker_processes = processes
         self.worker_cores = cores
         self.name = name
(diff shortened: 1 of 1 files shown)


guillaumeeb:
Closed in favor of #184.

PR #137, opened 2018-08-24
OARCluster implementation

lesteve:
[OAR](https://oar.imag.fr/) is used quite a bit in France at Inria centers as well as on [Grid5000](https://www.grid5000.fr/mediawiki/index.php/Grid5000:Home) and possibly other places that I don't know about.

This needs to wait for #131 to be merged. Just putting it out there in case someone wants to have a look and give early feed-back.

I tested this on a local cluster and everything seems to be working AFAICT.

I added a few simple tests, that only test the script produced and consequently don't need an OAR cluster to run.

The main OAR specificity I came across (copying and pasting my comment in the code below):
the submission script needs to exist on the worker **when the job starts on the worker**. This is different from other schedulers that only need the script on the submission node **at submission time**. That means that we can not use the same strategy as in JobQueueCluster: create a temporary submission file, submit the script, delete the submission file.

This is the most controversial part: in order to reuse the code in the base JobQueueCluster class, we read from the temporary file and reconstruct the full command line to execute. The script is passed in as a string (inline script in OAR jargon) rather than as a filename. 

That means that the `self.job_script()` is not really the file used in the scheduler command that gets run, I feel this is the simplest way to get going and also it still has the advantage of being able to look at `self.job_script()` to see the script that is being run (provided that my submission script to command-line translation does not have any bug).

Side-comment: you can see the specificity of the regex I used as well. `\d+` is not good enough because there are other numbers in the output. Here is an example:
```
$ oarsub 'echo test'
[ADMISSION RULE] Set default walltime to 7200.
[ADMISSION RULE] Modify resource description with type constraints
[SHORT QUEUE] This job is routed into the short queue
[CORES COUNT] You requested 0 cores
Generate a job key...
OAR_JOB_ID=91223
```

### DIFF
--- a/dask_jobqueue/oar.py
+++ b/dask_jobqueue/oar.py
@@ -46,6 +46,7 @@ class OARCluster(JobQueueCluster):
     submit_command = 'oarsub'
     cancel_command = 'oardel'
     scheduler_name = 'oar'
+    job_id_regexp = r'OAR_JOB_ID=(?P<job_id>\d+)'
 
     def __init__(self, queue=None, project=None, resource_spec=None, walltime=None, job_extra=None, **kwargs):
         if queue is None:
@@ -89,16 +90,6 @@ class OARCluster(JobQueueCluster):
         self.job_header = '\n'.join(header_lines)
 
         logger.debug("Job script: \n %s" % self.job_script())
-
-    def _job_id_from_submit_output(self, out):
-        job_id_regexp = r'OAR_JOB_ID=(?P<job_id>\d+)'
-        match = re.search(job_id_regexp, out)
-        if match is None:
-            raise ValueError('Unexpected output from the submission command')
-        group_dict = match.groupdict()
-        if 'job_id' not in group_dict:
-            raise ValueError('Unexpected output from the submission command')
-        return group_dict['job_id']
 
     def _submit_job(self, fn):
         # OAR specificity: the submission script needs to exist on the worker
--- a/dask_jobqueue/oar.py
+++ b/dask_jobqueue/oar.py
@@ -1,0 +1,122 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+import re
+import shlex
+
+import dask
+
+from .core import JobQueueCluster, docstrings
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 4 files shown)


dask_jobqueue/oar.py
```diff
@@ -0,0 +1,122 @@
...
+        # the base JobQueueCluster class, we read from the temporary file and
+        # reconstruct the command line where the script is passed in as a
+        # string (inline script in OAR jargon) rather than as a filename.
+        with open(fn) as f:
+            content_lines = f.readlines()
+
+        oar_lines = [line for line in content_lines if line.startswith('#OAR ')]
+        oarsub_options = [line.replace('#OAR ', '').strip() for line in oar_lines]
+        inline_script_lines = [line for line in content_lines if not line.startswith('#')]
+        inline_script = ''.join(inline_script_lines)
```
guillaumeeb: Won't this cause problem to join on empty string? Joining 'JOB_ID=${OAR_JOB_ID}' with dask-worker command won't lead to something weird?

lesteve: With `.readlines()` you still have the `\n` at the end of the lines.

guillaumeeb: Nevermind, sorry.

lesteve:
> Just asking, isn't it possible to copy the created tmp file into a real job script file? If I'm not mistaken, we only need one file for submitting every job, as all generated script files are identical. You could copy this one file and use it for every submission. And maybe find a way to delete it upon cluster shutdown...

I thought a bit about this and basically I feel the way it is currently done is a good trade-off: 
* At the moment at least the OAR specificity is contained in a single function rather that leaking into the core logic in multiple places.
* having a cleanup logic after the job starts running on the worker seems fragile. Side-comment: there is a OAR specific functionality to be notified when the job start so we could potentially use that.
* also there are caveat about where the submission file needs to be. Basically the folder needs to be the same on the exist on the worker so we would need to have another parameter for specifying where to put the submission file.

guillaumeeb:
> I thought a bit about this

Yeah I can imagine! Your arguments make real sense. It's fine for me.

lesteve:
I am going to merge this one. I'll bug some people bug some people to try to get feedback on this.

PR #140, opened 2018-08-28
Add a cluster detailed status method

### LINKED ISSUE #11: Implement a cluster status method, to now if worker are really running

guillaumeeb:
Closes #11.

Main concern here: is this really useful in addition to cluster.__repr__ or client.__repr__.
Personnaly, I like to have kind of detailed status, but find it only useful in some specific cases.

Any other feedback quite welcome.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -6,7 +6,7 @@ import shlex
 import subprocess
 import sys
 import warnings
-import html
+from xml.sax.saxutils import escape
 from collections import OrderedDict
 from contextlib import contextmanager
 
@@ -458,7 +458,7 @@ class ClusterStatus():
         if self.verbose:
             text += '    <ul>\n'
             for job_id, workers in self.cluster.running_jobs.items():
-                text += "      <li><b>%s: </b> %s\n" % (job_id, html.escape(str(workers)))
+                text += "      <li><b>%s: </b> %s\n" % (job_id, escape(str(workers)))
             text += "    </ul>\n"
         text += "  <li><b>Pending: </b>(cores=%d, memory=%s, workers=%d, jobs=%d)\n" % \
             self.pending_stats
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -428,11 +428,11 @@ class ClusterStatus():
         running_cores = running_workers * self.cluster.worker_threads
         running_memory = running_workers * self.cluster.worker_memory / self.cluster.worker_processes
         self.running_stats = (running_cores, format_bytes(running_memory), running_workers,
-                  len(self.cluster.running_jobs))
+                              len(self.cluster.running_jobs))
         self.pending_stats = (len(self.cluster.pending_jobs) * self.cluster.worker_cores,
-                   format_bytes(len(self.cluster.pending_jobs) * self.cluster.worker_memory),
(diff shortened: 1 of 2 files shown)


guillaumeeb:
Here is how it looks:

image

mrocklin:
I like the idea here.  Some thoughts:

1.  Is this something that we want to coordinate with other cluster managers (dask-yarn, dask-kubernetes).  If so, how?  cc @jcrist @jacobtomlinson 
2.  If yes, then what kind of data would we standardize around that is common to all three.  Should we also provide this data in a structured way?
2.  Would this be better as a dashboard plot or ipywidget that updated in real time?

jhamman:
> Is this something that we want to coordinate with other cluster managers (dask-yarn, dask-kubernetes). If so, how? cc @jcrist @jacobtomlinson

I'd argue that the running/pending/finished job/worker classification is going to be widely applicable (even for a LocalCluster). That said, I don't know enough about the other systems to generalize these classifications.

Also, +1 on adding something like this to the dashboard.

jacobtomlinson:
These would definitely be useful in `dask-kubernetes`. Running and Pending would be easy to implement, completed workers get cleaned up automatically so that doesn't really work for that.

jhamman:
Is this worth reviving? I still think this would be a worthwhile addition.

guillaumeeb:
I guess I didn't know what to do with @mrocklin comments... I feel this is applicable to other cluster managers, but I don't know how to standardize it, so that it can be easily migrated upstream and then in other projects. How to represent job vs pod in dask-kubernetes vs only processes in LocalCluster...

And probably was waiting for some more discussion on the form it should take.

mrocklin:
I'm happy to not standardize things for a while.  Please don't consider my comments as blockers on any progress here.

guillaumeeb:
So what do we want here to do here:

- I'm not sure the simple html representation is really useful, is it?
- Do anyone has a proposal on the format? Dashboard (I don't know how to do dashboard addition, and this may have to go upstream?), Widget, simple HTML, some table like Dataframes?
- What can we use to make the job/pod/other notion generic? Use some key if it exist, like our worker_key that adaptive uses? Anything else?

I guess I need more precise feedback :slightly_smiling_face:.

mrocklin:
My first thought was to suggest that the cluster object have a `cluster.status()` method or something similar return a standarized data structure that could be placed into a table.  A dashboard page could then be built that called that function periodically.

However, then I thought about how we might separate the cluster manager from the scheduler, at which point we'll no longer have access to the dashboard (except by explicitly passing messages).  It might be that this information might also be expressible through something like a JupyterLab extension, similar to what @ian-r-rose is building at https://github.com/dask/dask-labextension/pull/31

jacobtomlinson:

PR #143, opened 2018-08-29
Use released version of dask and distributed in CI

lesteve:
I think it was done like this because we needed development versions of dask and distributed at one point. But that's not the case anymore.

### DIFF
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -5,8 +5,6 @@ function jobqueue_before_install {
   ./ci/conda_setup.sh
   export PATH="$HOME/miniconda/bin:$PATH"
   conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION dask distributed flake8 pytest docrep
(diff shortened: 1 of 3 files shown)


jhamman:
I think this will be true after the next release of distributed. I hear this may happen tomorrow...

lesteve:
> I think this will be true after the next release of distributed. I hear this may happen tomorrow...

Interesting, I did not remember we needed distributed master. Can you remind me what it was?

The test seem to pass though ...

guillaumeeb:
IIRC in #97 there was an update on distributed side.

jhamman:
https://github.com/dask/distributed/pull/2171 recently went in to help fix some of the issues in #97.

lesteve:
I think this one should be merged. I don't know of any reason to use the development version of dask in the CIs. I'll push an empty commit to retrigger the CIs just to err on the cautious side.

jhamman:
Tests are green here. At some point we may want to test against both distributed/master and the latest release.

lesteve:
Merging this one.

PR #145, opened 2018-08-30
Add a nested folder for storing jobqueue log file, use separate files…

### LINKED ISSUE #141: Change stdout and stderr default setting in LSF and Slurm cluster implementation
As mentionned in #122 and #139, default values for job output files in LSF and Slurm implementation are
````python
"-e %s.err" % self.name
"-o %s.out" % self.name
````

If those configuration parameters have similar behavior as those from PBSPro, then this means every job submitted by dask-jobqueue will use this file name for writting its stdout and stderr when finished. So at one time, we will have only the last finished job stderr and stdout. 

I'm not sure this is what we want. At least when trying to understand some bug, this can be problematic. I propose to use templating propose by both scheduler, e.g. for Slurm, something like (maybe with escaping char):
````python
"-e %s-%j.err" % self.name
"-o %s-%j.out" % self.name
````

On the other hand, if we specify a value different for each job, we will get (as is happening currently with default PBSCluster implementation) the submission folder flooded with jobs stdout and stderr file.

### PR DESCRIPTION
**Pull request by guillaumeeb**
… for each job with LSF and SLURM

Closes #141

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import logging
 import math
+import os
 import re
 import shlex
 import subprocess
@@ -151,6 +152,7 @@ class JobQueueCluster(Cluster):
                  local_directory=None,
                  extra=None,
                  env_extra=None,
+                 log_directory=None,
                  walltime=None,
                  threads=None,
                  **kwargs
@@ -183,6 +185,8 @@ class JobQueueCluster(Cluster):
             extra = dask.config.get('jobqueue.%s.extra' % self.scheduler_name)
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
+        if log_directory is None:
+            log_directory = dask.config.get('jobqueue.%s.log-directory' % self.scheduler_name)
 
         if dask.config.get('jobqueue.%s.threads', None):
             warnings.warn(threads_deprecation_message)
@@ -240,6 +244,10 @@ class JobQueueCluster(Cluster):
             self._command_template += " --local-directory %s" % local_directory
         if extra is not None:
             self._command_template += extra
+        if log_directory is not None:
+            self.log_directory = log_directory
+            if not os.path.exists(self.log_directory):
+                os.mkdir(self.log_directory)
 
     def __repr__(self):
         running_workers = sum(len(value) for value in self.running_jobs.values())
@@ -290,7 +298,8 @@ class JobQueueCluster(Cluster):
             yield fn
 
     def _submit_job(self, script_filename):
-        return self._call(shlex.split(self.submit_command) + [script_filename])
+        return self._call(shlex.split(self.submit_command) + [script_filename],
+                          cwd=self.log_directory)
 
     def start_workers(self, n=1):
         """ Start workers and point them to our local scheduler """
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -244,10 +244,11 @@ class JobQueueCluster(Cluster):
             self._command_template += " --local-directory %s" % local_directory
         if extra is not None:
             self._command_template += extra
-        if log_directory is not None:
-            self.log_directory = log_directory
+
+        self.log_directory = log_directory
+        if self.log_directory is not None:
             if not os.path.exists(self.log_directory):
-                os.mkdir(self.log_directory)
+                os.makedirs(self.log_directory)
 
     def __repr__(self):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: dask_jobqueue/core.py [hunk 1]
```diff
@@ -290,7 +298,8 @@ def job_file(self):
             yield fn
 
     def _submit_job(self, script_filename):
-        return self._call(shlex.split(self.submit_command) + [script_filename])
+        return self._call(shlex.split(self.submit_command) + [script_filename],
+                          cwd=self.log_directory)
```
lesteve: Hmmm, this may have unintended side-effects, although I am not 100% sure ... for example if I specify some relative path (e.g. for data files) in my notebook they may not be found in the dask-worker process.

Is it not possible to pass a folder to `-e` and `-o` (SGE allows this but maybe this is very specific to SGE ...)?

I guess if I understand https://github.com/dask/dask-jobqueue/issues/141#issuecomment-417225928, there is no way to have variable expansion in PBS directive. I googled a bit and found http://community.pbspro.org/t/pbs-jobid-variable/176 which does not seem to leave too much hope ... I have seen "solutions" that used `qalter` after the submission e.g. [this](https://stackoverflow.com/a/28186725) maybe that's an option but that seems quite complicated ...

Just curious, so what do people do with PBSPro? They don't set the stdout/stderr file, they remember to change it each time with a different name?

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -79,3 +81,10 @@ def test_job_id_error_handling(Cluster):
             return_string = 'Job <12345> submited to <normal>.'
             cluster.job_id_regexp = r'(\d+)'
             cluster._job_id_from_submit_output(return_string)
+
+
+def test_log_directory():
```
lesteve: Nice pytest trick: you can use the `tmpdir` fixture. This gives you a unique folder on each run but pytest keeps it after the execution, allowing you to look at it in case something goes wrong. pytest keeps the last N (maybe 5) tmpdir folders IIRC. 

```py
def test_log_directory(tmpdir):

with PBSCluster(..., log_directory=tmpdir.strpath as cluster:
    ...
```

dask_jobqueue/slurm.py
```diff
@@ -73,8 +73,8 @@ def __init__(self, queue=None, project=None, walltime=None, job_cpu=None, job_me
         # SLURM header build
         if self.name is not None:
             header_lines.append('#SBATCH -J %s' % self.name)
-            header_lines.append('#SBATCH -e %s.err' % self.name)
-            header_lines.append('#SBATCH -o %s.out' % self.name)
+            header_lines.append('#SBATCH -e %s-%%j.err' % self.name)
```
lesteve: Just curious, I have wondered before, why not keep the default name? Is the default name not something very similar?

dask_jobqueue/core.py
guillaumeeb: Yes you are right, this may not be the best approach...

You've understood the problem with PBS, using -e or -o is possible but quite limited. Default output file name is $JOB_NAME.o$JOB_ID. So usually I just leave this and don't set anything.

Another solution is to just specify a folder. In this case PBS will fill it with $JOB_ID.OUT files.

dask_jobqueue/slurm.py
guillaumeeb: It was my first solution too, but I found that Slurm uses slurm-$JOB_ID.out as default. This is not too bad, but it misses the job name... So I'm hesitating here.

guillaumeeb:
So you're right, maybe cwd is not the best choice, and I've missed some degraded cases too. I will try to fix that. 

For Torque vs PBS, they were the same a long time ago, but they've diverged so are not fully compatible now. It maybe needed to implement a TorqueCluster at some point.

dask_jobqueue/core.py
lesteve: > Another solution is to just specify a folder. In this case PBS will fill it with $JOB_ID.OUT files.

If a specifying a folder is possible I think I would just do that for PBS (and SGE as well).

guillaumeeb:
So I used the `-e` and `-o` solution for specifying a folder. My only concern is that it adds some complexity (3 lines) in every JobQueueCluster implementation.

Should we leave it to user setting `job_extra` instead (with some documentation maybe ?), and removing -e and -o from every implementation?

guillaumeeb:
I'm gonna merge this one soon.

lesteve:
I just saw this now, very nice to see this merged!

guillaumeeb:
Thanks, I had doubts with this one, but it's really nice to not have undreds of output files inside my notebook folder!

PR #149, opened 2018-09-10
update env location

TomAugspurger:
The readthedocs.yml file overrides the dependencies file

cc @mrocklin


mrocklin:
Thanks @TomAugspurger

Issue #154, opened 2018-09-18
TypeError in dask-jobqueue/core.py when starting up workers (v0.4.0)


kaelancotter:
Error Seems pretty straightforward. Adding a quick int() fixed things on my end.  LMK if you want a PR and or CI coverage... (I didn't see a contributing guide anywhere in the docs)  For context grid is an SGECluster object.  Thanks for the great tool.

```
In [6]: grid.scale(2)

DEBUG:Scaling up to 2 workers.
DEBUG:Found 0 active/pending workers.
DEBUG:starting 2 workers
In [7]: tornado.application - ERROR - Exception in callback <functools.partial object at 0x7fba4979c208>
Traceback (most recent call last):
  File ".../miniconda/envs/dask/lib/python2.7/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
  File ".../miniconda/envs/dask/lib/python2.7/site-packages/tornado/stack_context.py", line 300, in null_wrapper
    return fn(*args, **kwargs)
  File ".../miniconda/envs/dask/lib/python2.7/site-packages/dask_jobqueue/core.py", line 375, in scale_up
    self.start_workers(n - active_and_pending)
  File ".../miniconda/envs/dask/lib/python2.7/site-packages/dask_jobqueue/core.py", line 299, in start_workers
    for _ in range(num_jobs):
TypeError: range() integer end argument expected, got float.
```

kaelancotter:
Looking a little deeper, its certainly a python 2.7 vs 3.x issue specifically https://www.python.org/dev/peps/pep-3141/

mrocklin:
Thanks for the bug report @kaelancotter .  A pull request would be welcome.  Contributor guidelines are available at documentation for the parent project: http://dask.pydata.org/en/latest/develop.html

kaelancotter:
I took a look through the travis ci logs for python 2.7 sge.  Right now it looks like the sge_master image in the compose stack that executes pytest uses a default 3.x python from miniconda3 instead of the travis_python_version.  I think fixing the container to use the travis python version will greatly improve 2.7 test coverage, including coverage of the original type error. I hipe Im not overstepping.  :)

guillaumeeb:
You're probably right, thanks for raising this too! It seems currently python 2.7 sge tests are useless.

It's really nice having other people stepping in this code, so you're very welcome here!

[...]

PR #157, opened 2018-09-22
Pass python version to docker container, see issue #154

kaelancotter:
Local building of the container indicates TRAVIS_PYTHON_VERSION is properly passed to the conda environment.  If this does work, travis ci might still fail on the sge 2.7 job.  I figure first attempt is free, but if something particularly unusual shows up in the CI logs, I'll try and setup Travis on my fork.  Wish me luck. ;)

### DIFF
--- a/ci/sge/Dockerfile-master
+++ b/ci/sge/Dockerfile-master
@@ -9,11 +9,7 @@ RUN curl -o miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-latest-L
     /opt/anaconda/bin/conda clean -tipy && \
     rm -f miniconda.sh
 ENV PATH /opt/anaconda/bin:$PATH
-RUN conda install -n root conda=4.4.11 && conda clean -tipy
-RUN conda install -c conda-forge dask distributed blas pytest mock ipython pip psutil && conda clean -tipy
-RUN pip install --no-cache-dir drmaa
-RUN pip install --no-cache-dir git+https://github.com/dask/dask.git --upgrade --no-deps
-RUN pip install --no-cache-dir git+https://github.com/dask/distributed.git --upgrade --no-deps
+RUN conda install -c conda-forge dask distributed pytest && conda clean -tipy
 
 COPY ./*.sh /
 COPY ./*.txt /
@@ -25,6 +21,5 @@ EXPOSE 6444
 EXPOSE 6445
 EXPOSE 6446
 
-ENV DRMAA_LIBRARY_PATH /usr/lib/gridengine-drmaa/lib/libdrmaa.so
 ENV SGE_ROOT /var/lib/gridengine/
 ENV SGE_CELL default
--- a/ci/sge/Dockerfile-master
+++ b/ci/sge/Dockerfile-master
@@ -9,7 +9,8 @@ RUN curl -o miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-latest-L
     /opt/anaconda/bin/conda clean -tipy && \
     rm -f miniconda.sh
 ENV PATH /opt/anaconda/bin:$PATH
-RUN conda install -c conda-forge dask distributed pytest && conda clean -tipy
+ARG TRAVIS_PYTHON_VERSION
+RUN conda install -c conda-forge python=$TRAVIS_PYTHON_VERSION dask distributed pytest && conda clean -tipy
 
 COPY ./*.sh /
 COPY ./*.txt /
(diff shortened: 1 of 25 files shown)


kaelancotter:
I think we can check for the environment variable in the start-sge.sh script and export a default value if not already set.

kaelancotter:
Can/should I change this to WIP for the time being...  I wasn't entirely expecting the current behavior.  There might be some additional 3.x dependency in the testing support scripts.

guillaumeeb:
> I think we can check for the environment variable in the start-sge.sh 

I would prefer to have that directly in the Docker images if possible.

> Can/should I change this to WIP for the time being...

Yes feeling free to do this if you need more time!

kaelancotter:
Looks like part of the issue was the use of  http.server vs SimpleHTTPServer in the run-master and run-slave scripts.  As far as locally running the various testing environments, is it worth building a new wrapper script? (This could/would also be a place for setting TRAVIS_PYTHON_VERSION)  LMK if there's already a unified/single script way for locally running the docker-compose based tests.

kaelancotter:
Side thought, should the slave containers be tied to TRAVIS_PYTHON_VERSION as in above?  Could be limited to just master.

ci/sge/Dockerfile-master
```diff
@@ -9,7 +9,9 @@ RUN curl -o miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-latest-L
     /opt/anaconda/bin/conda clean -tipy && \
     rm -f miniconda.sh
 ENV PATH /opt/anaconda/bin:$PATH
-RUN conda install -c conda-forge dask distributed pytest && conda clean -tipy
+ARG TRAVIS_PYTHON_VERSION
+ENV TRAVIS_PYTHON_VERSION=$TRAVIS_PYTHON_VERSION
```
guillaumeeb: Is the ENV line really needed, doesn't ARG suffice?

ci/sge/docker-compose.yml
```diff
@@ -6,6 +6,8 @@ services:
     build:
       context: .
       dockerfile: Dockerfile-master
+      args:
+        - TRAVIS_PYTHON_VERSION
```
guillaumeeb: So what will happen in local testing if TRAVIS_PYTHON_VERSION is not set?

ci/sge/run-master.sh
```diff
@@ -20,4 +20,8 @@ bash add_worker.sh dask.q slave_two 4
 
 sudo service gridengine-master restart
 
-python -m http.server 8888
+if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then
+    python -m SimpleHTTPServer 8888
+else
+    python -m http.server 8888
+fi
```
guillaumeeb: Launching this http server is just in order to have a blocking process when running the docker container. I don't really like complexifying this line, even if I don't have a better solution yet (maybe just use some blocking commands, see https://stackoverflow.com/questions/2935183/bash-infinite-sleep-infinite-blocking).

ci/sge/Dockerfile-master
kaelancotter: ENV allows the run master/slave scripts access to TRAVIS_PYTHON_VERSION for the blocking script below.

ci/sge/run-master.sh
kaelancotter: I'll try out some variants based on above and can hopefully remove both above conditional and Docker ENV

ci/sge/docker-compose.yml
kaelancotter: Current behavior uses 3.6.6, but I don't know that its explicitly supported conda behavior for `python=""  `

kaelancotter:
@guillaumeeb Is it alright to bump the docker-compose.yml version from 2 to 2.1?  Thats one option for default values in the compose stack.  [variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution) The other would be a `.env` file.  Replacement priority is listed here.  [env file](https://docs.docker.com/compose/environment-variables/#the-env-file).  If you or anyone else has a third suggestion I'm all ears.  Let me know which is preferable and I can hopefully get this PR wrapped up for ya.

guillaumeeb:
>Is it alright to bump the docker-compose.yml version from 2 to 2.1?

Sure! Compose 2.1 is already two years old, moreover we use docker only for CI tests, so we don't need backward compatibility here.

kaelancotter:
That should do it then.  Sorry about the crazy commit history.  I haven't done it before but I can try to rebase or something.

guillaumeeb:
And no worries for the commits, we're using squash and merge functionnality of github.

PR #158, opened 2018-09-24
Adding basic development guidelines to doc

guillaumeeb:
Following discussion in https://github.com/dask/dask-jobqueue/issues/154.

I've taken most of the text from distributed doc.


guillaumeeb:
Merging this, as this is not critical and may be useful to some newcomers.

PR #165, opened 2018-10-05
Missing an 's' in the link

ocaisa:


docs/configuration-setup.rst
```diff
@@ -26,7 +26,7 @@ Or as part of a configuration file:
    cluster = PBSCluster()
 
 For more information on handling configuration files see `Dask configuration
-documentation <https://dask.org/en/latest/configuration.html>`_.
+documentation <https://dask.org/en/latest/configurations.html>`_.
```
jhamman: Thanks for your contribution. I think this should actually be: `http://docs.dask.org/en/latest/configuration.html`

ocaisa: Ah ok, I just saw the dead link and didn't look too deeply...fixing now

ocaisa:
One of the travis builds failed for some reason

willirath:
Looks like connectivity issues at setup the resulted in a missing sge container and the job waiting "forever" (till the 50' time out) for it to come up?

mrocklin:
Thank you for raising this.  The bad link was my fault.  This happened when
we were moving things from dask.pydata.org to docs.dask.org.

I've gone through and made a few other changes and pushed to master.  Thank
you for raising this important issue.

PR #167, opened 2018-10-05
Add ability to override python used to start workers

### LINKED ISSUE #166: Make python command configurable
Using `sys.executable` in https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py#L238 seems unnecessarily strict. My use case is that I have KNL nodes available and I want to manage both KNL and standard nodes from dask_jobqueue. I've mirrored the software stacks, and all the necessary `module load`s are the same, but the absolute paths differ.

I'm happy to issue a PR but I wanted to know how you would consider handling it? In my case, I'm installing  `dask_jobqueue` as a module for our users so I know that `os.path.basename(sys.executable)` (for example) would work as intended. I could add a new keyword arg that sets `python` which defaults to the existing value?

ocaisa:
fixes #166

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -167,7 +167,7 @@ class JobQueueCluster(Cluster):
                  env_extra=None,
                  walltime=None,
                  threads=None,
-                 python=None,
+                 python=sys.executable,
                  **kwargs
                  ):
         """ """
@@ -198,8 +198,6 @@ class JobQueueCluster(Cluster):
             extra = dask.config.get('jobqueue.%s.extra' % self.scheduler_name)
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
-        if python is None:
-            python = sys.executable
 
         if dask.config.get('jobqueue.%s.threads', None):
             warnings.warn(threads_deprecation_message)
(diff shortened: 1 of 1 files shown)


dask_jobqueue/core.py
```diff
@@ -195,6 +198,8 @@ def __init__(self,
             extra = dask.config.get('jobqueue.%s.extra' % self.scheduler_name)
         if env_extra is None:
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
+        if python is None:
+            python = sys.executable
```
mrocklin: I think that we can safely include this directly in the signature.  It's not going to change.

dask_jobqueue/core.py
```diff
@@ -114,17 +114,19 @@ class JobQueueCluster(Cluster):
...
     Attributes
     ----------
     submit_command: str
         Abstract attribute for job scheduler submit command,
-        should be overriden
+        should be overridden
     cancel_command: str
         Abstract attribute for job scheduler cancel command,
-        should be overriden
+        should be overridden
```
mrocklin: Thanks for correcting these typos

mrocklin:
One small comment.  Otherwise this looks good to me.  Thanks @ocaisa

dask_jobqueue/core.py
guillaumeeb: So this means no config entry?

mrocklin: Right.  If we were to include this as part of the config (which I don't think is necessary) then we would want to check the default every time we called the constructor here, so we would need this check to be within the constuctor as you have it here.  However if the default value is just going to be `sys.executable` (which I think is best today) then we can include it as a default directly in the signature of the constructor.  This is a bit better both because it reduces code, and because it makes the default visible if people look at the docstring.

PR #173, opened 2018-10-11
Documentation reorg

### LINKED ISSUE #159: Reorganize online documentation
I think that the online documentation we provide needs some refactoring:

- Split the index landing page that is too big into smaller content.
- Change the toctree, add some captions, maybe the classical _Getting started_, _Configuration_, _Develp_
- Optionnaly use a standard `source` subdirectory instead of all files in the doc folder ?

guillaumeeb:
Fixes #159 

Still in progress, just to say I'm working on it.


guillaumeeb:
Ready for reviews. Some things that can be discussed:

- There are two configurations pages: configuration.rst, which is a subset of configuration-setup.rst one, much more detailed. We could just drop the first one.
- The toctree layout: howitworks could be in Help section for example. But I just feel it was important for people to understand what they are doing.

ping @jhamman @mrocklin @lesteve

guillaumeeb:
As we are only talking about documentation here, I think I will merge this in soon.

guillaumeeb:
For sure ! There is no hurry, I like to keep things moving, so I can look impatient.

mrocklin:
@guillaumeeb I pushed a few changes.  Let me know if you're ok with them.

Thanks generally for doing this.

guillaumeeb:
Thanks @mrocklin, I'm OK with your changes!

For dask-mpi vs dask-jobqueue, we should may be do it upstream in http://docs.dask.org/en/latest/setup/hpc.html.

Issue #172, opened 2018-10-08
LSFCluster worker doesn't execute all threads on the same node


louisabraham:
I am not 100% sure, but it seems to me that nothing forces a LSF task to execute all the threads on the same node.

For example, PBSCluster uses the option `ncpus:n` that requests cpus per node, and SLURMCluster specifies `-n 1`and uses `--cpus-per-task=n` to allocate n cpus on each host.

However, LSFCluster uses the `-n` option. I think that without a `--span[hosts=1]` option LSF can use processors from different hosts. See the relevant documentation [there](https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.2/lsf_admin/span_string.html).

Do I ignore some internals of LSF that make this acceptable?

guillaumeeb:
I think no core maintainer of this project uses LSF. 

But looking at the PR which introduced this was indeed identified: https://github.com/dask/dask-jobqueue/pull/78#issuecomment-400996012. But one way or another, it did not get implemented.

It would be nice to submit a PR to fix this!

louisabraham:
Ok, I'll do it this week.

guillaumeeb:
@louisabraham I'd like to make a small release by the end of the week before starting to work on bigger changes.

I'd like to take this fix in.
Would you be able to fix this or would you prefer someone else to do it ?

louisabraham:
I'll do it today.

My editor autoformats the code, and it seems the different files don't use the same formatter.

Is black like dask_ml OK?

louisabraham:
Also, currently `ncpus` is used but I think it should be `cores`.

I would need some explanation about the difference between `cores` and `processes`.

guillaumeeb:
We don't have code format template or requirement yet, except flake8 checks.

I don't know black formatter, but should be ok.

For cores vs ncpus, could you point to the corresponding lines in the code?

louisabraham:
Cores is not used in `LSFCluster`, but required as an instance of `JobQueueCluster`.

Other arguments are used like this:

https://github.com/dask/dask-jobqueue/blob/b510bb19bce24ededfa9ae73809949bfc830a7e4/dask_jobqueue/core.py#L231-L233

PBSCluster uses them like this:

https://github.com/dask/dask-jobqueue/blob/b510bb19bce24ededfa9ae73809949bfc830a7e4/dask_jobqueue/pbs.py#L84-L89

And LSFCluster:

[...]

PR #176, opened 2018-10-17
add ncpus_per_host option to LSFCluster

louisabraham:
fix https://github.com/dask/dask-jobqueue/issues/172

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -484,7 +484,7 @@ class JobQueueCluster(Cluster):
         job_id = match.groupdict().get('job_id')
         if job_id is None:
             msg = ("You need to use a 'job_id' named group in your regexp, e.g. "
-                   "r'(?P<job_id>\d+)', in your regexp. Your regexp was: "
+                   "r'(?P<job_id>\\d+)', in your regexp. Your regexp was: "
                    "{!r}".format(self.job_id_regexp))
             raise ValueError(msg)
 
--- a/dask_jobqueue/lsf.py
+++ b/dask_jobqueue/lsf.py
@@ -23,8 +23,11 @@ class LSFCluster(JobQueueCluster):
         `#BSUB -P` option.
     ncpus : int
         Number of cpus. Passed to `#BSUB -n` option.
+    ncpus_per_host : int
+        Indicates the number of processors on each host that should be allocated to the job.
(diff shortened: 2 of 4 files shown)


louisabraham:
@willirath the reformatting is done in separate commits, I don't understand what you mean.

willirath:
Sorry, I was not as verbose as I should have been: All these commits will be squashed to one commit representing this PR. Then, we'll have `git blame` telling the user that the reformatted lines have been changed in order to fix the LSFCluster issue.

louisabraham:
I'm am no expert in github, but I think it only happens if you actually squash.

Since I forked directly from the current HEAD, you can actually do [Rebase and merge](https://help.github.com/articles/about-pull-request-merges/#rebase-and-merge-your-pull-request-commits).

If you prefer, I can cherry-pick the commits and turn off my editor's autoformatting.

guillaumeeb:
Hm, I think @willirath has got a point. It will be clearer to avoid reformatting for this issue.
The rule here is to only use squash and commit, so we will lose the detail commit history. If you could only implement the changes necessary to fix the identified issue this would be better.

louisabraham:
Ok, I'll redo it. Anyway I think I'll need to change my code depending on your answer on #172.

Just for my personal knowledge: what is the point of using only "squash and commit"?

willirath:
> Just for my personal knowledge: what is the point of using only "squash and commit"?

The master branch evolving in steps that make sense from a functional point of view rather than reflecting the (often random) steps it took to develop and implement the solution.

Taking this PR as an example, we'd have one commit telling the reader "LSFCluster now correctly starts all threads on the same node and here's all we did to achieve that." vs different commits that make it very hard to later understand what happened.

guillaumeeb:
Thanks anyway @louisabraham for all the work you've already done, and for sticking with this! 
Sorry I was not really available today to answer your questions.

louisabraham:
I am the one who thanks both of you for actually taking more time to explain things to me than it would take you to fix the bug :)

willirath:
> I am the one who thanks both of you for actually taking more time to explain things to me than it would take you to fix the bug :)

I'm glad you take it this way! Communicating criticism in a well-meaning and constructive manner is not trivial and it'd probably be easy to read my remarks in an offensive tone.

PR #177, opened 2018-10-17
require all cores on the same node

### LINKED ISSUE #172: LSFCluster worker doesn't execute all threads on the same node

louisabraham:
see https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.2/lsf_admin/span_string.html

fixes #172 and replaces #176

### DIFF
--- a/dask_jobqueue/lsf.py
+++ b/dask_jobqueue/lsf.py
@@ -88,6 +88,10 @@ class LSFCluster(JobQueueCluster):
             logger.info("ncpus specification for LSF not set, initializing it to %s" % ncpus)
         if ncpus is not None:
             header_lines.append('#BSUB -n %s' % ncpus)
+            if ncpus > 1:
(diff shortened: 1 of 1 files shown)


guillaumeeb:
Thanks @willirath for your involvement too!

PR #183, opened 2018-10-30
Take into account grouped workers scale using ClusterManager

### LINKED ISSUE #152: Use the grouped workers `worker_key` function in scale outside of adaptive
Seen thanks to https://github.com/dask/dask-kubernetes/issues/98.

Currently, `_job_id_from_worker_name` function is used as worker_key in adaptive options: https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py#L145.

However, this is not in place when manually scaling down or up. This could be a nice thing to avoid stopping too many workers when scaling down.

This may have to be changed upstream in distributed, somewhere here: https://github.com/dask/distributed/blob/master/distributed/deploy/cluster.py#L103. We could use a `self.worker_key method` if defined as another kwarg.

@mrocklin any opinion on that?

guillaumeeb:
Fixes #152, supersedes #153.


guillaumeeb:
In order to go deeper in ClusterManager design, I will probably merge this soon (tomorrow at the latest).
If anybody has some remarks, these could be done in independent issues if this PR is merged.

PR #189, opened 2018-11-05
Add shebang option to support more shells

danpf:
Previously the shell(shebang) was hardcoded to be '#!/bin/bash'.
This fixes that, because some of the queueing schedulers had their own
shebangs hardcoded as well, so you actually ended up having two shebang
lines.  In order to accomodate more shells, we can
add an option to specify shell via shebangs.  If you think this
implementation is fine, I can/will add this to the documentation as
well.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -120,6 +120,8 @@ class JobQueueCluster(ClusterManager):
         Other commands to add to script before launching worker.
     python : str
         Python executable used to launch Dask workers.
+    shebang : str
+        Path to desired interpreter for your batch submission script.
     kwargs : dict
         Additional keyword arguments to pass to `LocalCluster`
 
@@ -218,6 +220,10 @@ class JobQueueCluster(ClusterManager):
 
         if memory is None:
             raise ValueError("You must specify how much memory to use per job like ``memory='24 GB'``")
+
+        if shebang is None:
+            raise ValueError("You must specify a batch script interpreter to use"
+                             " like ``#!/usr/bin/env bash``")
 
         # This attribute should be overridden
         self.job_header = None
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -143,7 +143,7 @@ class JobQueueCluster(ClusterManager):
     """
 
     _script_template = """
-#!/bin/bash
+%(shebang)s
 
 %(job_header)s
 
@@ -171,6 +171,7 @@ class JobQueueCluster(ClusterManager):
                  log_directory=None,
                  walltime=None,
                  threads=None,
+                 shebang=None,
                  python=sys.executable,
                  **kwargs
                  ):
@@ -206,6 +207,8 @@ class JobQueueCluster(ClusterManager):
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
         if log_directory is None:
             log_directory = dask.config.get('jobqueue.%s.log-directory' % self.scheduler_name)
+        if shebang is None:
+            shebang = dask.config.get('jobqueue.%s.shebang' % self.scheduler_name)
 
         if dask.config.get('jobqueue.%s.threads', None):
             warnings.warn(threads_deprecation_message)
@@ -241,6 +244,8 @@ class JobQueueCluster(ClusterManager):
         self.local_cluster.scheduler.add_plugin(self._scheduler_plugin)
 
         self._adaptive = None
+
+        self.shebang = shebang
 
         self._env_header = '\n'.join(env_extra)
 
@@ -313,7 +318,8 @@ class JobQueueCluster(ClusterManager):
 
     def job_script(self):
         """ Construct a job submission script """
-        pieces = {'job_header': self.job_header,
+        pieces = {'shebang': self.shebang,
+                  'job_header': self.job_header,
                   'env_header': self._env_header,
                   'worker_command': self._command_template}
         return self._script_template % pieces
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -221,10 +221,6 @@ class JobQueueCluster(ClusterManager):
         if memory is None:
             raise ValueError("You must specify how much memory to use per job like ``memory='24 GB'``")
 
-        if shebang is None:
-            raise ValueError("You must specify a batch script interpreter to use"
-                             " like ``#!/usr/bin/env bash``")
-
         # This attribute should be overridden
         self.job_header = None
 
--- a/dask_jobqueue/jobqueue.yaml
+++ b/dask_jobqueue/jobqueue.yaml
@@ -12,7 +12,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # OAR resource manager options
-    shebang: "#!/bin/bash"
+    shebang: "#!/usr/bin/env bash"
     queue: null
     project: null
     walltime: '00:30:00'
@@ -35,7 +35,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # PBS resource manager options
-    shebang: "#!/bin/bash"
+    shebang: "#!/usr/bin/env bash"
     queue: null
     project: null
     walltime: '00:30:00'
@@ -58,7 +58,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # SGE resource manager options
-    shebang: "#!/bin/bash"
+    shebang: "#!/usr/bin/env bash"
     queue: null
     project: null
     walltime: '00:30:00'
@@ -81,7 +81,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # SLURM resource manager options
-    shebang: "#!/bin/bash"
+    shebang: "#!/usr/bin/env bash"
     queue: null
     project: null
     walltime: '00:30:00'
@@ -105,7 +105,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # PBS resource manager options
-    shebang: "#!/bin/bash"
+    shebang: "#!/usr/bin/env bash"
     queue: null
     project: null
     walltime: '00:30:00'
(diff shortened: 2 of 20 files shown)


dask_jobqueue/core.py
```diff
@@ -206,6 +207,8 @@ def __init__(self,
             env_extra = dask.config.get('jobqueue.%s.env-extra' % self.scheduler_name)
         if log_directory is None:
             log_directory = dask.config.get('jobqueue.%s.log-directory' % self.scheduler_name)
+        if shebang is None:
```
willirath: I'd reconsider adding a kwarg for this.  In my understanding, kwargs are for things you change frequently.  This looks like an issue that you have to sort out once for your deployment and then never touch it again.  Just adding it to the config should cover all cases.

dask_jobqueue/jobqueue.yaml
```diff
@@ -34,6 +35,7 @@ jobqueue:
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
 
     # PBS resource manager options
+    shebang: "#!/bin/bash"
```
willirath: Shouldn't we go for `#!/usr/bin/env bash` ?

dask_jobqueue/tests/test_slurm.py
```diff
@@ -48,6 +48,7 @@ def test_job_script():
     with SLURMCluster(walltime='00:02:00', processes=4, cores=8, memory='28GB') as cluster:
 
         job_script = cluster.job_script()
+        assert job_script.startswith("#!/bin/bash")
```
willirath: This should also go into the tests for the other schedulers.

willirath:
Looks good!

> Off Topic:
>
> Looking at this PR, adding all these identical lines to the configs (e.g., <https://github.com/dask/dask-jobqueue/pull/189/files#diff-06f4938c65f61f791782337bf4187054R15>) feels wrong.

dask_jobqueue/tests/test_slurm.py
guillaumeeb: I believe this should only go in https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/tests/test_jobqueue_core.py as this is not specific to any scheduler.

dask_jobqueue/jobqueue.yaml
guillaumeeb: Yep, looks like it, good catch!

dask_jobqueue/core.py
guillaumeeb: I'm not sure, and I have no strong feeling on this. Currently it looks like we have kind of a match between kwargs and config file. I'm not sure which way we should go.

guillaumeeb:
And thanks @willirath for the review here, this is really appreciated.

dask_jobqueue/core.py
danpf: I did it this way only because that's how everything else in the yaml's are set.  I don't really mind either way.  maybe let me know @guillaumeeb ?

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -42,6 +42,58 @@ def test_command_template():
         assert ' --preload mymodule' in cluster._command_template
 
 
+def test_shebang_settings():
```
guillaumeeb: You should use parametrized test as the one below for all this!

dask_jobqueue/tests/test_slurm.py
```diff
@@ -66,7 +66,7 @@ def test_job_script():
     with SLURMCluster(walltime='00:02:00', processes=4, cores=8, memory='28GB',
                       env_extra=['export LANG="en_US.utf8"',
                                  'export LANGUAGE="en_US.utf8"',
-                                 'export LC_ALL="en_US.utf8"']
+                                 'export LC_ALL="en_US.utf8"'],
```
guillaumeeb: A comma is remaining here.

dask_jobqueue/core.py
guillaumeeb: Yeah, that is what I thought. If so, leave it as is, as the everything else is done. Maybe we'll need to check whether all the kwargs are justified or not in the future.

dask_jobqueue/tests/test_jobqueue_core.py
danpf: really cool!!!!

guillaumeeb:
LGTM, @willirath do you see something else?

danpf:
sorry to bother you more, please review the updates to the documentation I added as well.
Also added some functionality to make sure there was always a shebang + test for that.

willirath:
Apart from the question about the added dependency on `mock`, this looks very good!

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -42,6 +43,24 @@ def test_command_template():
...
+    with Cluster(cores=2, memory='4GB', shebang=python_shebang) as cluster:
+        job_script = cluster.job_script()
+        assert job_script.startswith(python_shebang)
+        assert 'bash' not in job_script
+    with Cluster(cores=2, memory='4GB') as cluster:
+        job_script = cluster.job_script()
+        assert job_script.startswith(default_shebang)
+
+    with pytest.raises(ValueError, match="batch script interpreter"):
+        with (dask.config.set(shebang=None)):
```
willirath: Sorry.  My example code was wrong.  `dask.config.set('jobqueue.%s.shebang' % Cluster.scheduler_name, None)` should work.

danpf: thanks, my initial idea was way worse.

willirath:
> LGTM, @willirath do you see something else?

@guillaumeeb Good to go from py point of view!

docs/source/configuration-setup.rst
```diff
@@ -8,7 +8,7 @@ You provide this information either as keyword arguments to the constructor:
 
 .. code-block:: python
 
-   cluster = PBSCluster(cores=36, memory='100GB', queue='regular', ...)
+   cluster = PBSCluster(cores=36, memory='100GB', shebang,"#!/usr/bin/env bash", queue='regular', ...)
```
guillaumeeb: Is there a comma instead of a equal sign here?
At the end, I would prefer to avoid exposing kwargs that are more rarely used everywhere in the docs. In this particular example I feel this is not needed.

dask_jobqueue/core.py
```diff
@@ -216,6 +221,10 @@ def __init__(self,
         if memory is None:
             raise ValueError("You must specify how much memory to use per job like ``memory='24 GB'``")
 
+        if shebang is None:
+            raise ValueError("You must specify a batch script interpreter to use"
+                             " like ``#!/usr/bin/env bash``")
```
guillaumeeb: Is this really a problem if there is no shebang? 
What concerns me here is that this exception might break existing jobqueue configurations, where no shebang is declared. From my point of view it shouldn't.

docs/source/configurations.rst
```diff
@@ -103,6 +105,7 @@ ARM Stratus
         name: dask-worker
         cores: 36
         memory: 270GB
+        shebang: '#!/usr/bin/env bash'
```
guillaumeeb: You should remove the shebang from "real life" configurations, as keeping the default shebang is perfectly fine apparently for those. This include ARM, but also above Cheyenne and Nersc.

docs/source/examples.rst
```diff
@@ -23,6 +24,7 @@ PBS Deployments
 
    cluster = PBSCluster(processes=18,
                         threads=4,
+                        shebang='#!/usr/bin/env bash',
```
guillaumeeb: Still the same, shebang is optional, maybe put it once, but not everywhere.

docs/source/configuration-setup.rst
danpf:
```suggestion
   cluster = PBSCluster(cores=36, memory='100GB', shebang="#!/usr/bin/env bash", queue='regular', ...)
```
great catch!

dask_jobqueue/core.py
danpf: We could maybe add a deprecation warning? and then a warning to say that if shebang is None everywhere, automatically set it to `#!/usr/bin/env bash`  to preserve the old configuration

You are right that this breaks existing configs in its current state.  But you cannot submit a batch script to most schedulers without a shebang:
```
sbatch: error: This does not look like a batch script.  The first
sbatch: error: line must start with #! followed by the path to an interpreter.
sbatch: error: For instance: #!/bin/sh
```

docs/source/configurations.rst
danpf: Are these what the DOE are using as their `jobqueue.yaml` ? in that case they 'should' set the shebang k-v in this updated version.  Currently the config is set per-scheduler and the shebang is pulled from the scheduler's k-v config. This is solved if we use the deprecation method I suggested.  Although I think it's best practice for the documentation to show these as set so people understand.

dask_jobqueue/core.py
guillaumeeb: Wait, I may be wrong about this. I need to try this PR to see if it breaks my config. Maybe it doesn't, default value should be taken from the module default jobqueue.yaml file.

docs/source/configurations.rst
guillaumeeb: I think they shouldn't if they keep the default value, see my response to the other comment. I will try this tomorrow. Default shebang value should be inherited from default config if not specified.

dask_jobqueue/core.py
guillaumeeb: Sorry did not find time today, and this will not be possible until friday. You could verify it though, just create a jobqueue.yaml config file, and do not add a shebang line. 
I believe the only way to reach that code is to create a local config file, and set shebang to null in it. I'm thus in favor of removing this code, as it seems very unlikely a user run in this case without knowing it.

docs/source/configurations.rst
guillaumeeb: And so based on the above discussion, I think we really should remove the majority of examples containing `shebang` in the doc, only keeping a few specific. Maybe add the config you are using in your cluster in the example section?

danpf: I deleted a bunch, I only left it in places where the docs describe the extra args as well!!

guillaumeeb:
You've got flake8 checks failing here.

willirath:
Looks like this is all fine now?

guillaumeeb:
Yes ! Thanks to both of you here, merging.

PR #192, opened 2018-11-10
allow caller to overwrite diagnostics server settings

twoertwein:
Some (academic) clusters might be directly exposed to the internet (and might have poor firewall settings). This lets them allow to control the `diagnostics_port` setting (e.g., bind it only to a local network or disable it).

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -231,9 +231,8 @@ class JobQueueCluster(ClusterManager):
             kwargs.setdefault('ip', '')
 
         # Bokeh diagnostics server should listen on all interfaces
-        diagnostics_ip_and_port = ('', 8787)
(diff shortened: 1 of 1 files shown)


twoertwein:
thanks, I will change the `setdefault` to a `pop` with a default value.

The documentation of LocalCluster mentions that `diagnostics_port` can be IP and port (https://github.com/dask/distributed/blob/32341216c9f62e37c2e01c898e26b117e2b872b3/distributed/deploy/local.py#L48) even though the name doesn't imply that.

guillaumeeb:
> The documentation of LocalCluster mentions that diagnostics_port can be IP and port 

Then sorry, I think I prefer the `setdefault` approach, could you revert back to the previous commit situation?

twoertwein:
done

Issue #195, opened 2018-11-14
Resource specification on GridEngine Clusters


ericmjl:
Hi Dask team!

My colleague @sntgluca and I have been very enthusiastic about the possibilities enabled by `dask-jobqueue` at NIBR! It's been a very productivity-enhancing tool for me. At the same time, we found something that we think might be a be a bug, but would like to disprove/confirm this before potentially working on a PR to fix it.

Firstly, we found that when using the `memory` keyword argument, the `SGECluster` will show that  `X`GB per worker node is allocated. However, the amount of RAM that is allocated, according to the queueing status screen, is only the default amount specified by the sysadmins. 

Here is some evidence that I collected from the logs on our machines.

Firstly, Dask's worker logs show that 8GB is allocated to them:

```
# Ignore this line, it is here to show the job ID.
/path/to/job_scripts/16524465: line 10: /path/to/activate: Permission denied
distributed.worker - INFO -               Threads:                          1
distributed.worker - INFO -                Memory:                    8.00 GB
```

However,  for the same job ID, when I used `qstat -j JOBID`:

```
==============================================================
job_number:                 16524465
...
sge_o_path:                 hard resource_list:         m_mem_free=4G,h_rt=259200,slot_limitA=1
...
granted_req.          1:    m_mem_free=4.000G, slot_limitA=1
```

As you can see, the resources granted were only 4GB of memory, not the 8GB requested, but the Dask worker logs show 8GB being allocated.

I have a hunch that this is a bug, but both @sntgluca have this idea that our end users shouldn't have to worry about GridEngine resource spec strings, and should be able to use the very nice `SGECluster` API to set these parameters correctly. Looking at the `SGECluster` source code, it looks doable with a small-ish PR to parse the `memory` (and other kwargs) into the correct resource specification string, if this is something that you would be open to. 

Please let us know!

ericmjl:
Thanks @jhamman! This is what the job script is:

```
#!/bin/bash

#!/usr/bin/env bash
#$ -N dask-worker
#$ -q default.q
#$ -l h_rt=259200
#$ -cwd
#$ -j y

activate mpnn

/home/maer3/anaconda/envs/mpnn/bin/python -m distributed.cli.dask_worker tcp://10.145.61.208:39018 --nthreads 1 --memory-limit 8.00GB --name dask-worker--${JOB_ID}-- --death-timeout 60
```

And this is how I call the SGECluster:

```python
cluster = SGECluster(queue='default.q',
[...]

PR #197, opened 2018-11-16
added a memory string parser

ericmjl:
cc: @jhamman, @sntgluca, @guillaumeeb 

I am proceeding with very small steps with this PR. This PR is a very focused one, and only adds a memory string parser, such that an end-user can specify the amount of RAM that they would like per worker node.

I have done the following:

- Added a memory string parser
- Updated the docstring to state that memory is a kwarg that can be specified in lieu of the resource string.

If I remember correctly, I can use multiple lines to specify resources. Hence:

```
#$ -l m_mem_free=8G
#$ -l h_rt=1500000
``` 

should be equivalent to:

```
#$ -l m_mem_free=8G,h_rt=1500000
```

@sntgluca, can you confirm/disprove my memory on this?

Still wondering about the following:

- Should I add a test for the memory string parser? If so, where is the best place to put it?
- Should I append to the example shown in the SGECluster docstring?

If merged, then this PR resolves in part #195, but not fully.

### DIFF
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -16,6 +16,9 @@ class SGECluster(JobQueueCluster):
     ----------
     queue : str
         Destination queue for each worker job. Passed to `#$ -q` option.
+    memory: str
+        The amount of memory requested per worker job. Passed to `#$ -l` option. Can be
+        specified in lieu of `resource_spec`.
     project : str
         Accounting string associated with each worker job. Passed to `#$ -A` option.
     resource_spec : str
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -62,8 +62,16 @@ class SGECluster(JobQueueCluster):
             header_lines.append('#$ -q %(queue)s')
         if project is not None:
             header_lines.append('#$ -P %(project)s')
+
+        # Define resource specifications. There are only two cases allowed in this
+        # implementation: either specify using resource_spec, or specify using the rest of the
+        # kwargs.
         if resource_spec is not None:
             header_lines.append('#$ -l %(resource_spec)s')
+        else:
+            header_lines.append('#$ -l m_mem_free=%s' % parse_memory(kwargs.get('memory')))
+            logger.info("Resource specification not set for SGE, initializing to %s")
+
         if walltime is not None:
             header_lines.append('#$ -l h_rt=%(walltime)s')
         if self.log_directory is not None:
@@ -82,3 +90,31 @@ class SGECluster(JobQueueCluster):
         self.job_header = header_template % config
 
         logger.debug("Job script: \n %s" % self.job_script())
+
+
+def parse_memory(memory: str):
+    """
+    Parses the memory string to format it correctly for SGE job scripts.
+
+    Simply uses a bunch of heuristics to parse the string.
+
+    :param memory: The amount of RAM requested per worker node.
+    :returns: memory, properly formatted
+    """
+    # Covering only the most common use cases here. The pattern should be obvious if we need to
+    # add more cases below.
+    mappings = {
+        'tb': 'T',
+        't': 'T',
+        'gb': 'G',
+        'g': 'G',
+        'mb': 'M',
+        'm': 'M',
+        'kb': 'K',
+        'k': 'K'
+    }
+    for ending, replacement in mappings.items():
+        if memory.lower().endswith(ending):
+            memory = memory.replace(ending, replacement)
+
+    return memory
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -16,9 +16,6 @@ class SGECluster(JobQueueCluster):
     ----------
     queue : str
         Destination queue for each worker job. Passed to `#$ -q` option.
-    memory: str
-        The amount of memory requested per worker job. Passed to `#$ -l` option. Can be
-        specified in lieu of `resource_spec`.
     project : str
         Accounting string associated with each worker job. Passed to `#$ -A` option.
     resource_spec : str
@@ -66,9 +63,6 @@ class SGECluster(JobQueueCluster):
         if project is not None:
             header_lines.append('#$ -P %(project)s')
 
-        # Define resource specifications. There are only two cases allowed in this
-        # implementation: either specify using resource_spec, or specify using the rest of the
-        # kwargs.
         if resource_spec is not None:
             header_lines.append('#$ -l %(resource_spec)s')
         else:
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ log
 *.swo
 .cache/
 .pytest_cache
+.mypy_cache/


dask_jobqueue/sge.py
```diff
@@ -16,6 +16,9 @@ class SGECluster(JobQueueCluster):
     ----------
     queue : str
         Destination queue for each worker job. Passed to `#$ -q` option.
+    memory: str
+        The amount of memory requested per worker job. Passed to `#$ -l` option. Can be
+        specified in lieu of `resource_spec`.
```
guillaumeeb: Not sure you should override default docstring here. The addition you mademoiselle should be implicit, even if it is not the case currently.

dask_jobqueue/sge.py
```diff
@@ -62,8 +65,16 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
             header_lines.append('#$ -q %(queue)s')
         if project is not None:
             header_lines.append('#$ -P %(project)s')
+
+        # Define resource specifications. There are only two cases allowed in this
+        # implementation: either specify using resource_spec, or specify using the rest of the
+        # kwargs.
```
guillaumeeb: @lesteve would say that this comment is useless, the code is explicit :)

dask_jobqueue/sge.py
```diff
@@ -62,8 +65,16 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
...
         if resource_spec is not None:
             header_lines.append('#$ -l %(resource_spec)s')
+        else:
+            header_lines.append('#$ -l m_mem_free=%s' % parse_memory(kwargs.get('memory')))
```
guillaumeeb: I'm wondering if you could add CPU resources spec too, even if this is not the original problem, shouldn't be too hard or introduce much noise.

dask_jobqueue/sge.py
```diff
@@ -82,3 +93,31 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
         self.job_header = header_template % config
 
         logger.debug("Job script: \n %s" % self.job_script())
+
+
+def parse_memory(memory: str):
```
guillaumeeb: I think the input to this fonction should be worker_memory, that we optain using distributed.utils.parse_bytes. Memory kwarg can be a number with no explicit unit if I'm not mistaken. You should probably use format_memory from the same package, and base your conversation from its output.

guillaumeeb:
As for your questions, it would definitly be valuable to have a test, at least checking that the job_script contains the correct resource_spec even without specifying it.

For the docstring as you wish, this does not look mandatory to me.

dask_jobqueue/sge.py
ericmjl: Removing!

dask_jobqueue/sge.py
ericmjl: Ok! Will take this out.

dask_jobqueue/sge.py
ericmjl: I'd like to hold off on that, but only for this PR. Happy to do it in a separate PR, so that we can settle issues independently.

dask_jobqueue/sge.py
ericmjl: Got it!

ericmjl: @guillaumeeb I did a search of dask/distributed for `format_memory`, but could not find it. Here's the URL to the search: https://github.com/dask/distributed/search?utf8=✓&q=format_memory&type=

Can you point to where the function `format_memory` is located?

mrocklin: https://github.com/dask/distributed/blob/964a917c45e8c11b16086cf05949d3529cc0986a/distributed/utils.py#L1129-L1168

ericmjl: Thanks @mrocklin! Just to confirm, then, @guillaumeeb, you were referring to `parse_bytes` when you wrote `format_memory`? Having found `parse_bytes`, I think I know how to move forward from here.

dask_jobqueue/sge.py
guillaumeeb: Ok, fine by me.

dask_jobqueue/sge.py
guillaumeeb: You've got `format_bytes` just above, sorry the function name I used was wrong: https://github.com/dask/distributed/blob/964a917c45e8c11b16086cf05949d3529cc0986a/distributed/utils.py#L1081-L1107.

The only thing is that format_bytes uses "real" Giga/Mega/Kilobytes and note Gibi/Mebi/Kibibytes (see https://en.wikipedia.org/wiki/Gibibyte). For example, PBS uses the GB term, but its meaning is GiB. See https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/pbs.py#L106-L128.  But I actually don't really like this part of the code, and since GiB>GB, I think it's fine to use Dask's GB as GiB for PBS/SGE, without doing a real conversion.

dask_jobqueue/sge.py
ericmjl: @guillaumeeb, actually, I think the default docstring should be overridden, so as to make clear what memory is being specified.

As I learned today, there are multiple memory specification possibilities for grid engine. Two that I currently know of are `m_mem_free` and `h_vmem`. In this PR, I am forcing the `memory` keyword argument to be passed into `m_mem_free`, and I think this should be made explicit for `SGECluster` users. 

As far as I can tell, `m_mem_free` is the most commonly used option for memory specification in job scripts here at work. `h_vem` is less commonly used. I expect that `m_mem_free` should be the correct default option for 95% of `SGECluster` use cases.

The Zen of Python states that "explicit is better than implicit", and I think it'd be better to state explicitly in the docstring that memory is passed to `#$ -l m_mem_free`, so that the end-user knows that if they need to specify `h_vmem` instead/on top the requested worker RAM, they still can do so with the `resource_spec` keyword argument as well.

guillaumeeb: I'm just not sure of what will append to the consolidated docstring, as memory is already defined in `JobQueueCluster` class in core.py:
````
    memory: str
        Total amount of memory per job
````

You may end up with two different memory entries in the docstring. This is worth testing, or maybe look at docrep module to see if it provide some tools for this case.

guillaumeeb:
Broken CI seems unrelated to your PR, see #201.

dask_jobqueue/sge.py
ericmjl: Understood. Thank you for pointing that out! I will ensure that the docs are clear and unambiguous on this pointer.

guillaumeeb:
CI problem is corrected on master, you can merge it to have correct Travis inputs.

ericmjl:
@guillaumeeb, it appears that my use of mypy-style type annotations is breaking tests. Happy to remove it, though can I just check, is mypy being used in dask-jobqueue?

guillaumeeb:
> is mypy being used in dask-jobqueue?

Nope, never used it. From what I've seen in Dask related projects, we usually just use docstrings to precise args type, or don't precise it at all.

dask_jobqueue/sge.py
lesteve: `m_mem_free` does not exist on my SGE cluster. Quickly googling this may only exist for Univa Grid Engine.
On my cluster there is `mem_req` that could be used. Does that exist on yours @ericmjl?

lesteve: It does not look like `mem_req` is standard at all ... not sure what to do.

ericmjl: @lesteve I am starting to discover this as well. It appears that mapping the memory keyword argument to any particular configuration string is not going to be productive.

I am beginning to think I should close this PR and open a different PR for documentation purposes.

guillaumeeb: Wow, that's good to know! It there is no memory option standard, then yes, no other choice to keep with `resources_spec` and document this.

guillaumeeb:
@ericmjl OK to close in favor of #220?

ericmjl:
@guillaumeeb yes, closing right now!

PR #200, opened 2018-11-20
Typo in config fie for slurm

ocaisa:
I'm guessing this was a typo since it doesn't match the other resource managers


guillaumeeb:
Thanks @ocaisa !

Issue #8, opened 2018-03-02
Unified test suite


mrocklin:
After we figure out CI testing in #2 we might consider refactoring the test suite into a unified `TestCase` class for common functionality.

jakirkham:
So are you wanting to use `unittest` or are you only thinking the organization structure of a class would be nice?

mrocklin:
No strong preference.  What do you think is best?  I would probably default to `unittest` just because that's what I've done before.  That's not a very good reason on it's own though :)

jakirkham:
Sure, using `unittest` is fine. Thought about `pytest` too. Though don't have any better arguments for it. :)

mrocklin:
Given that we will be using pytest we can probably at least avoid the
self.assertFoo methods :)

guillaumeeb:
Do we need to do something here? Or is the current configuration good enough? I admit I don't quite understand what is being suggested by @mrocklin at first.

mrocklin:
My original suggestion was to create a `unittest.TestCase` subclass that contains a set of tests, but that needs a bit of specialization to each cluster type.

```python
class TestBasic(unittest.TestCase):
    Cluster = None
    def test_cores(self):
        with self.Cluster(cores=4, processes=2) as cluster:
            cluster.scale(2)
            ... wait a bit ...
            with Client(cluster) as client:
                info = client.scheduler_info()
                assert len(info['workers']) == 4
                ... more assertions
    
    def test_header(self):
        ...

    def test_foo(self):
        ...

class TestPBSCluster(TestBasic):
    Cluster = PBSCluster
```

mrocklin:
I'm not sure how much reuse there is though.  This might or might not make sense.

PR #204, opened 2018-11-26
This commit adds support for a configuration parameter

josephhardinee:
This commit adds support for a configuration parameter to each of the cluster types in jobqueue. It does this by adding a configuration parameter that is used to override the scheduler_name class variable.
 I also demoted scheduler_name to an instance variable as you could now theoretically have multiple types of PBS clusters.

The form of usage of this is as suggested by @mrocklin. 

```python
cluster = PBSCluster(configuration='my_pbs') 
```
Will draw from the my_pbs dictionary in the yaml file rather than the normal pbs one. When called without a parameter, it still defaults to pbs. I made a similar change for all of the cluster type initiators.

I tested it on the pytests, as well as the docker tests for PBS. All passed except one at the end that I think is just a docker error with weird networking setups. 

Let me know if this looks acceptable or if you'd like any other changes.


mrocklin:
Thanks for doing this @josephhardinee !

Some questions/comments:

1.  Should we drop the `scheduler_name` attribute entirely?  Do we use it for anything else?  (This is more a question for other maintainers)
2.  Can we include a small test for this?  It might look like the following:

    ```python
    def test_configuration_name():
        conf = {'queue': 'foo'}        
        with dask.config.set({'jobqueue.my_custom_configuration': conf}):
            cluster = SGECluster(configuration='my_custom_configuration')
            assert cluster.queue == 'foo'
    ```
3.  Is there a better name than `configuration`?  This might be confused with all of the key-value pairs in the configuration instead.

dask_jobqueue/lsf.py
```diff
@@ -50,9 +50,10 @@ class LSFCluster(JobQueueCluster):
     # Override class variables
     submit_command = 'bsub <'
     cancel_command = 'bkill'
-    scheduler_name = 'lsf'
 
-    def __init__(self, queue=None, project=None, ncpus=None, mem=None, walltime=None, job_extra=None, **kwargs):
+    def __init__(self, queue=None, project=None, ncpus=None, mem=None, walltime=None, job_extra=None,
+                 configuration='lsf', **kwargs):
+        self.scheduler_name = configuration
```
guillaumeeb: I believe that for clarity, `self.scheduler_name` attribute should be renamed to something else. `self.config_name`?
I know this will mean much more change than this PR is currently doing.

dask_jobqueue/pbs.py
```diff
@@ -55,7 +55,9 @@ class PBSCluster(JobQueueCluster):
     cancel_command = 'qdel'
     scheduler_name = 'pbs'
```
guillaumeeb: You didn't removed scheduler_name here.

guillaumeeb:
You're right, but how do we pass this value from PBSCluster to JobQueueCluster constructor? We need to declare the kwarg in JobQueueCluster.__init__() with None as a default too?

mrocklin:
Yes, I think that that would work

dask_jobqueue/pbs.py
josephhardinee: This should be fixed in newest commits on the PR.

dask_jobqueue/lsf.py
josephhardinee: I made this change (check notes on most recent commit for more of a sense of scope).

josephhardinee:
I added another commit that I think addresses all of the points. Commit log has messages but essentially 
1. Changed to config name and it replaces self.scheduler name, but is no longer class variable.
2. Added default kwarg to JobQueueCluster constructor with it set to None. 
3. Added tests for each scheduler (Although they feel a little hacky, having to set all of the parameters). 

In particular if you could double check that 
1. The tests meet your standards and
2. The change to config_name handling on L189 of core.py is correct

mrocklin:
This looks good to me.  Thank you for the implementation and the tests @josephhardinee . Merging.

The duplication in tests is interesting.  At some point we might consider making a unittest-style `TestCase` that tests a bunch of standard things for each of the Cluster classes.  I'll raise a separate issue as a followup.

mrocklin:
Ah, was already raised as https://github.com/dask/dask-jobqueue/issues/8

guillaumeeb:
Thanks for that @josephhardinee!

Issue #181, opened 2018-10-23
Multiple cores per process/thread


dgasmith:
We have a use case where we would like to dask.distributed to parallelize a python-bound C++ program that would work best if it could consume 8-32 threads depending on the problem size and will manage threading internally. Normally, a Dask worker is run on a node that has 32 cores with the worker using 2 processes at 1 thread each so that we can give each process 16 cores.

Looking at the code currently, it seems that `nthreads = ncores / nprocesses` without exceptions, is there a canonical way to change this so that we can orchestrate our normal Dask worker operation with dask-jobqueue?

guillaumeeb:
Funnily enough, I think we just had a conversation on this point in #179.

As you point out `cores`/`processes` kwargs are used both to configure dask-worker command, and to reserve jobqueue system resources. However, the second part is just by default guessed from this kwargs. In every JobQueueCluster implementation, there should be another `kwarg` allowing you to force different resource reservation on the underlying scheduling system. For example with PBSCluster:

````python
cluster = PBSCluster(cores=2, processes=2, memory='100GB', 
                     resource_spec='select=1:ncpus=32:mem=100GB')
````

or LSFCluster:
````python
cluster = LSFCluster(cores=2, processes=2, ncpus=32)
````
 
It seems we should better document this though, and this seems useful, we were previously thinking of removing this possibility...

dgasmith:
saw on LSF you could select cpus, but I wasn't aware of the PBSCluster abilities to write out resource specs. To make things simpler, would it be possible to set the number of threads manually so that you would have `nthreads = None` which would default to current behavior.

It is worth considering this in a larger context where one might wish to start heterogeneous workers with labels where expanding `def start_workers(self, n=1):` to allow a user to override the internal settings would be very useful. Very happy to code and document here, mostly trying to figure out the current bounds of the project.

guillaumeeb:
I think there is something to do with the cores/processes/threads/ncpus/ressources_spec.. At least some standardisation between JobQueueCluster implementations. I'm not sure how it should look like though, we previously removed the use of `threads` in order to replace it by `cores`. But maybe we should remove `ncpus` and alike and add backs `threads`...

It would be more than welcome if you could propose some simple design choices here so we can discuss it, and then if we agree on something, propose a PR!

For the second part on the larger context, I'm not sure I get your point, but maybe it has another scope and should be discussed in another issue?

dgasmith:
[...]

PR #211, opened 2018-12-17
Update examples.rst

leej3:
Add an example passing arguments to workers using the extra keyword


docs/source/examples.rst
```diff
@@ -123,11 +123,10 @@ As described `here <http://distributed.dask.org/en/latest/resources.html>`_, abs
...
-       cores = 8,
-       extra=['--resources foo=2'],
-       job_extra = ["--time=03:00:00 "],
-       env_extra=['export OMP_NUM_THREADS="8"'])
+    cluster = SLURMCluster(queue='norm',
+                           memory='8g',
+                           processes=1,
+                           cores=8,
+                           extra=['--resources foo=2'],
+                           job_extra=["--time=03:00:00 "],
```
guillaumeeb: Is the space after the time needed?

lesteve:
Would it be possible to show in the example what resources can be useful through some code, or maybe just mention it in the text? At the moment it looks quite abstract ...

leej3:
@guillaumeeb you are correct, apologies.

@lesteve, what do you think of the updated version. I did want to emphasize that the resources are indeed abstract, but I agree that an example makes it a lot clearer as to why someone might wish to do this. I also wanted it to be clear what the extra keyword was doing. Is this clearer now?

leej3:
@guillaumeeb how about this? I could make it a minimal example by setting it up with something like:

```
def load(fn):
    pass
def process(data):  # foo intensive
    import time
    time.sleep(5)
def reduce(*data):  # memory intensive
    return "Processing finished..."

from dask import delayed
raw = [delayed(load)(i) for i in range(10)]
processed = [delayed(process)(r) for r in raw]
reduced = delayed(reduce)(*processed)
```
That would be useful to people. Perhaps too much to digest though...

willirath:
I might be wrong about this, but I don't see the point of having resources specified when dask-jobqueue does not yet support heterogeneous clusters.

Having said that, I'd be very interested in joining any effort towards implementing / documenting heterogeneous clusters.

guillaumeeb:
@willirath, this is more or less needed right know to nicely limit the number of tasks on a given node while still reserving all the computing resources, see https://github.com/dask/dask-jobqueue/issues/181#issuecomment-436711950.

> Having said that, I'd be very interested in joining any effort towards implementing / documenting heterogeneous clusters.

That would be really cool!

I don't know if something is feasible by hand right now, maybe it is if we modify JobQueueCluster object between scale calls. But the real idea is to properly implement that in ClusterManager. See https://github.com/dask/distributed/issues/2118 and https://github.com/dask/distributed/issues/2208#issuecomment-419140500 if you've not already.

guillaumeeb:
@leej3 your example feels hard to understand. Maybe we can start eaysier? I'm not sure.

leej3:
Is something like this more along the lines of what you mean?

docs/source/examples.rst
```diff
@@ -110,3 +110,43 @@ SLURM Deployment: Low-priority node usage
...
+The client can then be used as normal. Additionally, required resources can be
+specified for certain steps in the processing. For example:
+
+    def process(data):  # foo intensive
+        return "Result for: %s"% data
+
+
+    processed = [delayed(process)(i) for i in range(10)]
+    futures = client.compute(processed,
+                             resources={tuple(processed): {'foo': 1}})
```
guillaumeeb: Why do you use `{tuple(processed): {'foo': 1}}`, and not directly `{'foo': 1}`?

willirath:
> @willirath, this is more or less needed right know to nicely limit the number of tasks on a given node while still reserving all the computing resources, see #181 (comment).

Then my suggestion would be not to try and come up with an abstract example but present one or two from the real world.  So what about:

```python
from distributed import Client
from dask import delayed

cluster = SLURMCluster(memory='8g',
                       processes=1,
                       cores=8,
                       extra=['--resources "ssdGB=200 GPU=2"'])

cluster.start_workers(2)
client = Client(cluster)


def step_1_w_single_GPU(data):
    return "Step 1 done for: %s" % data


def step_2_w_local_IO(data):
    return "Step 2 done for: %s" % data


stage_1 = [delayed(step_1_w_single_GPU)(i) for i in range(10)]
stage_2 = [delayed(step_2_w_local_IO)(s2) for s2 in stage_1]

result_stage_2 = client.compute(stage_2,
                                resources={tuple(stage_1): {'GPU': 1},
                                           tuple(stage_2): {'ssdGB': 100}})
```

(I did not test this!)

willirath:
> I don't know if something is feasible by hand right now, maybe it is if we modify JobQueueCluster object between scale calls. But the real idea is to properly implement that in ClusterManager. See dask/distributed#2118 and dask/distributed#2208 (comment) if you've not already.

Thanks for these pointers!  I'll definitely read into this.

leej3:
Sounds good. I tested it to confirm it works.

docs/source/examples.rst
```diff
@@ -110,3 +110,52 @@ SLURM Deployment: Low-priority node usage
...
+
+
+SLURM Deployment: Providing additional arguments to the dask-workers
+-----------------------------------------
+
+Keyword arguments can be passed through to dask-workers. An example of such an
+argument is for the specification of abstract resources, described `here
+<http://distributed.dask.org/en/latest/resources.html>`_. This might be to
+specify special hardware availibility that the scheduler is not aware of, for
+example GPUs. Below, an arbitrary resource "foo" is specified. Notice that the
```
guillaumeeb: It is not an arbitrary resource anymore!

leej3:
apologies. Missed that. Fix now. Thanks

PR #213, opened 2018-12-17
Fix PBS CI : updating PBS version and centos

### LINKED ISSUE #212: PBS CI build is broken
I tracked down the problem to an update of a centOS 7 dependency: libical-devel went from 1.x to 3.x, and the path to the header changed from `/usr/include/ical.h` to `/usr/include/libical/ical.h`.

However, I was not able to force previous version installation, or tell PBS to point to the new location of ical.h during build. I can only specified `something/include/ical.h`.

guillaumeeb:
PBS tag used was quite old. Trying with a more recent version of PBS which should fix the problem encountered in #212. I must admit I'm not sure this will work.

Closes #212.

### DIFF
--- a/ci/pbs/Dockerfile
+++ b/ci/pbs/Dockerfile
@@ -8,7 +8,7 @@ RUN yum install -y gcc make rpm-build libtool hwloc-devel libX11-devel \
 	postgresql-devel python-devel tcl-devel  tk-devel swig expat-devel \
     openssl-devel libXext libXft git
 # get known PBS Pro source code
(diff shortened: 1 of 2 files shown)


guillaumeeb:
Hm, looking at https://github.com/PBSPro/pbspro/pull/756 and the state of v18.1.3 branch, I think this will not work.

We will probably need to point to pbspro master branch for a while.

guillaumeeb:
So this fixes CI for PBS, but we know point to a specific commit on PBSPro master branch.
I will merge this for now, but probably modify PBS CI later on once https://github.com/PBSPro/pbspro/pull/910 is merge, so that we point to a specific release if possible.

Issue #196, opened 2018-11-15
Add support for job arrays 


sntgluca:
Dear dask community,

I am moving my first steps with Dask and distributed, and I wish to bring up a couple of ideas and suggestions.
The first one is about enhancing the integration with the schedulers.

Dask workers are usually requested in bulk.
When the schedulers allow it, i.e. with SGE, I think it's a good practice to submit those as job (task) arrays rather than multiple independent jobs.

In systems with heavy loads it would make the work of the scheduler much easier.

I don't know the code well enough yet to come up with a complete solution, I think there are multiple places where the code needs changes. So I would like to be sure it's a good idea before working on it. 

What do you think?

guillaumeeb:
The first thought that comes to my mind here is that we've designed dask-jobqueue with simplicity in mind, as discussed and argumented by @jhamman here: https://github.com/dask/dask-jobqueue/pull/84#issuecomment-401080262.

The discussion was about multi nodes reservation at first, but I feel this apply, even with a lesser extend, to job arrays. I'd be curious to have @jhamman advice here.

What goes into your direction @sntgluca is that I believe that job arrays have an implementation almost every job queue softwares. But still, I'm not sure if this is an easy thing to implement in dask-jobqueue, and we certainly don't want to make it too complex.

One question if we implemented this, how will we able to easily implement `scale_up` and `scale_down` functions with arrays? Is it always easy to `qdel` or equivalent only part of an array of jobs? Will this be compatible with the internal state (running_jobs, pending_jobs ...) we store?
I think if we wanted to use arrays, we should do it systematically to keep only simple job templates... 

Looks like a quite heavy change.

jhamman:
@sntgluca - in the early days of dask on hpc, we explored using job arrays as a tool for deploying dask (https://github.com/dask/distributed/issues/1260#issue-243277119). IIRC, we found this approach was generally difficult to abstract to generic array jobs. It might we worth reading https://docs.dask.org/en/latest/setup/hpc.html if you have not already. 

I think there is probably a use case for array jobs with dask but I don't think dask-jobqueue is the right place. While I do see how they may lighten the load on the (SGE) scheduler, they are far less capable of dynamic scaling than what we have now.

mrocklin:
Some IT groups prefer job arrays over many single jobs because it reduces
stress of the resource manager (SLURM, SGE, etc..)  They view submitting
thousands of independent jobs as an anti-pattern.  If there is an easy way
[...]

PR #217, opened 2019-01-13
Add support for job [task] arrays in SGE/UGE

sntgluca:
This is the PoC to implement using job arrays (task arrays) #196
For the moment the implementation is _only available for the SGE cluster_.

The implementation redefines `self._cluster_template` to include the `'SGE_TASK_ID'` variable and rewrites the `start_workers()` method in `sge.py:SGECluster`.
There's a new test `test_sge.py:test_taskarrays()` to check basic stats on the new function.
I also changed the `QUEUE_WAIT` time in the tests to allow for more time (my tests required more time).

As a starting strategy, the use of task arrays is determined via a new `ta` argument in `SGECluster:start_workers()`.
* If `start_workers(ta=True)`:
the system checks if it can use task arrays and **runs the new function**.
* If `start_workers(ta=False)` or if task arrays cannot be run:
it defaults to the original `JobQueueCluster:start_workers()` function.

The `ta` argument is True by default, and it's not yet exposed to `scale()`. Which means, the use of task arrays is automatically activated by default.

### DIFF
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -86,7 +86,8 @@ class SGECluster(JobQueueCluster):
         logger.debug("Job script: \n %s" % self.job_script())
 
     def start_workers(self, n=1, ta=True):
-        __doc__ = docstrings.with_indents(""" Start workers as a task array
+        """
+        Start workers as a task array
 
         Parameters
         ----------
@@ -94,13 +95,14 @@ class SGECluster(JobQueueCluster):
             Total number of workers to start
         ta : bool
             If true, try to use task arrays
-        """, 4)
+        """
         logger.debug('starting %s workers', n)
 
         # Check if Task arrays can be used
         ta = ta and bool(os.getenv('SGE_TASK_ID', None))
 
         if ta:
+            logging.debug('Submission using task array')
             _original_command_template = self._command_template
             self._command_template = self._command_template.replace('${JOB_ID}', '${JOB_ID}.${SGE_TASK_ID}')
 
@@ -119,4 +121,5 @@ class SGECluster(JobQueueCluster):
             self.job_header = header_template
             self._command_template = _original_command_template
         else:
+            logging.debug('Submission using job ids')
             super(SGECluster, self).start_workers(n=n)
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -85,9 +85,8 @@ class SGECluster(JobQueueCluster):
 
         logger.debug("Job script: \n %s" % self.job_script())
 
-
     def start_workers(self, n=1, ta=True):
-        """ Start workers as a task array and point each task to our local scheduler
+        __doc__ = docstrings.with_indents(""" Start workers as a task array
 
         Parameters
         ----------
@@ -95,7 +94,7 @@ class SGECluster(JobQueueCluster):
             Total number of workers to start
         ta : bool
             If true, try to use task arrays
-        """
+        """, 4)
         logger.debug('starting %s workers', n)
 
         # Check if Task arrays can be used
@@ -120,4 +119,4 @@ class SGECluster(JobQueueCluster):
             self.job_header = header_template
             self._command_template = _original_command_template
         else:
-            super().start_workers(n=n)
+            super(SGECluster, self).start_workers(n=n)
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -45,7 +45,7 @@ class SGECluster(JobQueueCluster):
     cancel_command = 'qdel'
 
     def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
-                 config_name='sge', **kwargs):
+                 config_name='sge', use_job_arrays=True, **kwargs):
         if queue is None:
             queue = dask.config.get('jobqueue.%s.queue' % config_name)
         if project is None:
@@ -56,6 +56,10 @@ class SGECluster(JobQueueCluster):
             walltime = dask.config.get('jobqueue.%s.walltime' % config_name)
 
         super(SGECluster, self).__init__(config_name=config_name, **kwargs)
+
+        # Use job arrays?
+        # Revert to separate jobs if 'SGE_TASK_ID' is not defined
+        self.use_job_arrays = use_job_arrays and bool(os.getenv('SGE_TASK_ID', None))
 
         header_lines = []
         if self.name is not None:
@@ -71,6 +75,9 @@ class SGECluster(JobQueueCluster):
         if self.log_directory is not None:
             header_lines.append('#$ -e %(log_directory)s/')
             header_lines.append('#$ -o %(log_directory)s/')
+        if self.use_job_arrays:
+            header_lines.append('JOB_ID=${JOB_ID}.${SGE_TASK_ID}')
+            header_lines.append('#$ -t 1-%(num_jobs_placeholder)s')
         header_lines.extend(['#$ -cwd', '#$ -j y'])
         header_template = '\n'.join(header_lines)
 
@@ -80,12 +87,13 @@ class SGECluster(JobQueueCluster):
                   'processes': self.worker_processes,
                   'walltime': walltime,
                   'resource_spec': resource_spec,
-                  'log_directory': self.log_directory}
+                  'log_directory': self.log_directory,
+                  'num_jobs_placeholder': '%(num_jobs)d'}
         self.job_header = header_template % config
 
         logger.debug("Job script: \n %s" % self.job_script())
 
-    def start_workers(self, n=1, ta=True):
+    def start_workers(self, n=1):
         """
         Start workers as a task array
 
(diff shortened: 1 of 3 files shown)


sntgluca:
The unit test and the CI environment are not integrated.

Currently the test is passed because the `SGE_TASK_ID` variable, necessary to activate the task array, is not available, and the assertions are skipped.

I don't know the environment well enough to fix this. May someone help me?

guillaumeeb:
ping @lesteve who implemented SGECluster and associated CI.

I will try to look at this PR too.

dask_jobqueue/sge.py
```diff
@@ -82,3 +84,42 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
         self.job_header = header_template % config
 
         logger.debug("Job script: \n %s" % self.job_script())
+
+    def start_workers(self, n=1, ta=True):
```
guillaumeeb: I know this is only a poc on SGE, but I would propose a `use_job_arrays`, or `use_task_arrays` at class level. So directly in JobQueueCluster `__init__()` method in the end.

dask_jobqueue/sge.py
```diff
@@ -82,3 +84,42 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
...
+        """
+        logger.debug('starting %s workers', n)
+
+        # Check if Task arrays can be used
+        ta = ta and bool(os.getenv('SGE_TASK_ID', None))
+
+        if ta:
+            logging.debug('Submission using task array')
+            _original_command_template = self._command_template
+            self._command_template = self._command_template.replace('${JOB_ID}', '${JOB_ID}.${SGE_TASK_ID}')
```
guillaumeeb: Maybe we should always have an ID that takes nito account potential job arrays or task arrays. the task array part would potentially be empty.

dask_jobqueue/sge.py
```diff
@@ -82,3 +84,42 @@ def __init__(self, queue=None, project=None, resource_spec=None, walltime=None,
...
+
+            num_jobs = int(math.ceil(n / self.worker_processes))
+            header_template = self.job_header
+            self.job_header = header_template + '\n#$ -t 1-{}'.format(num_jobs)
```
guillaumeeb: Yes I think that this too should go in `__init__()`

dask_jobqueue/tests/__init__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import absolute_import, division, print_function
 
-QUEUE_WAIT = 15  # seconds
+QUEUE_WAIT = 120  # seconds
```
guillaumeeb: 120 seconds is a lot of time!

dask_jobqueue/tests/test_sge.py
```diff
@@ -66,3 +70,70 @@ def test_config_name_sge_takes_custom_config():
...
+
+
+@pytest.mark.env("sge")  # noqa: F811
+def test_taskarrays(loop):  # noqa: F811
+
+    def scale_cluster(cluster, size):
+        """ Starts a small cluster and returns information on jobs and tasks """
+        cluster.scale(size)
+        # waiting to allow cluster to start (no pending jobs)
+        sleep(2)
```
guillaumeeb: is this `sleep(2)` mandatory?

guillaumeeb:
> Currently the test is passed because the SGE_TASK_ID variable, necessary to activate the task array, is not available in the CI environment, and the assertions are skipped.

I'm not sure I get it right, you don't have the SGE_TASK_ID variable despite the fact of having launched a job array?

Or maybe your talking about the line
```python
ta = ta and bool(os.getenv('SGE_TASK_ID', None))
```

I did not see this, but when you call this, you're in the Notebook process, not inside a job. So there is no reason this environment variable is declared, is there?

sntgluca:
Hi @guillaumeeb ,

I am confused because I consider `SGE_TASK_ID` a standard part of the UGE suite.
For instance, it's in the wiki:
http://wiki.gridengine.info/wiki/index.php/Simple-Job-Array-Howto

But in the CI test suite it's undefined.
Therefore the line you refer to above is used to "bypass" the test if the environment (UGE) setup doesn't implement `SGE_TASK_ID`

IMHO to fix this we should find how to configure the tests to include/allow job arrays. 
I am not familiar with your test suite and with UGE setup. I was wondering if someone can help understanding if it can be done, and how.

dask_jobqueue/sge.py
sntgluca: Hi @guillaumeeb ,
this is a nice suggestion. I can implement the change.
I am working on this in my free time, so I am quite slow to follow up...

dask_jobqueue/sge.py
sntgluca: It would be excellent, I am afraid it's difficult to abstract it for all schedulers. I don't remember how job arrays are defined in different schedulers.
Will consider and come up with a suggestion.

dask_jobqueue/tests/test_sge.py
sntgluca: I need to check if it can be removed.
I added it to give time for the `cluster.pending_jobs` to be populated.
I was thinking that without it, `pending_jobs` could be empty because the jobs haven't been scheduled, not because they are running.

Do you have a better suggestion?

dask_jobqueue/sge.py
guillaumeeb: Sure, take your time, I know this situation :)

guillaumeeb:
> SGE_TASK_ID is not setup in the CI test environment, and I don't know why.

I was probably not clear. There is no reason `SGE_TASK_ID` variable is populated if you're not in an SGE job, is there?

Maybe your talking about a problem you've got later in the test process? In this case, please remove this check so that we can see what is going on!

dask_jobqueue/sge.py
sntgluca: Hi, I find this one a bit trickier, since `num_jobs` is defined when the `start_workers` function is called.
I could only think of passing `num_jobs` as
- a placeholder in the configuration (current solution):
    ```
    config = {'num_jobs_placeholder': '%(num_jobs)d'}`
    ...
    self.job_header = self.job_header % {'num_jobs' : num_jobs}
    ```
- an environmental variable to be picked up by `self.job_header`
- an attribute to substitute in `self.job_header`

Any other idea?

dask_jobqueue/sge.py
sntgluca: Hi @guillaumeeb ,
thanks for the hint!! I used your approach, and redefined JOB_ID to handle task arrays.
One caveat to note: `JOB_ID` is the variable name that UGE/SGE uses to define their Job ID. We are kinda breaking the consistency with the scheduler.

In other schedulers, as PBS, the same variable is called `PBS_JOBID`, hence there's no problem...

I don't see it as a problem, but it's worth to keep it in mind.

dask_jobqueue/tests/__init__.py
sntgluca: Yeah, the scheduler I could test with has longer waiting times...
I run my tests with a longer QUEUE_WAIT, and I forgot to revert to the default value before submitting the pull request.

Should be fixed now.

sntgluca:
Hi @lesteve ,

may you take a look at and evaluate this pull request? 
In particular, I noticed the SGE environment in the CI isn't setup for using task arrays, making my tests useless...
@guillaumeeb suggested that you would be able to help!

> ping @lesteve who implemented SGECluster and associated CI.
> 
> I will try to look at this PR too.

Thanks

lesteve:
Sorry I did not have time to look at this. To sum up, my feeling echoes the feeling of other people in the associated issue (#196): task array support does not seem worth the effort. I will comment on #196 with more complete thoughts.

Sorry if you feel disappointed because you have spent time on this without seeing it get into `dask-jobqueue`. What I am hoping is that you got more familiar with `dask-jobqueue` and it will help you for further contributions.

Contributions to `dask-jobqueue` are very welcome, it seems you are working with @ericmjl so maybe something you could look at is #195.

dask_jobqueue/sge.py
guillaumeeb: Yep, maybe we would need to change the name at one point.

dask_jobqueue/sge.py
guillaumeeb: You're right, no better idea. Just try to avoid as much as possible code duplication.

sntgluca:
@lesteve ,
I understand. I agree supporting task arrays is not a necessary feature. I partially disagree about it's importance ;) based on my experience with HPC. I doubt you'll ever receive many requests for supporting this, as it's going to affect mainly sysadmins and people working closely with them.

I am indeed more comfortable with the package now. I will try to contribute and see where this can lead me :)

sntgluca:
@guillaumeeb,

I initially added a check in SGECluster to be sure that task arrays could be used, and reverted to _using normal jobs_ if `SGE_TASK_ID` wasn't found.

To answer you question, I removed the checks from the code and from the tests. Waiting to see what happens...

sntgluca:
Closing this pull request, not in line with current ideas about task array support.

PR #220, opened 2019-01-15
SGECluster docs

### LINKED ISSUE #219: Memory specification on SGECluster
On an SGE cluster, requested memory is specified by using the `-l` flag. The exact keyword that gets used to assign memory can change from configuration to configuration; as such, it would be difficult to assume a standard string name. For example, some clusters use `m_mem_free=2G`, some use `mem_free=2G`, and others may have a different configuration.

With the SGECluster object, however, it appears that the amount of memory specified has to be passed into the `memory=` kwarg alongside the `resource_spec` string, which means that it is possible for a user to have discordant memory requested.

```python
cluster = SGECluster(queue='our.q',
                     walltime="259200",
                     processes=1,
                     cores=1,
                     # this appears to be for Dask usage only
                     memory='4GB',
                     # this is what actually gets used by the SGE schedule
                     resource_spec='m_mem_free=2G',  
                    )
```

This gives the following job script:

```
#!/bin/bash

#!/usr/bin/env bash
#$ -N dask-worker
#$ -q default.q
#$ -l m_mem_free=2G
#$ -l h_rt=259200
#$ -cwd
#$ -j y

. /home/maer3/github/research/cadd-ns-nav1.1/custom_env.sh

/home/maer3/anaconda/envs/nav/bin/python -m distributed.cli.dask_worker tcp://10.145.61.204:40499 --nthreads 1 --memory-limit 4.00GB --name dask-worker--${JOB_ID}-- --death-timeout 60
```

As far as I can see, I have requested for 2 GB of RAM from the scheduler (per worker), while the `--memory-limit 4.00GB` is reflected from the `memory` kwarg. 

Am I understanding correctly that a dask worker will think that it has 4GB of RAM, even though in reality it only was assigned 2GB here? Does this also mean, then, that if the worker is filled at 2GB of usage, the dashboard will only show 50% utilization (which would be incorrect)?

The other UI/UX issue here is that of the API - there is duplication here in specifying the `memory`  &  `resource_spec` kwargs separately. The best idea I have at the moment is to improve the documentation as it currently stands, such that it reflects this necessity. Might there be a better way to reduce this duplication?

ericmjl:
This PR closes #219. It adds a docstring update to `SGECluster`, and adds a note about memory resource requests in the Examples page.

In the limited amount of time that I attempted this PR, I was not able to build the docs locally. Hence, I have no screenshot of whether they render correctly or not. However, if they don't render properly, I am happy to edit the PR further.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -55,9 +55,21 @@ SGE Deployments
 
 On systems which use SGE as the scheduler, ``SGECluster`` can be used.
 
-SGE systems have a lot of flexibility in how they are configured, hence it is not possible to use the ``memory`` keyword argument to automatically specify the amount of RAM requested. Instead, you specify the resources desired according to how your system is configured, using the ``resource_spec`` keyword argument, in addition to the ``memory`` keyword argument (which is used by Dask internally for memory management).
+SGE systems have a lot of flexibility in how they are configured, hence it is
+not possible to use the ``memory`` keyword argument to automatically specify
+the amount of RAM requested. Instead, you specify the resources desired
+according to how your system is configured, using the ``resource_spec`` keyword
+argument, in addition to the ``memory`` keyword argument (which is used by Dask
+internally for memory management, see `this
+<http://distributed.dask.org/en/latest/worker.html#memory-management>`_ for
+more details).
 
-In the example below, our system administrator has used the ``m_mem_free`` keyword argument to let us request for RAM. Other known keywords may include ``h_vmem`` and ``mem_free``. We had to check with our cluster documentation and/or system administrator for this. At the same time, we must also correctly specify the ``memory`` keyword argument, to enable Dask's memory management to do its work correctly.
+In the example below, our system administrator has used the ``m_mem_free``
+keyword argument to let us request for RAM. Other known keywords may include
+``mem_req`` and ``mem_free``. We had to check with our cluster documentation
+and/or system administrator for this. At the same time, we must also correctly
+specify the ``memory`` keyword argument, to enable Dask's memory management to
+do its work correctly.
 
 .. code-block:: python
 
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -57,7 +57,7 @@ On systems which use SGE as the scheduler, ``SGECluster`` can be used.
 
 SGE systems have a lot of flexibility in how they are configured, hence it is not possible to use the ``memory`` keyword argument to automatically specify the amount of RAM requested. Instead, you specify the resources desired according to how your system is configured, using the ``resource_spec`` keyword argument, in addition to the ``memory`` keyword argument (which is used by Dask internally for memory management).
 
-In the example below, our system administrator has used the ``m_mem_free`` keyword argument to let us request for RAM. At the same time, we must also correctly specify the ``memory`` keyword argument, to enable Dask's memory management to do its work correctly.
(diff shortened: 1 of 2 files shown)


docs/source/examples.rst
```diff
@@ -53,7 +53,11 @@ can be used, called ``MoabCluster``:
 SGE Deployments
 ---------------
 
-On systems which use SGE as the scheduler, ``SGECluster`` can be used:
+On systems which use SGE as the scheduler, ``SGECluster`` can be used.
+
+SGE systems have a lot of flexibility in how they are configured, hence it is not possible to use the ``memory`` keyword argument to automatically specify the amount of RAM used. Instead, you must request for resources according to how your system is configured.
```
guillaumeeb: "the amount of RAM requested" ?
"Instead, you should specify the resources (...), using the resource_spec kwarg."

ericmjl: @guillaumeeb:

> "the amount of RAM requested" ?

Yes! Thank you for the catch.

> "Instead, you should specify the resources (...), using the resource_spec kwarg."

Yes again! I will change both.

ericmjl:
> But again, this shall be documented for CPU too, I imagine?

Yes, I'll add that to the PR.

docs/source/examples.rst
lesteve: Add the link to the distributed doc about dask memory management: http://distributed.dask.org/en/latest/worker.html#memory-management

docs/source/examples.rst
```diff
@@ -53,16 +53,22 @@ can be used, called ``MoabCluster``:
...
 
     from dask_jobqueue import SGECluster
 
     cluster = SGECluster(queue='default.q',
                          walltime="1500000",
-                         processes=10,
-                         memory='20GB')
+                         processes=10,   # we request 10 processes per worker
+                         memory='20GB',  # for memory requests, this must be specified
+                         resource_spec='m_mem_free=20G',  # for memory requests, this also needs to be specified
```
lesteve: maybe here or in the text above specify that `m_mem_free` is scheduler specific and that you should check your cluster doc or ask your sys admin if you are unsure. Also, if that makes sense, adding the most commonly used memory names would be a good idea (from a quick googling) could be a good idea.

lesteve:
>> But again, this shall be documented for CPU too, I imagine?

> Yes, I'll add that to the PR.

This is up to you whether you want to do that in this PR. My feeling is that the CPU is slightly more complicated, so I would be completely fine doing it in a separate PR.

My personal bias: I'd rather merge small PRs in a few days rather than having almost-finished bigger PRs lying around for 1-2 weeks ...

guillaumeeb:
> This is up to you whether you want to do that in this PR. My feeling is that the CPU is slightly more complicated, so I would be completely fine doing it in a separate PR.
> 
> My personal bias: I'd rather merge small PRs in a few days rather than having almost-finished bigger PRs lying around for 1-2 weeks ...

Agreed!

ericmjl:
> My personal bias: I'd rather merge small PRs in a few days rather than having almost-finished bigger PRs lying around for 1-2 weeks ...

Understood! Let's keep CPU docs for a different PR then.

> maybe here or in the text above specify that m_mem_free is scheduler specific and that you should check your cluster doc or ask your sys admin if you are unsure. Also, if that makes sense, adding the most commonly used memory names would be a good idea (from a quick googling) could be a good idea.

Agreed! Let me fix that in the next push.

lesteve:
I pushed some minor tweaks:
* breaking long lines
* using `mem_req` rather than `h_vmem`. My understanding is that `h_vmem` is virtual memory so I think it confuses matter. On my cluster, using PyTorch `h_vmem` needs to be set way higher than `mem_req` (probaby because of CUDA uniform memory but also maybe because PyTorch is doing some memmapping on import)

lesteve:
Merging this one, thanks a lot! A PR for CPU would be very welcome if you find the time!

ericmjl:
@lesteve thank you for the changes, and for reviewing! Same to you @guillaumeeb, for the reviews!

PR #223, opened 2019-01-16
added wait until n workers arguments

danpf:
Sometimes my scheduler is very full and it can take me a very long time to get a worker.  This way, when I do something like:
```
cluster.adapt(minimum=10,wait_until_n=2)
...{setup client}...
client.scatter([mything])
```
I don't get a timeout error on the scatter if it takes me 1hr+ to get those workers.

not sure if you are actually interested in this feature, just posting in case you are.

related to
https://github.com/dask/distributed/issues/2454

### DIFF
--- a/dask_jobqueue/deploy/cluster_manager.py
+++ b/dask_jobqueue/deploy/cluster_manager.py
@@ -2,6 +2,7 @@ import dask
 import logging
 import math
 import os
+import time
 
 from tornado import gen
 
@@ -83,8 +84,14 @@ class ClusterManager(object):
         self._adaptive_options = adaptive_options
         self._adaptive_options.setdefault('worker_key', self.worker_key)
 
+    def sleep_until_n_workers(self, n):
+        '''Block by sleeping until we have n active workers'''
+        while self._count_active_workers() < n:
+            time.sleep(1)
+
     def adapt(self, minimum_cores=None, maximum_cores=None,
-              minimum_memory=None, maximum_memory=None, **kwargs):
+              minimum_memory=None, maximum_memory=None,
+              wait_until_n=0, **kwargs):
         """ Turn on adaptivity
         For keyword arguments see dask.distributed.Adaptive
         Instead of minimum and maximum parameters which apply to the number of
@@ -100,6 +107,8 @@ class ClusterManager(object):
             Minimum amount of memory for the cluster
         maximum_memory: str
             Maximum amount of memory for the cluster
+        wait_until_n: int
+            Blocking sleep until we have this many active workers
         Examples
         --------
         >>> cluster.adapt(minimum=0, maximum=10, interval='500ms')
@@ -122,6 +131,7 @@ class ClusterManager(object):
                 kwargs['maximum'] = self._get_nb_workers_from_memory(maximum_memory)
         self._adaptive_options.update(kwargs)
         self._adaptive = Adaptive(self.scheduler, self, **self._adaptive_options)
+        self.sleep_until_n_workers(wait_until_n)
         return self._adaptive
 
     @property
@@ -173,7 +183,7 @@ class ClusterManager(object):
                 self.scale_down(to_close, n)
             self._target_scale = n
 
-    def scale(self, n=None, cores=None, memory=None):
+    def scale(self, n=None, cores=None, memory=None, wait_until_n=0):
         """ Scale cluster to n workers or to the given number of cores or
         memory
         number of cores and memory are converted into number of workers using
@@ -186,6 +196,8 @@ class ClusterManager(object):
             Target number of cores
         memory: str
             Target amount of available memory
+        wait_until_n: int
+            Blocking sleep until we have this many active workers
         Example
         -------
         >>> cluster.scale(10)  # scale cluster to ten workers
@@ -200,6 +212,7 @@ class ClusterManager(object):
         # TODO we should not rely on scheduler loop here, self should have its
         # own loop
         self.scheduler.loop.add_callback(self._scale, n, cores, memory)
+        self.sleep_until_n_workers(wait_until_n)
 
     def _widget_status(self):
         workers = len(self.scheduler.workers)
--- a/dask_jobqueue/deploy/cluster_manager.py
+++ b/dask_jobqueue/deploy/cluster_manager.py
@@ -86,7 +86,7 @@ class ClusterManager(object):
 
     def wait_until_n_workers(self, n):
         '''Block by sleeping until we have n active workers'''
-        while len(self.scheduler.workers) < n:
+        while n and len(self.scheduler.workers) < n:
             time.sleep(1)
 
     def adapt(self, minimum_cores=None, maximum_cores=None,
--- a/dask_jobqueue/deploy/cluster_manager.py
+++ b/dask_jobqueue/deploy/cluster_manager.py
@@ -84,9 +84,9 @@ class ClusterManager(object):
         self._adaptive_options = adaptive_options
         self._adaptive_options.setdefault('worker_key', self.worker_key)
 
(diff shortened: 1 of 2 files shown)


danpf:
it is the same problem as there, but it isn't.
Since jobqueue ClusterManagers don't inherit from distributed Clusters, they both would have to implement this separately.

there is also i think a bug here since Adaptive gets the wait_until_n kwarg... but i haven't seen any problems yet.

lesteve:
> Since jobqueue ClusterManagers don't inherit from distributed Clusters, they both would have to implement this separately.

I see, good point. Still I think something like this should be done in distributed because it is applicable in other projects (`LocalCluster` in `distributed`, `KubernetesCluster` in `dask-kubernetes`, etc ...) and I think a PR in `distributed` would be very welcome. 

We can then think about how to have the same functionality in `dask-jobqueue` with minimal code and deferring to the `distributed` implementation. Full disclosure: I haven't followed very closely the ClusterManager development so the situation may be a bit more complicated than this, not 100 % sure ...

You may do something like this already, but in the mean time, I would recommend to use a work-around like https://github.com/dask/distributed/issues/2138#issuecomment-419605424.

guillaumeeb:
For the reason behind ClusterManager, see #170. The idea is to use dask-jobqueue as an experimental project for designing the common part from distributed codebase. 

The `wait_for_n` feature here is another to add to the ClusterManager features. Maybe it is simple enough to be put in distributed right now, I'm not sure how we want to do things here. @mrocklin any opinion?

dask_jobqueue/deploy/cluster_manager.py
```diff
@@ -83,8 +84,14 @@ def __init__(self, adaptive_options={}):
         self._adaptive_options = adaptive_options
         self._adaptive_options.setdefault('worker_key', self.worker_key)
 
+    def sleep_until_n_workers(self, n):
+        '''Block by sleeping until we have n active workers'''
+        while self._count_active_workers() < n:
```
guillaumeeb: In ClusterManager, we should not rely on functions from JobQueueCluster. You need to find something than can be called on the scheduler remotly at one point. Or we need the ClusterManager to be informed of any incoming worker.
What shall be kept in mind is that the ClusterManager is potentially another process than the Scheduler, and that it is build to be usable by other Cluster implementation than those of dask-jobqueue.

mrocklin:
No strong opinion from me

guillaumeeb:
Let's work on this in dask-jobqueue first, and we'll see how it can be made available to distributed if needed.

danpf:
it's pretty much done. pending tests.

danpf:
There is some test leaking somehow. I'm not sure why but If i just move some tests aroudn like in the last commit the tests pass.

dask_jobqueue/deploy/cluster_manager.py
```diff
@@ -122,6 +131,7 @@ def adapt(self, minimum_cores=None, maximum_cores=None,
                 kwargs['maximum'] = self._get_nb_workers_from_memory(maximum_memory)
         self._adaptive_options.update(kwargs)
         self._adaptive = Adaptive(self.scheduler, self, **self._adaptive_options)
+        self.wait_until_n_workers(wait_until_n)
```
mrocklin: Adapt can be called within the event loop, so we can't have a `while ...: time.sleep()` call in there, otherwise things will halt/block.

dask_jobqueue/deploy/cluster_manager.py
```diff
@@ -83,8 +84,14 @@ def __init__(self, adaptive_options={}):
         self._adaptive_options = adaptive_options
         self._adaptive_options.setdefault('worker_key', self.worker_key)
 
+    def wait_until_n_workers(self, n):
+        '''Block by sleeping until we have n active workers'''
+        while n and len(self.scheduler.workers) < n:
+            time.sleep(1)
```
mrocklin: If we keep this function then I would decrease this sleep significantly, probably to around 10ms.  It'll still be way less than 1% cpu time.

It's common in Dask to build functions that are async-friendly first, and then wrap around them with `self.cluster.sync`.  If we're supporting Python 2 then we do this with `@gen.coroutine`, if we're not then we use `async def`.

dask_jobqueue/deploy/cluster_manager.py
```diff
@@ -200,6 +212,7 @@ def scale(self, n=None, cores=None, memory=None):
         # TODO we should not rely on scheduler loop here, self should have its
         # own loop
         self.scheduler.loop.add_callback(self._scale, n, cores, memory)
+        self.wait_until_n_workers(wait_until_n)
```
mrocklin: I think that having a keyword like this in many methods is a sign that maybe we shouldn't include it, and instead we should expect users to make two calls.

```diff
-cluster.scale(10, wait_until_n=10)
+cluster.scale(10)
+cluster.wait_until_n_workers(10)
```

mrocklin: Otherwise keyword arguments like that tend to propagate to many new methods, resulting in an `n x m` maintenance problem.

mrocklin:
Should have said first though, this is nice work.  People have asked for things like this before, so thank you @danpf for pushing on it.

mrocklin:
I've added a comment to the original issue: https://github.com/dask/distributed/issues/2138#issuecomment-469098420

guillaumeeb:
@mrocklin from what I understand from your comment, expecting user to make and explicit call to the method also solve the problem related to the lack of async support, or am I missing something?

mrocklin:
If you don't call wait within scale or adapt then having a synchronous-only method is fine, as long as people don't try to call it from async contexts.  Most people *don't* use async programming so supporting async isn't a big deal.  The implementation in the issue *should* work out OK though, and hopefully isn't too complex.

danpf:
As one of the few people that use async -->
I don't have, nor forsee any reasons to asynchronously scale nor adapt.  100 % of the time currently for me my scripts look like:
1. distributed client is initted with adapt, or scale, wait until workers are up
2.  then actually start doing complex async workflows.

since we depend on distributed workers for actual computations, there's really nothing else we could be doing instead of synchronously sleeping, so it's not a big deal for us.
I understand the need to plan ahead, but figured I would just mention my actual usage to provide some context.

guillaumeeb:
Even if we leave the `wait_until_n_workers` code into scale or adapt, there's no reason dask-labexstension would call it with a `wait` kwarg > 0, so I don't think this would block the loop.

Anyway I'm really new to async programming, so I leave this discussion to both of you. @danpf, is there a downside into handling async call, except a few lines of code more?

mrocklin:
> there's really nothing else we could be doing instead of synchronously sleeping

Well, there are some things you would be doing, like having the scheduler respond to the workers that are coming in and update its count of current workers.  The current loop, if run within the event loop, will never terminate because the scheduler will never be able to acknowledge the new workers.

Calling blocking calls within the event loop effectively shuts down the entire system.

guillaumeeb:
As moving to async as propose by @mrocklin in https://github.com/dask/distributed/issues/2138#issuecomment-469098420 seems not to be a big deal, I propose @danpf implements it, along with the sleeping time at 0.01, or 0.1 which is probably reactive enough.

As per not including the call in `scale` and `adapt` and the associated keyword, I'm still undecided...

danpf:
Sorry I'm really busy for a while, probably ~2 weeks I can do this.  apologies if that's too late.

guillaumeeb:
@danpg, do you think you can follow up on this one in the near future?

guillaumeeb:
So @danpf, I believe this is superseeded by https://github.com/dask/distributed/pull/2688, so we can close this one?

danpf:
yup!

PR #224, opened 2019-01-18
DOC: mention that SGECluster can be used with any Grid Engine derivative

lesteve:
I added a bit more content about the reason behind the variety of Grid Engine-like deployments.

@ericmjl if you want to have a quick look, let me know if this can be mader clearer.


ericmjl:
@lesteve I looked through the updated text, and it looks good to me!

lesteve:
Thanks a lot for your comment! Let's merge this one then!

PR #232, opened 2019-02-04
Support older SLURM versions.

lesteve:
`--parsable` exists for SLURM version >= 14-03-03-1 (released March 2014).

Fix #227.

PR #233, opened 2019-02-04
Simplify loop fixture usage in tests

lesteve:
By putting the `loop` import in `conftest.py`, it is readily available in all tests.

This way we can get rid of the hacks we had for making flake8 happy (F401/F801 dance that does not seem to work anymore with the latest version of `flake8`, see https://github.com/dask/dask-jobqueue/pull/232#issuecomment-460215666).

I could not resist fixing a few flake8 violations in conftest.py.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,14 +1,21 @@
 # content of conftest.py
 
+# Make loop fixture available in all tests
+from distributed.utils_test import loop  # noqa: F401
+
 import pytest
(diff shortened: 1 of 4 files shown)


lesteve:
Merging this one since this will fix the flake8 error and seems uncontroversial.

@guillaumeeb don't hesitate if you have comments/explain.

Issue #235, opened 2019-02-07
SGE parallel environment spec


ericmjl:
GridEngine clusters apparently allow for a `-pe` flag. This maps to a "parallel environment" specification. 

What are opinions on modifying the SGECluster source code to add in a `pe_spec` kwarg, such that end users who are knowledgeable can add this in?

Wanted to get opinions on this before I did work on it.

jhamman:
I would think this is already possible using `job_extra=['-pe orte']` (for example). The `job_extra` kwarg is there to allow arbitrary additions to the job script, line by line.

ericmjl:
Oh, this is great to know! I guess I should PR into the docs example instead, then :smile:

guillaumeeb:
I'm closing this issue as there is an already built-in solution for cases like this as answered by @jhamman . But @ericmjl do not hesitate to propose a PR for documentation update, this would be more than welcomed!

PR #256, opened 2019-03-22
Add a `job_extra` flag to the SGECluster

Lnaden:
The SGECluster has no `job_extra` setting like all the other clusters do, which limits its abilities. This PR adds the `job_extra` flag in the same way as it appears in the other cluster files.

xref #219 and #235 which seemed to allude to `SGECluster` having the `job_extra` flag, but no such flag appears to exist.


lesteve:
Merging, thanks a lot @Lnaden!

guillaumeeb:
Glad to see you back around @lesteve, I'm quite taken by work since several weeks!

PR #260, opened 2019-04-08
Use pytest rather than py.test

lesteve:
It's been a while since `pytest` command exists (pytest 3.0.0 released in August 2016). From [this](https://docs.pytest.org/en/latest/changelog.html#id760):

> Introduce pytest command as recommended entry point. Note that py.test still works and is not scheduled for removal. Closes proposal #1629. Thanks @obestwalter and @davehunt for the complete PR (#1633).


lesteve:
CIs are passing, merging!

PR #274, opened 2019-05-23
LSF: Add unit scaling other than MB

Lnaden:
The `lsf.conf` file can specify units other than MB for memory, and
any other large units of scaling, through the `LSF_UNIT_FOR_LIMITS`
variable.

Dask Jobqueue assumes that the units are MB, which is true under
the default setup of LSF, but not always.

* If the admin has changed it to something else in the file.
* If the file does not exist, LSF assumes KB.

This PR adds the ability for users to both specify or allow auto
detection of the unit system through an `lsf_units` kwarg or `lsf-units`
YAML arg.

Auto-detection routine is based on the LSF 9.1.3 docs for where to
look, defaults, and variable names.

### DIFF
--- a/dask_jobqueue/jobqueue.yaml
+++ b/dask_jobqueue/jobqueue.yaml
@@ -139,6 +139,7 @@ jobqueue:
     mem: null
     job-extra: []
     log-directory: null
+    lsf-units: null
 
   htcondor:
     name: dask-worker
--- a/dask_jobqueue/lsf.py
+++ b/dask_jobqueue/lsf.py
@@ -33,7 +33,7 @@ class LSFCluster(JobQueueCluster):
         prepended with the #LSF prefix.
     lsf_units : str
         Unit system for large units in resource usage set by the 
-        LSF_UNIT_FOR_LIMTS in the lsf.conf file of a cluster.
+        LSF_UNIT_FOR_LIMITS in the lsf.conf file of a cluster.
     %(JobQueueCluster.parameters)s
 
     Examples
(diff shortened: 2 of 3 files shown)


Lnaden:
For reference, here are the official LSF docs on the mater

https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.3/lsf_config_ref/lsf.conf.lsf_unit_for_limits.5.html

Lnaden:
I think I fixed all the bugs now, I'm not sure where I introduced `trailing whitespace` though, I don't think I did unless the number of blank lines in the docstrings/YAML file is important?

guillaumeeb:
Thanks @Lnaden for this, I did not have the time to take a look at it tight now, but I'll try as soon as possible. In the mean time if any other maintainer or experimented user can look at this, this could be nice!

guillaumeeb:
> I don't think I did unless the number of blank lines in the docstrings/YAML file is important?

You've got a trailing white space on line 35 of lsf.py, in the docstring.

Also did you apply black on your codes?

Lnaden:
I'll work on a test and black today. There are also some changes in the 10.1+ version of LSF, so I'm going to make changes to work in both

Lnaden:
I made some changes to include the LSF 10.1+ way of doing things, you can apparently specify everything from KB up to ZB, and even then in single letter short-hand. I think there may be an issue with the `ceil` function if an LSF cluster has ZB or something else high order specified, but I can't confirm that so may have to leave that as an untested corner case?

For reference, here is the 10.1 docs:

https://www.ibm.com/support/knowledgecenter/en/SSWRJV_10.1.0/lsf_config_ref/lsf.conf.lsf_unit_for_limits.5.html

I made the following changes as well as requested:

* Fixed trailing whitespace
* Added tests for the LSF auto-unit detection routine
* Ran Black

PR #277, opened 2019-06-06
Add missing parameter docstring and removed unused parameter.

lesteve:
* `log_directory` and `config_name` were not in the docstring
* walltime parameter is not used inside `JobQueueCluster` (already in derived classes)
* Moved down python in the docstring to match signature order.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -121,10 +121,14 @@ class JobQueueCluster(ClusterManager):
         Additional arguments to pass to `dask-worker`
     env_extra : list
         Other commands to add to script before launching worker.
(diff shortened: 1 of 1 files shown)


dask_jobqueue/core.py
```diff
@@ -174,7 +178,6 @@ def __init__(
         extra=None,
         env_extra=None,
         log_directory=None,
-        walltime=None,
```
guillaumeeb: Just one question about this removal: shouldn't we instead remove it from all JobqueueCluster impl?

guillaumeeb:
Let's merge this for now.

PR #278, opened 2019-06-13
Use host= keyword rather than ip=

mrocklin:
This will change in the next release.  We should merge this and release after we next release dask/distributed.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -241,9 +241,9 @@ class JobQueueCluster(ClusterManager):
 
         if interface:
             extra += ["--interface", interface]
-            kwargs.setdefault("ip", get_ip_interface(interface))
+            kwargs.setdefault("host", get_ip_interface(interface))
         else:
-            kwargs.setdefault("ip", "")
+            kwargs.setdefault("host", "")
 
         # Bokeh diagnostics server should listen on all interfaces
         kwargs.setdefault("dashboard_address", ("", 8787))
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -101,7 +101,7 @@ def test_forward_ip():
         cores=8,
         memory="28GB",
         name="dask-worker",
-        ip=ip,
+        host=ip,
     ) as cluster:
         assert cluster.local_cluster.scheduler.ip == ip
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 dask>=1.2.0
-distributed
+distributed>=1.27
 docrep
+pre-commit
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 dask>=1.2.0
 distributed>=2
 docrep
-pre-commit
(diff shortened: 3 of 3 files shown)


lesteve:
@mrocklin do you think we should merge this one? My understanding:
* `ip` is marked deprecated in `LocalCluster` docstring but using `ip` does not create a warning in master yet.
* there was a failure at the time you opened the PR (distributed 1.28.1) which seemed genuine (the test is supposed to check that by default the scheduler starts on an IP address that is reachable from other nodes). This seems fixed in distributed 2.1 although I was not able to  find the related entry in the distributed changelog.

More details: this PR was failing when @mrocklin originally opened the PR (June 13), see this [Travis log](https://travis-ci.org/dask/dask-jobqueue/jobs/549040567). The error was:

```
_______________________________ test_forward_ip ________________________________
    def test_forward_ip():
        ip = "127.0.0.1"
        with PBSCluster(
            walltime="00:02:00",
            processes=4,
            cores=8,
            memory="28GB",
            name="dask-worker",
            host=ip,
        ) as cluster:
            assert cluster.local_cluster.scheduler.ip == ip
    
        default_ip = socket.gethostbyname("")
        with PBSCluster(
            walltime="00:02:00", processes=4, cores=8, memory="28GB", name="dask-worker"
        ) as cluster:
>           assert cluster.local_cluster.scheduler.ip == default_ip
E           AssertionError: assert '127.0.0.1' == '0.0.0.0'
E             - 127.0.0.1
E             + 0.0.0.0
dask_jobqueue/tests/test_jobqueue_core.py:104: AssertionError
=============== 1 failed, 78 passed, 15 skipped in 2.92 seconds ================
```

I was able to reproduce the original error locally with distributed=1.28.1.

mrocklin:
> ip is marked deprecated in LocalCluster docstring but using ip does not create a warning in distributed master yet.

Right.  At one point in development we were adding a warning if the user used `ip=`, but then after submitting this PR I decided that that was probably too aggressive, particularly because this would effectively force dask-jobqueue to drop Python 2 going forward (dask 2.0 drops the `ip=` keyword and Python 2, so dask-jobqueue would have to either accept annoying warnings, or drop Python 2 as well).  I thought that HPC users are probably a little bit less able to move to Python 3 and so there wasn't a rush.

I would be happy to merge this any time if you like.  I'll leave it to you.

lesteve:
We dropped Python 2 support for 0.6 (roughly 2 weeks ago). 0.6 was supposed to be compatible with dask >= 2 but it looks like we missed a few things, which I am trying to fix in the next release.

About dropping Python 2 support for HPC users, all of the people around me use Python 3 already with a conda installed as a user in a shared network drive. If you want to bring in other people that may have another experience from HPC setup into the conversation, you would be more than welcome to do so!

mrocklin:
If the dask-jobqueue maintainers want to drop Python 2 then I'm all for
it.  I just didn't want to force the issue.

lesteve:
Fine, let's merge this one then!

lesteve:
Thanks @mrocklin!

lesteve:
> 0.6 was supposed to be compatible with dask >= 2 but it looks like we missed a few things, which I am trying to fix in the next release.

BTW @guillaumeeb or @mrocklin if you could give me the rights on PyPI, that would be great! My user name on PyPI is lesteve as well.

mrocklin:
Done!

lesteve:
Great thamks a lot!

Issue #280, opened 2019-06-26
worker_spec is set incorrectly


danpf:
@guillaumeeb 

Hey I haven't updated dask-jobqueue in a while but after updating i've found a slight bug. that `worker_spec` (at least in `distributed` should be a dictionary of ids to workers(ie the len of the current # of workers), not a specification of workers to make (like `worker_spec = {'cores': 4, 'memory': '16 GB'}`)

currently if you have your minimum number of workers at 0, (and possibly anything less than 2, distributed/deploy/adaptive.Adaptive checks 
```
        try:
            current = len(self.cluster.worker_spec)
        except AttributeError:
            current = len(self.cluster.workers)
```
to get the number of current workers, and so we just forever sit and spin our wheels because Adaptive thinks we have enough workers already (in reality we haven't even scaled yet :o )

it looks like this happened on https://github.com/dask/dask-jobqueue/pull/184 ?

i can see what you're tying to do, maybe we just need to name jobquque 'worker_spec' to something else?

guillaumeeb:
I believe this is a regression due to https://github.com/dask/distributed/pull/2675 and https://github.com/dask/distributed/pull/2735, the introduction of SpecificationCluster in Dask Distributed.

I wanted to open an issue for this, but I still didn't find the time to dive into SpecificationCluster changes. Obviously, your issue shows that dask-jobqueue is not compatible with this new interface.

So the fastest change is probably to rename this attribute. I encourage you to propose a PR for this. In the longer term, we need to be compatible with SpecificationCluster, I'll open an issue for that.

danpf:
@guillaumeeb 
could you bump the version / make a release so that the conda version gets built?

guillaumeeb:
Closed by #282.

Release done!

PR #282, opened 2019-06-26
Rename worker_spec attribute

ocaisa:
Short-term fix for #280

For tests to pass, requires #284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,19 +9,10 @@ services:
   - docker
 matrix:
   include:
-    - python: "2.7"
-      env:
-        - OS=ubuntu-14.04
-        - JOBQUEUE=sge
     - python: "3.6"
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
-    - python: "2.7"
-      env:
-        - OS=ubuntu-14.04
-        # JOBQUEUE=none is for tests that do not need a cluster to run
-        - JOBQUEUE=none
     - python: "3.6"
       env:
         - OS=ubuntu-14.04
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,19 @@ services:
   - docker
 matrix:
   include:
+    - python: "2.7"
+      env:
+        - OS=ubuntu-14.04
+        - JOBQUEUE=sge
     - python: "3.6"
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
+    - python: "2.7"
+      env:
+        - OS=ubuntu-14.04
+        # JOBQUEUE=none is for tests that do not need a cluster to run
+        - JOBQUEUE=none
     - python: "3.6"
       env:
         - OS=ubuntu-14.04
--- a/dask_jobqueue/tests/test_sge.py
+++ b/dask_jobqueue/tests/test_sge.py
@@ -32,7 +32,7 @@ def test_basic(loop):
             workers = list(client.scheduler_info()["workers"].values())
             w = workers[0]
             assert w["memory_limit"] == 2e9 / 4
-            assert w["ncores"] == 2
+            assert w["nthreads"] == 2
 
             cluster.scale(0)
 
--- a/dask_jobqueue/tests/test_sge.py
+++ b/dask_jobqueue/tests/test_sge.py
@@ -32,7 +32,12 @@ def test_basic(loop):
             workers = list(client.scheduler_info()["workers"].values())
             w = workers[0]
             assert w["memory_limit"] == 2e9 / 4
-            assert w["nthreads"] == 2
+            if "nthreads" in w:
+                nthreads = "nthreads"
+            else:
+                # for python2
+                nthreads = "ncores"
+            assert w[nthreads] == 2
 
             cluster.scale(0)
 
--- a/dask_jobqueue/tests/test_sge.py
+++ b/dask_jobqueue/tests/test_sge.py
@@ -32,12 +32,7 @@ def test_basic(loop):
             workers = list(client.scheduler_info()["workers"].values())
             w = workers[0]
             assert w["memory_limit"] == 2e9 / 4
-            if "nthreads" in w:
-                nthreads = "nthreads"
-            else:
-                # for python2
-                nthreads = "ncores"
-            assert w[nthreads] == 2
+            assert w["nthreads"] == 2
 
             cluster.scale(0)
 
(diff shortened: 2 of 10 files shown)


ocaisa:
Closing and  reopening to see if that affects tests passing

danpf:
thank you for doing this @ocaisa !!!

one of your problems is that ncores is now nthreads

I'm not really sure what the other problems could be. some of the complaints are about AttributeErrors but that should be caught by:
https://github.com/dask/distributed/blob/master/distributed/deploy/adaptive.py#L216
(and the line numbers in the error match so it should be okay?)

danpf:
you can revert that last change, requirements.txt isn't respected :(
it's built by conda and just uses the newest (in this case 2.0)
https://github.com/dask/dask-jobqueue/blob/master/ci/slurm/Dockerfile#L8

I would suggest giving `JobQueuePlugin` a `workers` property that returns `self.scheduler.workers`

I don't actually know if that is the answer, but digging deeper into the logs 
```
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/gen.py", line 209, in wrapper
    yielded = next(result)
  File "/opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 259, in _adapt
    recommendations = self.recommendations()
  File "/opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 218, in recommendations
    current = len(self.cluster.workers)
AttributeError: 'SLURMCluster' object has no attribute 'workers'
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOLoop object at 0x7f81de0b95f8>>, <Future finished exception=AttributeError("'SLURMCluster' object has no attribute 'workers'",)>)
  File "/opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 216, in recommendations
    current = len(self.cluster.worker_spec)
AttributeError: 'SLURMCluster' object has no attribute 'worker_spec'
During handling of the above exception, another exception occurred:
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/opt/anaconda/lib/python3.6/site-packages/tornado/gen.py", line 209, in wrapper
    yielded = next(result)
  File "/opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 259, in _adapt
    recommendations = self.recommendations()
  File "/opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 218, in recommendations
    current = len(self.cluster.workers)
AttributeError: 'SLURMCluster' object has no attribute 'workers'
```

that's probably at least one of the problems

if you have a better idea feel free to try that instead, i think that should get us in the right direction though

ocaisa:
Thanks @danpf , that suggestion worked. The last issue is that when using Python2, we still have`ncores` rather than `nthreads`, I can make this work with an if/else to get this set of tests to pass but it's probably an indicator of a necessary break with Python2 support.

willirath:
Dask has dropped Python 2 with v2.0.0:

- https://distributed.dask.org/en/latest/changelog.html#id2
- https://docs.dask.org/en/latest/changelog.html#core

In my opinion, this should be reflected here (the repo) as well.

guillaumeeb:
I agree with @willirath on the Python 2 support.

I'll try to review the PR next week, but if someone else (@danpf @willirath) can do it before, I'll be quite happy!

dask_jobqueue/tests/test_sge.py
```diff
@@ -32,7 +32,12 @@ def test_basic(loop):
             workers = list(client.scheduler_info()["workers"].values())
             w = workers[0]
             assert w["memory_limit"] == 2e9 / 4
-            assert w["ncores"] == 2
+            if "nthreads" in w:
+                nthreads = "nthreads"
+            else:
+                # for python2
+                nthreads = "ncores"
+            assert w[nthreads] == 2
```
danpf: revert changes for ncores, we will remove python2 from testing

ocaisa:
I looked at what was done in `distributed` (which is what you suggested) and, given the behaviour we've seen in the testing here, it looks like that  is enough to stop serving to python2.

danpf:
Great! thanks @ocaisa & nice job! everything looks good to me!

willirath:
Looks good to me as well.  The pedant in me tells me that this should be two PR's:

1. Drop Python 2 testing.  Should also add a clear statement about dropping 2 to the https://github.com/dask/dask-jobqueue/blob/master/docs/source/changelog.rst

2. Fix #280

ocaisa:
@willirath I agree you with there, something like dropping python2 support deserves a distinct PR. Will open now.

.travis.yml
```diff
@@ -9,19 +9,10 @@ services:
   - docker
 matrix:
   include:
-    - python: "2.7"
```
willirath: Don't do this here, but tackle in #283.

.travis.yml
```diff
@@ -9,19 +9,10 @@ services:
...
-      env:
-        - OS=ubuntu-14.04
-        - JOBQUEUE=sge
     - python: "3.6"
       env:
         - OS=ubuntu-14.04
         - JOBQUEUE=sge
-    - python: "2.7"
```
willirath: Don't do this here, but tackle in #283.

setup.py
```diff
@@ -21,6 +21,7 @@
     description="Easy deployment of Dask Distributed on job queuing systems "
     "such as PBS, Slurm, or SGE.*",
     url="https://github.com/dask/dask-jobqueue",
+    python_requires=">3.5.0",
```
willirath: Don't do this here, but tackle in #283.

willirath:
@ocaisa Sorry again for the confusion.  The proposed separation was not so much about the importance or impact of dropping Python 2 warranting a separate PR, but about keeping a clean history that separates commits to master when different purposes are served.  I have tried to clarify this in the reviews of #282 and #283 now, I hope:

#282 just fixes `worker_spec` and disregards red marks on Py2 related tests for now.

#283 drops Python 2 testing but doesn't touch anything else.

PR #290, opened 2019-07-16
Add links to @willirath tutorial

lesteve:
I added them on the docs and in the README. Any comments, let me know!

This is how it looks on the documentation generated locally:

image.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,6 +6,7 @@ Deploy Dask on Job Queueing systems
 Easy deployment of Dask Distributed on job queuing systems such as PBS, Slurm,
 or SGE.  See documentation_ for more information.
 
+
 LICENSE
 -------
 
--- a/README.rst
+++ b/README.rst
@@ -6,6 +6,11 @@ Deploy Dask on Job Queueing systems
 Easy deployment of Dask Distributed on job queuing systems such as PBS, Slurm,
 or SGE.  See documentation_ for more information.
 
+A 4 hour ``dask`` and ``dask-jobqueue`` tutorial was presented in July 2019 by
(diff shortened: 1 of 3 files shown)


lesteve:
I wanted to say it is slightly weird to have the Tutorial non-section before the "Getting started" but I figure it is better to make the tutorial prominent rather than hidden in a sub-page.

willirath:
Thanks for adding these!

There's a DOI now: https://doi.org/10.5281/zenodo.3337784

But the look and feel is much nicer in the Github repo.  So I'm not sure which to better include.  The DOI is much more stable, though.

willirath:
> I wanted to say it is slightly weird to have the Tutorial non-section before the "Getting started" but I figure it is better to make the tutorial prominent rather than hidden in a sub-page.

What about putting it in a Tutorials page under "Detailed Use"? There, we could also link the video from the front page and the slides / video of @andersy005's SciPy2019 talk.

lesteve:
I haven't had time to catch up with SciPy I have to admit (just watching Matt's talk now). My not super strong opinions:
* I'd be in favour to make the tutorial(s) prominent. So maybe a reasonable compromise is to have a link with tutorials on the main documentation page that goes to a "Tutorials" sub-section.
* I'd go for the github repo rather than the DOI. user-friendly trumps stability in this particular case.

willirath:
Sounds great!

lesteve:
Pushed an update. I put this in the "Getting started" section, because I feel this is where I would expect a tutorial to show up rather than "detailed use". Let me know what you think.

image

willirath:
Looks great!

willirath:
If there's no objections within the next hour or so, I'd merge.

PR #297, opened 2019-07-23
Remove threads from JobQueueCluster arguments.

lesteve:
Trying to use `threads` has raised an informative error message since dask-jobqueue 0.3 (about one year), more precisely this commit: https://github.com/dask/dask-jobqueue/commit/d4d47329eeb5c6216334987032e21e058a6f0cf8.

I think we can safely remove this. Objections more than welcome!

As a reminder: this was when we migrated from `processes` and `threads` (implicitly per dask worker) to `cores` and `processes`.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -7,7 +7,6 @@ import re
 import shlex
 import subprocess
 import sys
-import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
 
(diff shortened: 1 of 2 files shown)


dask_jobqueue/core.py
```diff
@@ -223,9 +207,6 @@ def __init__(
         if shebang is None:
             shebang = dask.config.get("jobqueue.%s.shebang" % config_name)
 
-        if dask.config.get("jobqueue.%s.threads", None):
```
lesteve: Note: having `threads: something` in a dask config would warn and be ignored before this PR. After this PR, it will be silently ignored.

lesteve:
Merging this one.

PR #301, opened 2019-07-26
Fix spacing between parameter and parameter type in the API docs

lesteve:
This can be quite confusing: e.g. it looks like the parameters are `queuestr` and `projectstr` in this screenshot:
image

* this is related to a change in sphinx 2. This was reported here in numpydoc: https://github.com/numpy/numpydoc/issues/215
* there has been a sphinx fix in https://github.com/sphinx-doc/sphinx/pull/5976
* that sphinx_rtd_theme does not use basic.css. Since dask_sphinx_theme does extends sphinx_rtd_theme we are not benefitting from the sphinx fix. Hopefully this will get merged eventually https://github.com/readthedocs/sphinx_rtd_theme/pull/747

The simplest fix I found was to add `basic.css` through `setup(app)` in `conf.py`

cc @TomAugspurger since i saw he had similar problems in `dask-ml` which he fixed by pinning `sphinx < 2` (except that we can not do that on RTD, which is another story by itself ...)

This is the same screenshot as above from a locally generated doc:
image

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,6 +6,8 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/stable/config
 
+from dask_jobqueue import __version__ as version
+
 # -- Path setup --------------------------------------------------------------
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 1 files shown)


lesteve:
I am going to merge this one, let me know if you spot problems in the doc that may be related to this PR.

guillaumeeb:
Thanks @lesteve! Noticed this too, this caused some missunderstanding in some issues. It's really nice to have it fixed!

Issue #298, opened 2019-07-23
Cobalt Support


Phantom139:
Continuing conversation from #202, I have an allocation on Argonne's Theta cluster which uses the Cobalt job scheduling system. While the general framework of dask-jobqueue does seem to fit in with the background process once the job is in the queue, there are a few nitpicky things I have been trying to work out over the course of the day while also dealing with a high user workload causing testing to be slow.

My work so far has been placed on the develop-cobalt branch of my fork located here: https://github.com/Phantom139/dask-jobqueue

Here are the primary differences I have noticed with Cobalt compared to some of the other clusters.

- First, when submitting a job on Cobalt (Might be an Argonne specific thing), the job id gets sent through stdout while additional job information is passed through stderr prompting a return code of 1 when calling **qsub**, which in the current implementation of _call() throws an exception.
- Second, Cobalt requires all job script files to be executable while the script only writes out a plaintext file for the job, this was fixed with a simple addition of a chmod +x call before calling self._submit_job() in the start_workers() definition
- I'm unsure if this is due to the newline terminator after the 'shebang', but for some reason Cobalt is not recognizing the headers created in the jobfile, so I added capability to instead attach the relevant parameters to the command line call of **qsub**.

I still need to test my code a bit more to verify things are working, and I have no idea if this will work for other Cobalt clusters, but it should be a start.

Phantom139:
Just an update on my current progress on implementing Cobalt support. I can get the job script to be generated, for my test application it looks like this:

```
#!/bin/bash

#COBALT -q debug-cache-quad
#COBALT -A climate_severe
#COBALT -t 60
#COBALT -n 8



/projects/climate_severe/Python/anaconda/bin/python3.7 -m distributed.cli.dask_worker tcp://140.221.69.22:39363 --nthreads 64 --memory-limit 100.00GB --name dask-worker--${JOB_ID}-- --death-timeout 60
```
This is being submitted to the cluster and the job does start, however there is no output from the job and it does not save the log files that Cobalt generated (JOBID.output & JOBID.error). The script that is using this cluster is set up like so:

```
		cluster = CobaltCluster(processes=1,
							   memory="100GB",
							   cores=64,
							   ncpus=8,
							   project="climate_severe",
							   walltime="60",
							   queue="debug-cache-quad")
		cluster.scale(1)
		dask_client = Client(cluster, timeout=60)
		calculation_future = dask_client.map(run_calculation_routines, fList)	
		wait(calculation_future) 
[...]

PR #302, opened 2019-07-26
Add a way to get network interfaces without ifconfig

lesteve:
Apparently on some clusters `ifconfig` is configured to need root access. This was mentioned by @Phantom139 in https://github.com/dask/dask-jobqueue/issues/298.


lesteve:
OK I am going to merge this one.

PR #303, opened 2019-07-29
Prepare for Dask 2.2 release

mrocklin:
This is just to test against new versions prior to a release upstream

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -550,3 +550,7 @@ class JobQueueCluster(ClusterManager):
 
     def worker_key(self, worker_state):
         return _job_id_from_worker_name(worker_state.name)
+
+    @property
+    def scheduler_comm(self):
+        return self.local_cluster.scheduler_comm
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -548,7 +548,8 @@ class JobQueueCluster(ClusterManager):
 
         return job_id
 
-    def worker_key(self, worker_state):
+    @staticmethod
+    def worker_key(worker_state):
         return _job_id_from_worker_name(worker_state.name)
 
     @property
--- a/dask_jobqueue/deploy/cluster_manager.py
+++ b/dask_jobqueue/deploy/cluster_manager.py
@@ -386,3 +386,7 @@ class ClusterManager(object):
             "attribute, needed for scaling with "
             "cores or memory kwargs.".format(self.__class__.__name__)
         )
+
+    @property
+    def loop(self):
+        return self.scheduler.loop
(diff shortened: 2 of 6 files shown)


mrocklin:
OK, this passes with both master and latest release.  I've changed the title.  I think that this is ready for review.

guillaumeeb:
Hm, this is difficult to review without context... There is little changes here, I suspect this is the minimum to make it work, but that we should rewrite a lot of the code base to comply with SpecCluster.

mrocklin:
>  I suspect this is the minimum to make it work

Right, the thing to notice here is the two green check CI marks at the end of the list of commits.  One is for using git master, and one is for using the current release (See the last commit, which transitions from git master back to current release).  If we don't merge this then dask-jobqueue will not be compatible with the next release.

> we should rewrite a lot of the code base to comply with SpecCluster.

I agree.  I'm changing SpecCluster now to include most of the functionality that dask-jobqueue has had to build on its own.  Hopefully the rewrite will be very simple once I'm done.  I think that that will happen after the Dask 2.2 release though, so we should merge and release this before the rewrite happens.

dask_jobqueue/core.py
```diff
@@ -548,5 +548,10 @@ def _job_id_from_submit_output(self, out):
 
         return job_id
 
-    def worker_key(self, worker_state):
+    @staticmethod
+    def worker_key(worker_state):
```
mrocklin: The method wasn't easy to serialize because the class has some unserializable elements.  Fortunately we don't need `self` here, so I changed this to a `staticmethod` in order to remove need to serialize the class.

dask_jobqueue/core.py
```diff
@@ -548,5 +548,10 @@ def _job_id_from_submit_output(self, out):
...
         return _job_id_from_worker_name(worker_state.name)
+
+    @property
+    def scheduler_comm(self):
+        return self.local_cluster.scheduler_comm
```
mrocklin: This is for compatibility with `SpecCluster`, which some elements, like adaptive, are starting to expect

dask_jobqueue/deploy/cluster_manager.py
```diff
@@ -131,7 +131,10 @@ def adapt(
             elif maximum_memory is not None:
                 kwargs["maximum"] = self._get_nb_workers_from_memory(maximum_memory)
         self._adaptive_options.update(kwargs)
-        self._adaptive = Adaptive(self.scheduler, self, **self._adaptive_options)
+        try:
+            self._adaptive = Adaptive(self.scheduler, self, **self._adaptive_options)
+        except Exception:
+            self._adaptive = Adaptive(self, **self._adaptive_options)
```
mrocklin: We dropped the need to explicitly pass in the scheduler in recent versions.  We will query the scheduler over a network connection instead so that it can be remote in the future.

guillaumeeb: That's perfect!

guillaumeeb:
Thanks for doing this!

Issue #205, opened 2018-11-27
LSF scheduler seems to have redundant memory keyword


josephhardinee:
It looks like specifying the lsf scheduler requires two memory keywords (mem, and memory). Should mem on the LSF scheduler be turned into memory so that it lines up with other schedulers and the core object?

josephhardinee:
Also it appears that memory should be of form '24 GB', but mem need to be passed as an integer number of bytes.

josephhardinee:
Okay, I don't have a strong preference either way except  that I got a little confused. 

Also I assume we would want the input form for the two to match (Using unit in the name as a string, vs passing an int for memory). To update this I assume we'd want to test the type for an int or str on mem keyword and do the conversion that way so as to not break previously existing scripts? Or would you rather it is just updated for both to take a str?

guillaumeeb:
Hm, I think it would be a minor breaking change and not impact a lot of people. I'd be more in favor of just update `mem` to be a string, and update the code using it.

guillaumeeb:
I think we can do two things to clarify this:
- Update the documentation and examples to explain that for each job queuing system, there are workers kwargs, and job related kwargs.
- For LSF, modify the ncpus and mem kwarg, prefixing them using `job_`.

PR #307, opened 2019-08-05
Rewrite dask-jobqueue with SpecCluster

mrocklin:
This is the first step towards rewriting with SpecCluster
I mostly copied the implementations from the Cluster classes,
but then removed the cluster bits

For people reviewing, I recommend looking at the tests first.


dask_jobqueue/tests/test_job.py
```diff
@@ -0,0 +1,19 @@
...
+def test_basic():
+    job = PBSJob(scheduler="127.0.0.1:12345")
+    assert "127.0.0.1:12345" in job.job_script()
+
+
+@pytest.mark.env("pbs")
+@pytest.mark.asyncio
+async def test_live():
+    async with Scheduler(port=0) as s:
+        async with PBSJob(s.address, name="foo") as job:
```
mrocklin: This currently depends on https://github.com/dask/distributed/pull/2927

lesteve:
For some reason, I could not get the PBS test to pass locally. I managed to get a very similar test passing with SGE though and I pushed into your PR branch, hope you don't mind.

mrocklin:
Thanks @lesteve !  I'm very glad to see this.  Both because it's always nice to have help (you're very welcome to push to this branch or any other I have) and because it's nice to see that this approach is accessible.

mrocklin:
I'll make PBSCluster/SGECluster functions next to get your feedback on how they feel.

mrocklin:
OK, I've added a rough draft using SpecCluster to build out a generic JobQueueCluster implementation.  It should compose well with the Job classes.  

It's currently tricky to pass around the keyword arguments correctly, but it's certainly less work than before.  If anyone has an opportunity to try this out on a real cluster and provide feedback, that would be welcome.

mrocklin:
OK, this use of job-name seems to work, at least for PBS.  I'm not sure what's up with SGE.

This change is because we are now using the `name=` attribute with spec-cluster, so it's a bit harder to also use it for dask-jobqueue.

lesteve:
I have looked at this quickly: at the moment in `JobQueueCluster`, the created `Job` instances do not know about the scheduler addresses (I am guessing this is done later as part of `SpecCluster` but I have not managed to figure out where).

```py
from dask_jobqueue import SGEJob
from dask_jobqueue.job import JobQueueCluster

env_extra = ['source /sequoia/data1/lesteve/miniconda3/etc/profile.d/conda.sh',
             'conda activate dask-dev',
            ]
resource_spec = 'h_vmem=10G,mem_req=500M'
queue = 'all.q'

cluster = JobQueueCluster(1, Job=SGEJob, queue=queue, env_extra=env_extra,
                     cores=1, processes=1,
                     memory='16GB',
                     resource_spec=resource_spec,
                     interface='ib0'
                    )
```

Log from the worker:
```
distributed.nanny - INFO -         Start Nanny at: 'tcp://10.149.0.3:54180'
distributed.diskutils - INFO - Found stale lock file and directory '/home/lesteve/dev/dask-jobqueue/worker-y509nqij', purging
distributed.diskutils - INFO - Found stale lock file and directory '/home/lesteve/dev/dask-jobqueue/worker-0ksc2mwf', purging
distributed.worker - INFO -       Start worker at:     tcp://10.149.0.3:60429
distributed.worker - INFO -          Listening to:     tcp://10.149.0.3:60429
distributed.worker - INFO -          dashboard at:           10.149.0.3:43263
distributed.worker - INFO - Waiting to connect to:                 tcp://None
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -               Threads:                          1
distributed.worker - INFO -                Memory:                   16.00 GB
distributed.worker - INFO -       Local Directory: /home/lesteve/dev/dask-jobqueue/worker-6to6xw20
distributed.worker - INFO - -------------------------------------------------
distributed.nanny - ERROR - Failed to start worker
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/nanny.py", line 674, in run
    await worker
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/worker.py", line 983, in start
    await self._register_with_scheduler()
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/worker.py", line 767, in _register_with_scheduler
    self.scheduler.address, connection_args=self.connection_args
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/core.py", line 215, in connect
    quiet_exceptions=EnvironmentError,
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/tcp.py", line 347, in connect
    ip, port = parse_host_port(address)
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/addressing.py", line 91, in parse_host_port
    port = _default()
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/addressing.py", line 70, in _default
    raise ValueError("missing port number in address %r" % (address,))
ValueError: missing port number in address 'None'
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x2aaab3b39e10>>, <Task finished coro=<Nanny._on_exit() done, defined at /home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/nanny.py:369> exception=ValueError("missing port number in address 'None'")>)
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/nanny.py", line 372, in _on_exit
    await self.scheduler.unregister(address=self.worker_address)
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/core.py", line 747, in send_recv_from_rpc
    comm = await self.pool.connect(self.addr)
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/core.py", line 874, in connect
    connection_args=self.connection_args,
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/core.py", line 215, in connect
    quiet_exceptions=EnvironmentError,
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/tcp.py", line 347, in connect
    ip, port = parse_host_port(address)
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/addressing.py", line 91, in parse_host_port
    port = _default()
  File "/home/lesteve/miniconda3/envs/dask-dev/lib/python3.7/site-packages/distributed/comm/addressing.py", line 70, in _default
    raise ValueError("missing port number in address %r" % (address,))
ValueError: missing port number in address 'None'
distributed.nanny - INFO - Closing Nanny at 'tcp://10.149.0.3:54180'
distributed.dask_worker - INFO - End worker
```

It looks like None is passed as the scheduler address and not changed later basically. I have no clue why the same issue would not happen for PBS ...

mrocklin:
SpecCluster starts the scheduler and then passes that address to each of the workers as they start as the first argument.

I think the problem is that we're not passing through `*args`

mrocklin:
Tests are now passing.  Thanks for pointing out that we weren't getting the scheduler address @lesteve , that was the issue.

mrocklin:
Some next steps:

1.  I need to unbreak the tests :)
2.  We should try this out on a real cluster to make sure that all of the UI elements of SpecCluster function well
3.  Increase testing
4.  Move the scheduler out to a Job

lesteve:
Just a note before I forget, we were doing some worker naming based on the job scheduler jobid with `dask-worker --name` and we are not doing it in the `Job` implementation.

IIRC this was used in `JobQueueSchedulerPlugin` for tracking pending jobs (submitted jobs but with workers that have not yet connected to the scheduler for example because they are waiting in the job scheduler queue). https://github.com/dask/dask-jobqueue/pull/63 probably gives more context.

mrocklin:
I'm aware.  I think that we still have this logic, but now it is handled
more generally.  SpecCluster maintains a set of workers that it wants to
exist (worker_spec) a set of workers that it has successfully submitted to
the resource manager (workers) and a set of workers that have arrived at
the scheduler (scheduler_info["workers"]).  This is used in adaptivity, see
https://github.com/dask/distributed/blob/master/distributed/deploy/adaptive_core.py

lesteve:
I see, great to hear that this is something that `SpecCluster` is addressing already.

mrocklin:
I stole the idea from dask-jobqueue really :)

We'll also need to steal `adapt(core=, memory=)` in order to maintain compatibility

lesteve:
Some comments I have that are not urgent at all:
- for debugging purposes it would be nice to have the equivalent of the old `cluster.job_script()` i.e. the script that was actually submitted to the job scheduler. At the moment `job.job_script()` is giving partial information (for example the scheduler address is None which can be confusing)
- from a user perspective, I reckon it is preferable to have a cluster class per job scheduler as before. Somehow:
```py
from dask_jobqueue import SGECluster
cluster = SGECluster(...)
```
feels easier than:
```py
from dask_jobqueue import JobQueueCluster, SGEJob
cluster = JobQueueCluster(Job=SGEJob, ...)
```

lesteve:
On a real cluster, the UI elements on Specluster looks fine. The only thing is that the link to the dashboard (proxy/8787/status appears black (rather than blue for the usual link colour) so that it is not obvious you can click on it:
image

lesteve:
One issue I found with `adaptive`:

```py
from dask_jobqueue import SGEJob
from dask_jobqueue.job import JobQueueCluster

env_extra = ['source /sequoia/data1/lesteve/miniconda3/etc/profile.d/conda.sh',
             'conda activate dask-dev',
            ]
resource_spec = 'h_vmem=10G,mem_req=500M'
queue = 'all.q'

cluster = JobQueueCluster(1, Job=SGEJob, queue=queue, env_extra=env_extra,
                     cores=1, processes=1,
                     memory='16GB',
                     resource_spec=resource_spec,
                     interface='ib0'
                    )

cluster.adapt(minimum=0, maximum=1)
```

The single worker I started when I created the cluster gets killed because of adaptive after a few seconds (that makes sense):
```
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']                                           
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Suggest closing workers: ['tcp://10.149.0.3:48873']
distributed.scheduler - INFO - Closing worker tcp://10.149.0.3:48873
distributed.scheduler - INFO - Remove worker tcp://10.149.0.3:48873
distributed.core - INFO - Removing comms to tcp://10.149.0.3:48873
distributed.scheduler - INFO - Lost all workers
distributed.deploy.adaptive - INFO - Retiring workers ['tcp://10.149.0.3:48873']
```

But then `cluster.workers` is not updated:
```py
print(cluster.workers)
```
```
{0: <SGEJob: status=running>}
```

`qstat` shows that the job was already `qdeled`.

Trying to forcibly kill it with `cluster.scale(0)`, I get this error (consistent with what I said above: the cluster thinks it still has a worker but the job has already been `qdel`ed):

```
Task exception was never retrieved
future: <Task finished coro=<Job.close() done, defined at /home/lesteve/dev/dask-jobqueue/dask_jobqueue/job.py:274> exception=RuntimeError('Command exited with non-zero exit code.\nExit code: 1\nCommand:\nqdel 7162332\nstdout:\ndenied: job "7162332" does not exist\n\nstderr:\n\n')>
  File "/home/lesteve/dev/dask-jobqueue/dask_jobqueue/job.py", line 277, in close
    self._call(shlex.split(self.cancel_command) + [self.job_id])
  File "/home/lesteve/dev/dask-jobqueue/dask_jobqueue/job.py", line 321, in _call
    "stderr:\n{}\n".format(proc.returncode, cmd_str, out, err)
RuntimeError: Command exited with non-zero exit code.
Exit code: 1
Command:
qdel 7162332
stdout:
denied: job "7162332" does not exist

stderr:

```

mrocklin:
Hrm, that must be an issue with the upstream adaptive code.  Is it also in `cluster.worker_spec`?

lesteve:
It is also in `cluster.worker_spec`:

```
cluster.worker_spec
```

```
{0: {'cls': dask_jobqueue.sge.SGEJob,
  'options': {'queue': 'all.q',
   'env_extra': ['source /sequoia/data1/lesteve/miniconda3/etc/profile.d/conda.sh',
    'conda activate dask-dev'],
   'cores': 1,
   'processes': 1,
   'memory': '16GB',
   'resource_spec': 'h_vmem=10G,mem_req=500M',
   'interface': 'ib0',
   'protocol': 'tcp://',
   'security': None}}}
```

mrocklin:
Regarding the comments in https://github.com/dask/dask-jobqueue/pull/307#issuecomment-519156296 I entirely agree that we will want `FooCluster` classes for the reasons you mention, and probably more.  I think that we are building up to that.  I'm curious how far we can build and test things centrally before we then expand out and add the final usability layer.

mrocklin:
Hrm, I tried to get this going locally, but I'm not sure that I still have hours on my HPC project (cc @jhamman )

@lesteve , if you're able to investigate further, this is the path that I would expect would lead to closing down the worker:

Adaptive decides to scale down workers

https://github.com/dask/distributed/blob/8ee867be3b0de454467f736963bf7e5501aa3815/distributed/deploy/adaptive_core.py#L193

Which then calls `cluster.scale_down`

https://github.com/dask/distributed/blob/8ee867be3b0de454467f736963bf7e5501aa3815/distributed/deploy/adaptive.py#L155

That deletes the worker from worker_spec and waits until things normalize

https://github.com/dask/distributed/blob/8ee867be3b0de454467f736963bf7e5501aa3815/distributed/deploy/spec.py#L374-L378

and waits until things normalize

https://github.com/dask/distributed/blob/8ee867be3b0de454467f736963bf7e5501aa3815/distributed/deploy/spec.py#L282

Which closes and deletes the workers

https://github.com/dask/distributed/blob/8ee867be3b0de454467f736963bf7e5501aa3815/distributed/deploy/spec.py#L251-L257

Clearly this isn't happening.  The question is where and why.  If you have time to dive in here that would be welcome.  It would also be a nice tour of the current scaling/adaptive logic.  It would be nice to have another pair of eyes on that code.  However, we're also well beyond the 1-2 hours per week goal that you set, so no worries if you're busy.

andersy005:
I tried reproducing this on a real cluster that uses PBS. 

```python
In [1]: from dask_jobqueue import PBSJob

In [2]: from dask_jobqueue.job import JobQueueCluster

In [3]: from dask.distributed import Client

In [4]: resource_spec='nodes=1:ppn=48'

In [5]: queue = "medium"

In [6]: cluster = JobQueueCluster(n_workers=1, Job=PBSJob, queue=queue, cores=1, processes=1, resource_spec=resource_spec, walltime="01:00:00")
```


```python
In [9]: cluster.adapt(minimum=0, maximum=1)
Out[9]: <distributed.deploy.adaptive.Adaptive at 0x7f9a3df3c668>
```

 At first I noticed the exact behavior pointed out by @lesteve:

```python
In [11]: cluster.workers
Out[11]: {0: <PBSJob: status=running>}

In [12]: cluster.worker_spec
Out[12]:
{0: {'cls': dask_jobqueue.pbs.PBSJob,
  'options': {'queue': 'medium',
   'cores': 1,
   'processes': 1,
   'resource_spec': 'nodes=1:ppn=48',
   'walltime': '01:00:00',
   'interface': None,
   'protocol': 'tcp://',
   'security': None}}}

```

However, when I checked back (roughly 1 min after calling `.adapt()`), I noticed that cluster worker_spec was up-to-date:


_After ~ 1 minute_


```python

In [19]: cluster.worker_spec
Out[19]: {}

In [20]: cluster.workers
Out[20]: {}
```

lesteve:
Thanks a lot for the links @mrocklin! I am afraid that I won't be able to look at this for a week or so.

lesteve:
@mrocklin I should have a bit of time this week to look at this. Let me know if/how I can help.

mrocklin:
I'm also trying to focus on this later today, and I hope to engage some
people at the Pangeo meeting happening later this week.

Hopefully later today I'll have a better sense on some short-term tasks.
Until then, I recommend just using it and finding issues.

mrocklin:
As an experiment, I've removed the old `SLURMCluster` and replaced it with a `functools.partial(JobQueueCluster, Job=SLURMJob)`.  Mostly I did this to see what would need to change in tests.  Mostly things were fine.

mrocklin:
OK, I'm confused about why the SLURM tests continue to fail.  Things work well when I test against a live SLURM cluster.  I haven't set up a Docker environment though to try to reproduce the failures there (currently travelling with only an OSX machine).

mrocklin:
I think that other than the SLURM errors the next step here might be to remove the old `FOOCluster` classes, and see if we can pass the current test suites with a minimum of modification.  When I removed the SLURMCluster class I did have to remove some lines that were highly specific to the old implementation.  But otherwise things seemed to work well.

I think that after that we might try to improve the usability of the FOOCluster classes, breaking the simple implementaitons of

```python
SLURMCluster = functools.partial(JobQueueCluster, Job=SLURMJob)
```

I'm going to add `scale(memory=..., cores=...)` upstream.

mrocklin:
I've moved cores and memory upstream so hopefully we can remove that code safely.

https://github.com/dask/distributed/pull/2974

dask_jobqueue/job.py
```diff
@@ -0,0 +1,394 @@
...
+from distributed.deploy.spec import ProcessInterface, SpecCluster
+from distributed.scheduler import Scheduler
+
+from distributed.utils import format_bytes, parse_bytes, tmpfile, get_ip_interface
+
+logger = logging.getLogger(__name__)
+
+
+class Job(ProcessInterface):
+    """ Base class to launch Dask workers on Job queues
```
guillaumeeb: We should change the docstring here and below, shouldn't we? I guess this is already planned.

dask_jobqueue/job.py
```diff
@@ -0,0 +1,394 @@
...
+                "protocol": protocol,
+                "interface": interface,
+                "dashboard_address": dashboard_address,
+                "security": security,
+            },
+        }
+        kwargs["interface"] = interface
+        kwargs["protocol"] = protocol
+        kwargs["security"] = security
+        worker = {"cls": Job, "options": kwargs}
```
guillaumeeb: That's probably here we should add other options needed by adaptive, e.g. cores and memory (see https://github.com/dask/distributed/pull/2974).

dask_jobqueue/job.py
```diff
@@ -0,0 +1,394 @@
...
+        self.example_job = Job("tcp://scheduler:8786", name="name", **kwargs)
```
guillaumeeb: Is the hardcoded port number temporary?

mrocklin:
> So finaly I've tested this on our PBS cluster. Mostly it works fine. 

Thank you for testing this.  Also I hope that you had a nice vacation!

> The principal question I have is how will we handle job killed by external means: either reaching walltime, or manualy deleted.

@jacobtomlinson also ran into this in cloud deployments and raised an issue here: https://github.com/dask/distributed/issues/2968

I put up some links to code and a possible solution there.  I think that it's probably an accessible problem if anyone wants to try contributing upstream to get some more familiarity with SpecCluster.  I can handle it though.  No pressure.

guillaumeeb:
Hm, I think we will have trouble when using `processes` > 1.

`cluster.worker_spec`:
```
{0: {'cls': dask_jobqueue.pbs.PBSJob,
  'options': {'processes': 2,
   'cores': 4,
   'memory': '20Gb',
   'walltime': '01:00:00',
   'interface': None,
   'protocol': 'tcp://',
   'security': None}},
```

`cluster.scheduler_info["workers"]`:
```
{'tcp://10.xxx.yyyy.zzz:44589': {'type': 'Worker',
  'id': '1-1',
  'host': '10.xxx.yyyy.zzz',
  'resources': {},
  'local_directory': '/tmp/pbs.7781044.admin01/worker-c7wa09q_',
  'name': '1-1',
  'nthreads': 2,
  'memory_limit': 10000000000,
  'services': {'dashboard': 46426},
  'nanny': 'tcp://10.xxx.yyyy.zzz:37393'},
 'tcp://10.xxx.yyyy.zzz:41674': {'type': 'Worker',
  'id': '1-0',
  'host': '10.xxx.yyyy.zzz',
  'resources': {},
  'local_directory': '/tmp/pbs.7781046.admin01/worker-7kva_j1g',
  'name': '1-0',
  'nthreads': 2,
  'memory_limit': 10000000000,
  'services': {'dashboard': 33210},
  'nanny': 'tcp://10.xxx.yyyy.zzz:39267'},
```

So our worker_spec is a job_spec. And the scale argument is actually the number of jobs, and not the number of resulting workers anymore. This also leads to weird displaying of the widget too:
image

I encountered this while working on https://github.com/dask/distributed/pull/299.

guillaumeeb:
> One solution would be to include the same job many times in the worker spec, and ensure that starting and closing a job is idempotent.

> An alternative would be to break the 1-1 mapping between workers in the spec and workers that the scheduler sees, but I'm hoping that the solution above is simpller.

From what I've seen of SpecCluster, the first solution is probably easier even if not trivial.

> I also don't know if this is likely to have any unintended consequences, 

I really don't know!

mrocklin:
I'm confused about why the PBS and SLURM tests are hanging (the SLURM tests in particular).  If anyone has any interest in spinning up a docker environment that doing some testing that would be welcome.

mrocklin:
That worked!  I'm so happy!

lesteve:
I debugged the problem in the SLURM docker image and here is what I found:
* if you run `sbatch /tmp/test.sh` from a folder that does not exist in the worker node (e.g. `/dask-jobqueue`), the job terminates instantly. Somehow this seems related to the SLURM logging mechanism because if I set -e and -o to a folder that exists on the worker node the job runs fine.
* it does not make any sense to me why the behaviour would be different in master vs this PR (note that I am using the exact same docker images for master and this PR, the only difference is `/dask-jobqueue` which is a docker volume pointing to the local dask-jobqueue git checkout). The only thing I can think of, is that somewhere we do a `cd` or equivalent in master (or something has a side-effect that does a `cd`).

mrocklin:
OK, some TODOs

- Work on @guillaumeeb 's PR here to identify workers that have been killed: https://github.com/dask/distributed/pull/2990
-  Move over the remaining resource managers, like HTCondor and OAR
-  Improve docstrings and other UI elements
-  Fix up PBS tests

I've been using this pretty happily over the last week on a few different systems (I'm ssh'ed into three HPC centers right now :) )

mrocklin:
Interestingly, the PBS failure is happening on `LocalCluster`, not `PBSCluster`

lesteve:
`test_scale_group` is still going to fail but this is on purpose. I think there is a problem with the scaling of "grouped" workers:

```py

cluster = PBSCluster(
        walltime="00:02:00",
        processes=2,
        cores=2,
        memory="2GB",
        local_directory="/tmp",
        job_extra=["-V"],
    )

# scale is in term of dask workers and I want 4 of them
# this should start 2 jobs (since each job has 2 workers each) but it starts 4 jobs instead
cluster.scale(4)

client = Client(cluster)
len(client.scheduler_info()['workers']) # should be 4 and I get 8 instead.
```

mrocklin:
I suggest that we change the meaning of first argument of `scale` back to refer the number of jobs rather than the number of processes.  We might also change our documentation to use `cores=` and `memory=` more frequently in order to avoid the job/proocess ambiguity.

cc'ing @lesteve @guillaumeeb @jhamman , who I think were all involved in the original decision to make scale refer to processes rather than jobs.

mrocklin:
Green check mark.  Thank you for fixing CI @lesteve !

mrocklin:
OK, I've moved over OAR and HTCondor.  I've also removed most of the old code that is no longer being used (this PR now has more deletions than additions, which is nice).  

Now might be a good time for another round of critical reviews from the core maintainers to make sure that we're ok with this architecture.

If things are mostly ok then we should probably think about docstrings and other UI elements.

I will personally focus on https://github.com/dask/distributed/pull/2990 next

andersy005:
> I suggest that we change the meaning of first argument of scale back to refer the number of jobs rather than the number of processes. We might also change our documentation to use cores= and memory= more frequently in order to avoid the job/proocess ambiguity.

👍

lesteve:
> I suggest that we change the meaning of first argument of scale back to refer the number of jobs rather than the number of processes. We might also change our documentation to use cores= and memory= more frequently in order to avoid the job/proocess ambiguity.

I am a bit torn about changing `.scale` argument semantic:
1. if you know dask, you expect `.scale` to be in terms of dask workers, e.g. to be similar to `LocalCluster`.
2. if you are familiar with your HPC cluster and jobs, I agree `.scale` in terms of job to launch is easier to understand

I am slightly leaning towards 1. for consistency reasons between `LocalCluster` and dask-jobqueue `*Cluster` classes.

I completely agree that using `cores=` everywhere in our doc would lead to less confusion.

mrocklin:
When I show dask-jobqueue to people we first start by populating the keyword arguments around cores, memory, processes, and so on.  Then we print out the job script and they say "ahh".   Then I say "now use scale to ask for `n` jobs" and it seems to make sense to them.  The people that I talk to seem to think less about the number of dask workers that are going to be present, but how many nodes they are using.  How these nodes are cut up into Dask workers is an inconsequential detail to them.

This is just my own experience though.  I suspect that other people have other experiences.

guillaumeeb:
> if you are familiar with your HPC cluster and jobs, I agree .scale in terms of job to launch is easier to understand

>  Then I say "now use scale to ask for n jobs" and it seems to make sense to them

Yep, these is true, I always had trouble explaining the "number of worker processes" argument of scale. 
But this also means I'll have to explain that this is not true anymore 😭, and that this is as they thought in the first place.

In the end, if this is simpler to implement SpecCluster with scale based on job number, I say let's go for it.

guillaumeeb:
So stopped or killed jobs seems to be handle correctly thanks to dask/distributed#2990. But we still have the problem of jobs with multiple processes, that should be fixed if we switch to scaling with number of jobs instead of processes.

I've encountered one race condition though, when calling scale multiple times, sometimes down, sometimes up. I can reproduce it this way:
```py
cluster = JobQueueCluster(Job=PBSJob, processes=1, cores=4, memory='20Gb', walltime="01:00:00")
cluster.scale(6)
#wait for workers to come up
cluster.scale(2)
# wait for scaling down (fast)
cluster.scale(6)
#Only 5 workers are up
```
What I observed with PBS is that on the last scale call, 6 jobs are effectively running at first, but after a few seconds, one is killed.... Not sure if it comes from dask/distributed#2990 or something else... Hm, looking closely, I've got a 
```
tornado.util.TimeoutError: Timed out starting worker
```

Never experienced this before though...

mrocklin:
Here is my attempt to reproduce locally (also, meet `dask_jobqueue.LocalCluster`)

```python
In [1]: from dask_jobqueue import LocalCluster

In [2]: cluster = LocalCluster(cores=2, memory="4GB")

In [3]: from dask.distributed import Client

In [4]: client = Client(cluster); client
Out[4]: <Client: 'tcp://172.18.46.217:63488' processes=0 threads=0, memory=0 B>

In [5]: cluster.scale(6)

In [6]: client
Out[6]: <Client: 'tcp://172.18.46.217:63488' processes=3 threads=6, memory=12.00 GB>

In [7]: client
Out[7]: <Client: 'tcp://172.18.46.217:63488' processes=6 threads=12, memory=24.00 GB>

In [8]: cluster.scale(2)

In [9]: client
Out[9]: <Client: 'tcp://172.18.46.217:63488' processes=2 threads=4, memory=8.00 GB>

In [10]: cluster.scale(6)

In [11]: client
Out[11]: <Client: 'tcp://172.18.46.217:63488' processes=2 threads=4, memory=8.00 GB>

In [12]: client
Out[12]: <Client: 'tcp://172.18.46.217:63488' processes=5 threads=10, memory=20.00 GB>

In [13]: client
Out[13]: <Client: 'tcp://172.18.46.217:63488' processes=6 threads=12, memory=24.00 GB>
```

mrocklin:
Does anyone have suggestions on how we handle docstrings and cluster classes?

Previously we had a docstring on the JobQueueCluster class, with additional docstrings on each of the FooCluster classes.  We would automatically copy over the JobQueueCluster docstring to the FooCluster.

Now things are a bit more complex, we have both `FooJob` and `FooCluster` classes.  I'll take a look at this tomorrow and see if I can come up with something similar to what we have now.  However, if anyone has suggestions then please say something.

lesteve:
I haven't played too much with `docrep` so I don't know how far we can get with it. If it turns out to be too complicated, I would say the main thing is to have good docstrings for for `FooCluster` which are the one that the users will use. The `FooJob` classes are more implementation details.

mrocklin:
I've made classes, updated docstrings, removed core.py and docrep, and applied linters.

I had to fix something upstream (CI will likely fail until that gets in) https://github.com/dask/distributed/pull/3037

I'm removing the WIP label.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -247,26 +236,19 @@ def test_adaptive_grouped(loop):
...
-
-            start = time()
-            processes = cluster.worker_processes
+            processes = cluster.example_job.worker_processes
             while len(client.scheduler_info()["workers"]) != processes:
                 sleep(0.1)
                 assert time() < start + QUEUE_WAIT
 
 
+@pytest.mark.xfail(reason="adapt doesn't yet have cores/memory")
```
mrocklin: This is a regression until we move this into core.  Are we ok with losing this functionality short term?

cc @guillaumeeb

mrocklin:
I wouldn't mind merging this soon.  It probably needs review before that though.  If anyone is able to help out with this that would be welcome.

lesteve:
I'll try to look at this in more details tomorrow (European time). If you have anything in particular you think I should pay more attention to, let me know!

Here are a few comments from a cursory look:
* I seem to remember there were a few TODOs left on the part I did modify. The LocalJob one is probably fine. About the `cluster.running_jobs` (and others like `cluster.pending_jobs`), disappearing, I am wondering whether there is a way to have an equivalent information at the `distributed` level, maybe using `group` from the worker spec?  
* The tests modifications were a bit ad-hoc in places. In particular `client.wait_for_workers` is not great because the test will hang and no other tests will run. In an ideal world client.wait_for_workers would have a timeout (or we can go back to the previous assert + time.sleep previous solution). Other than this, the fact that the tests need to be run from a given folder (SLURM and PBS IIRC) is not great.

More generally with a reorganization like this, I would be more comfortable if at least one person per cluster class was able to give it a go on a real cluster before we do a new release. Just for completeness, it is fine to have this PR merged in master without this thorough testing.

mrocklin:
> About the cluster.running_jobs (and others like cluster.pending_jobs), disappearing, I am wondering whether there is a way to have an equivalent information at the distributed level, maybe using group from the worker spec?

Currently there is this, which is I think the same.  We could change names here though if you have suggestions.

https://github.com/dask/distributed/blob/fb733478bd30d71f0889119ec3859993dfa0315f/distributed/deploy/cluster.py#L379-L389

> In an ideal world client.wait_for_workers would have a timeout

That seems doable and useful regardless.  The testing decorator `@gen_cluster` also has a timeout builtin, which is very useful for problems like this.  I wonder if we should replace the `@pytest.mark.asyncio` decorator with something that includes a timeout. 

> I would be more comfortable if at least one person per cluster class was able to give it a go on a real cluster before we do a new release.

In general I agree.  I suspect that for some of the less common managers this won't happen soon, so I'd like to avoid them.  I have personally run with PBS, SLURM, and LSF, although not with the latest commit.  The clusters that I am running on also may not be representative.  It would be good to have others test.

dask_jobqueue/tests/test_pbs.py
```diff
@@ -315,17 +295,18 @@ def test_scale_grouped(loop):
             cluster.scale(4)  # Start 2 jobs
 
             start = time()
-            while len(cluster.running_jobs) != 2:
-                sleep(0.100)
-                assert time() < start + QUEUE_WAIT
+            # TODO: Is there a replacement to check for number of jobs (rather than workers)
+            # while len(cluster.running_jobs) != 2:
+            #     sleep(0.100)
+            #     assert time() < start + QUEUE_WAIT
```
mrocklin: The number of total submitted jobs is `cluster.workers`.  

We could check the names of all workers that have checked in, remove `-0` from them

```python
len({w["name"]split("-")[0] for w in self.scheduler_info.values()})
```

dask_jobqueue/tests/test_sge.py
```diff
@@ -107,19 +107,23 @@ def test_complex_cancel_command(loop):
...
+            start = time()
+            while not client.scheduler_info()["workers"]:
+                sleep(0.100)
+                assert time() < start + QUEUE_WAIT
 
-        cluster.stop_all_jobs()
+            # TODO: Is there a replacement for .stop_all_jobs? stop_all_jobs
+            # does make sure that the pending jobs get qdeled.
+            # cluster.stop_all_jobs()
+            cluster.scale(0)
```
mrocklin: Does this not do that?

lesteve: I think what I had in mind at the timem was that `cluster.stop_all_jobs()` is useful in itself: 
* it will `qdel` pending jobs (i.e. jobs that have been submitted but whose dask workers have not had time to connect to the dask scheduler, for example because the jobs are still sitting in the queue)
* this is useful if you want to cleanly remove all jobs that were created from the `cluster` object. Yes all the jobs will eventually be in running state and terminate after some time if the scheduler is gone (through the `death-timeout` mechanism). Sometimes it feels nicer not to have an explicit way of cleaning up.
* there may be a replacement for `cluster.stop_all_jobs()` that I am not aware of.


To be explicit: `cluster.scale(0)` will `qdel` only the jobs whose dask workers have connected to the dask scheduler.

mrocklin:
> To be explicit: cluster.scale(0) will qdel only the jobs whose dask workers have connected to the dask scheduler.

I don't think that this is true, or at least it is my intention that it is not true.

scale(0) will remove everything from `worker_spec`, and then ask the SpecCluster to make that happen.  It will first as the scheduler to politely close all active workers.  Then it will call `await job.close()` on everything in `cluster.workers`.  The `.close` method calls `qdel` (or equivalent).

lesteve:
> I don't think that this is true, or at least it is my intention that it is not true.

Great to here, I am going to blame my limited understanding of the dask internals. I'll double check that tomorrow.

mrocklin:
It may be true, and it may be that things aren't working correctly.  We should expect many of these errors.  But if it isn't true then I would prefer to fix it in `scale` rather than have a separate `stop_all_jobs` method.

lesteve:
I can confirm that `cluster.scale(0)` deletes pending jobs if there are any.

I pushed some small tweaks with some TODOs. Sorry I should have added inline comments but I won't have time to do it today.

Something I noticed is that forcibly killing a job (on SGE with `qdel`), `cluster.workers` is empty (after some time) thanks to https://github.com/dask/distributed/pull/2990 but `cluster.worker_spec` will not be empty (this may be intentional for example see the end of https://github.com/dask/distributed/pull/2990#issue-310385863). I have to say I haven't fully understood whether `.worker_spec` is supposed to be the desired state or the actual state.

Generally speaking the organization with `FooJob` and `FooCluster(Job=FooJob)` seems fine 
I think the way docstrings are handled is fine too (I was never a big fan of `docrep` I have to admit).

Some nitpicks (that I think are not in the TODO inside the code):
- maybe I would use `job_cls` (or better name) rather than `Job` for the parameter name
- I am not convinced `{jobs}` and `{cluster}` in the docstring template should have an empty line separating them (I seem to remember weird edge cases in the past in numpydoc that was not very happy with empty lines between parameters)

mrocklin:
> I can confirm that cluster.scale(0) deletes pending jobs if there are any.

Great.  I'm relieved to hear it

> I pushed some small tweaks with some TODOs. Sorry I should have added inline comments but I won't have time to do it today.

Thank you.  I appreciate your time here.

> Something I noticed is that forcibly killing a job (on SGE with qdel), cluster.workers is empty (after some time) thanks to dask/distributed#2990 but cluster.worker_spec will not be empty (this may be intentional for example see the end of dask/distributed#2990 (comment)). I have to say I haven't fully understood whether .worker_spec is supposed to be the desired state or the actual state.

worker_spec is desired state.  If we `await cluster` then it tries to correct things.  So far I haven't turned this on to automatically update though.  We can do that easily, but it's a policy decision that we should talk about.  I was planning to ignore this question for now.

> Generally speaking the organization with FooJob and FooCluster(Job=FooJob) seems fine.
I think the way docstrings are handled is fine too (I was never a big fan of docrep I have to admit).

Same.  I prefer the low-tech solution.

> Some nitpicks (that I think are not in the TODO inside the code):

> maybe I would use job_cls (or better name) rather than Job for the parameter name
I am not convinced {jobs} and {cluster} in the docstring template should have an empty line separating them (I seem to remember weird edge cases in the past in numpydoc that was not very happy with empty lines between parameters)

These sound reasonable to me.  I'll make these changes.

mrocklin:
cc @rjzamora you may find this PR interesting

dask_jobqueue/tests/test_pbs.py
```diff
@@ -336,23 +317,29 @@ def test_scale_grouped(loop):
...
 
             start = time()
-            while len(cluster.running_jobs) != 1:
+            # TODO
+            # while len(cluster.running_jobs) != 1:
+            #     sleep(0.100)
+            #     assert time() < start + QUEUE_WAIT
+
+            # assert len(cluster.running_jobs) == 1
+            # workers = list(client.scheduler_info()["workers"].values())
```
mrocklin: Do we need to make this change, or does the check below suffice?

dask_jobqueue/tests/test_sge.py
mrocklin: I think we learned that this was fine. Resolved.

dask_jobqueue/local.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _submit_job(self, script_filename):
+        # Should we make this async friendly?
+        with open(script_filename) as f:
+            text = f.read().strip().split()
+        self.process = subprocess.Popen(
+            text, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        # TODO this should raise if self.process.returncode != 0. Refactor
+        # Job._call to be able to return process (so that we can return self.process.pid below)
```
mrocklin: In the `LocalJob` case things are a little different.  Normally our command just calls `qsub` to submit a job, and then returns quickly.  In this case the process actually runs `dask-worker`, so it doesn't return until the worker goes away.

dask_jobqueue/tests/test_pbs.py
mrocklin: I'm inclined to just not bother here though.  I think that things will be ok without this.

dask_jobqueue/local.py
lesteve: I realised this afterwards. The error I had to  debug in the PBS case (https://github.com/dask/dask-jobqueue/pull/307#issuecomment-526649040 failure with LocalJob in the PBS case) was because basically starting `dask-worker` would fail with an error (the test is run as `pbsuser` and can does to create the `global.lock` or the `worker-bla` folder). It would be great if there was a way to fail early in this case.

A few alternatives I can see, not sure which one is the less fragile:
* read a few more lines from `self.process.stderr` and raise an exception if ERROR is in the stderr
* read a few more lines from `self.process.stderr` and raise an exception if "Start worker at" is not in the stderr
* wait for some time, say 1s, poll and if the process has finished raise an exception with the full stderr.

dask_jobqueue/__init__.py
```diff
@@ -1,13 +1,14 @@
...
 from . import config
-from .core import JobQueueCluster
-from .moab import MoabCluster
-from .pbs import PBSCluster
-from .slurm import SLURMCluster
-from .sge import SGECluster
-from .lsf import LSFCluster
-from .oar import OARCluster
-from .htcondor import HTCondorCluster
+from .local import LocalJob, LocalCluster
```
lesteve: I would rather not have `LocalCluster` and `LocalJob` available at the root package level. That could create some confusion with `distributed.LocalCluster`. Do you see a use case for dask_jobqueue.local.LocalCluster outside of tests?

I am slightly inclined to also leave all the job classes outside of `dask_jobqueue` namespace as well since they are implementation details. Also slightly annoying to have tab-completion not work as well:
```py
# this gives two solutions: SGEJob and SGECluster instead of immediately completing
from dask_jobqueue import SGE<TAB>
```

If you agree this is something I can definitely do myself when I am feeling not too productive.

lesteve:
I pushed a few changes, you can review them [here](https://github.com/dask/dask-jobqueue/pull/307/files/7d13e6ba58321aa17bcb142f8948be1ef58d76de..ccfe946523a8cf92b316f84576aad8bf7210fe0d)

I think it is fine to have this merged in master and have people test more extensively on their clusters when it is in master.

I'd like to get these points sorted out before the PR can be merged:
- I would rename job.py to core.py. `Job` and `JobQueueCluster` are the core classes in `dask-jobqueue` and it's a bit weird to have `JobQueueCluster` in the `job` module.
- there are ad-hoc changes in tests that ideally we need to remove (e.g. the commented out `job_extra=["-D"]` in `test_slurm.py`
- I'd like to undersand `.example_job` in `JobQueueCluster` more, I'll put an inline comment there with my questions
- have some changelog entries with breaking changes. So far I have: `stop_all_jobs` -> `.scale(0)`, `name` -> `job_name`, `cluster.running_jobs`, `cluster.pending_jobs`, `cluster.canceled_jobs`. There may be mores

These points I think would be nice to have but I am fine with doing them after the PR is merged:
- it would be great if we could find a way not to have to `cd` in a different folder (jobqueue scheduler-specific, mostly PBS and SLURM). In an ideal world the `/dask-jobqueue` volumes would be mounted on the worker images I think but also `pbsuser` would have the right to write into `/dask-jobqueue`. Not sure how easy it is to do the second part in docker.
- xfail for `.adapt(cores=...)` is OK I think as long as we do remember to fix the issue in distributed before the next release

dask_jobqueue/job.py
```diff
@@ -0,0 +1,480 @@
...
+        try:
+            address = self.scheduler.address
+        except AttributeError:
+            address = "tcp://scheduler:8786"
+        return self.job_cls(
+            address or "tcp://scheduler:8786", name="name", **self._kwargs
+        )
+
+    @property
+    def job_header(self):
```
lesteve: I think `.job_header` is overriden in all the derived class I looked at. But maybe this is for when you use `JobQueueCluster(job_cls=MyCustomClass)`. Maybe this is the same thing for `job_name` below?

dask_jobqueue/job.py
```diff
@@ -0,0 +1,480 @@
...
+
+        if n_workers:
+            self.scale(n_workers)
+
+    @property
+    def example_job(self):
+        try:
+            address = self.scheduler.address
+        except AttributeError:
+            address = "tcp://scheduler:8786"
```
lesteve: I don't really understand the case where `self.scheduler.address` does not exist ...

dask_jobqueue/job.py
```diff
@@ -0,0 +1,480 @@
...
+            kwargs["config_name"] = config_name
+        kwargs["interface"] = interface
+        kwargs["protocol"] = protocol
+        kwargs["security"] = security
+        self._kwargs = kwargs
+        worker = {"cls": self.job_cls, "options": kwargs}
+        if "processes" in kwargs and kwargs["processes"] > 1:
+            worker["group"] = ["-" + str(i) for i in range(kwargs["processes"])]
+
+        self.example_job  # trigger property to ensure that the job is valid
```
lesteve: I have not completely understood the magic behind `.example_job` yet ... I understand its main role is to implement `cluster.job_script` and you need to have the scheduler address for `.cluster_script` to be correct. Is there more to understand?

dask_jobqueue/job.py
mrocklin: This is in `JobQueueCluster`, not `Job`.  This is mostly here for backwards compatibility.

dask_jobqueue/job.py
mrocklin: Not really.  Users today are trained to look, at `job_script`.  We want to make it easy to print this out.  All of the logic for this is actually within the Job class now, so we want an easy way to generate a Job object.  

In this particular line we want to create a Job with all of our keywords because that is where all of the error handling and informative error messages are.  However, also at this point, we don't yet know the scheduler address, so we have to place in a dummy address.

dask_jobqueue/job.py
mrocklin: When we first construct the cluster, and have only called `__init__` we have not yet awaited the scheduler, so the address is not yet known.

lesteve:
I thought a bit more about `.scale` semantic and whether it should be in terms of dask worker processes (basically `len(cluster.scheduler_info['workers'])`) or jobs (in other words what we called at one point "grouped workers", i.e. when you do `dask-worker --nprocs 4` you have one "grouped worker" but 4 worker processes). The issue was raised in https://github.com/dask/dask-jobqueue/pull/307#issuecomment-526783524 and discussed in a few following comments.

I think it is fine to leave it as it is in this PR (and have an entry in the changelog about this).

Here is a snippet to show the behaviour, everything seems to work in terms of "grouped worker" (maybe it is obvious for everyone else, but it was not for me):
```py
import time
from dask_jobqueue.local import LocalCluster

# one grouped worker
cluster = LocalCluster(n_workers=1, cores=6, processes=2, memory='1GB')
time.sleep(3)
print('one "grouped worker"')
print('cluster:\n', cluster)
print('Number of "grouped workers":', len(cluster.workers))
print('Number of worker processes:', len(cluster.scheduler_info['workers']))

print('scaling to two (grouped) workers')
cluster.scale(2)
time.sleep(3)
print('cluster:\n', cluster)
print('Number of "grouped workers":', len(cluster.workers))
print('Number of worker processes:', len(cluster.scheduler_info['workers']))
```

Output:
```
one "grouped worker"
cluster:
 LocalCluster('tcp://10.200.23.86:46275', workers=1, threads=6, memory=999.99 MB)
Number of "grouped workers": 1
Number of worker processes: 3
scaling to two "grouped workers"
cluster:
 LocalCluster('tcp://10.200.23.86:46275', workers=2, threads=12, memory=2.00 GB)
Number of "grouped workers": 2
Number of worker processes: 6
```

lesteve:
Having said that, I have two comments:
* there will places where `workers` is a bit ambiguous (difference between "grouped workers" vs worker processes). For example, with this PR, everything will work in terms of "grouped workers" for `dask_jobqueue.FooCluster`. On the other hand, `Client.wait_for_workers` seems to work in terms of dask worker processes (quickly looking at the source code, it uses the equivalent of `len(client.scheduler_info()['workers'])` AFAICT)
* the cluster uses 3 threads per process, but it is not that obvious from the `__repr__` which says `workers=2, threads=12`

Note: I am using distributed from master for the two comments above which may affect some of the things I am saying.

mrocklin:
I've fixed a couple of things upstream based on comments here: 

1.  `cluster.adapt(maximum_memory="1 TB")`: https://github.com/dask/distributed/pull/3057
2.  Use the number of reported dask workers rather than the length of `cluster.workers` for the `workers=` value in Cluster.repr. https://github.com/dask/distributed/pull/3058

dask_jobqueue/job.py
mrocklin: Is this ok with you @lesteve ?  Should I hit resolve here?  

Is there some comment I could add to the code that would make this more clear?

mrocklin:
I've added a changelog entry that includes context and a list of breaking changes.

dask_jobqueue/job.py
lesteve: Some suggestions:
* rename the variable to `._dummy_job` (better name more than welcome!) with a comment above saying that it is mostly to implement `JobQueueCluster.job_script`
* add a comment in `._dummy_job` about the `AttributeError` case (more or less what you replied in https://github.com/dask/dask-jobqueue/pull/307#discussion_r324386131). I still haven't fully understood your comment but mentioning that this is related to async may help

lesteve:
> Use the number of reported dask workers rather than the length of cluster.workers for the workers= value in Cluster.repr. dask/distributed#3058

https://github.com/dask/distributed/pull/3057 completely makes sense and fits my understanding of what is understood by "workers=4" inside Dask i.e. "worker processes=4".

But if https://github.com/dask/distributed/pull/3057 is merged (I definitely think it should be merged) the discrepancy between "grouped workers" and "worker process" will be more obvious inside `dask-jobqueue` the output of my previous snippet (https://github.com/dask/dask-jobqueue/pull/307#issuecomment-531775753) will be confusing. In particular:

```py
cluster = LocalCluster(n_workers=1, cores=6, processes=3, memory='1GB')

# If https://github.com/dask/distributed/pull/3057 is merged
# the next line will show workers=3 although I just asked n_workers=1 above
print('cluster:\n', cluster)
```

All in all I am still torn about the whole thing (as you can probably guess from my back and forth on the topic). The simplest to me still feels that we should stick to "workers" means "worker process" everywhere in Dask, and encourage people (e.g. through `dask-jobqueue` documentation) to use `.scale(cores=...)` so that it is less confusing for cluster users who think in terms of numbers of job.

An improvement we could make (although maybe for later so as not to drag this PR for too much longer), would be to have an additional `n_jobs` argument at the `dask-jobqueue` level, for example in `JobQueueCluster.__init__` and `JobQueueCluster.scale` and related arguments `min_jobs`/`max_jobs` in  `JobQueueCluster.adapt`. But then you'd probably want to change the `JobQueueCluster` widget to add a `n_jobs` there as well and there may be additional things I missed. My main concern going down this route is that we'd start having some specific `dask-jobqueue` tweaks on top of `SpecCluster`. It is extremely convenient for maintenance reason to have as little of these tweaks as possible, as recent history of `distributed`/`dask-jobqueue` has shown.

mrocklin:
> All in all I am still torn about the whole thing 

Yes, I think that we'll probably end up running into usability problems no matter what we do.

@lesteve is what we have here good enough, or would you like for us to revert `scale(5)` back to meaning 5 workers?  (I personally prefer what we have now, but am flexible).

lesteve:
> @lesteve is what we have here good enough, or would you like for us to revert scale(5) back to meaning 5 workers?

I'd rather revert to `.scale(5)` meaning 5 workers and using `.scale(cores=...)` everywhere in the doc to reduce confusion.

mrocklin:
OK, I can change that back if desired.  Anecdotally I've found that it's easier to explain scale in terms of jobs, but I can understand that it might make things inconsistent.

cc'ing @guillaumeeb on this for his perspective.

guillaumeeb:
As said in https://github.com/dask/dask-jobqueue/pull/307#issuecomment-527181117, I think I'm currently leaning more towards `scale()` based on number of jobs rather than number of worker processes. This is mainly based on an easier understanding from users, and because I thought it would be simpler to implement with the current SpecCluster mechanism.

I'm not sure about this last part, and I don't quite grasp the consequences of this choice yet. I've not been eanble to thouroughly testing the current state of the PR, and it's not a strong opinion, so keeping the current implemented beahviour is fine with me too...

lesteve:
> OK, I can change that back if desired

That'd be my preference yes. Sorry if I changed my mind a few times in this issue. More than happy to hear other opinions of course!

I am going to try to explain my reasoning for future reference (and clarify for myself as well). Let me know if I forgot something:

### Pros of moving to `scale(n_jobs)`:
* `.scale(n_jobs)` is easier to understand for people familiar with a cluster 

### Cons
* this is a behaviour change (and a silent one), existing users have learned to live with this quirk and will need to correct their code
* many inconsistencies:
  - misleading widget: it will report things like `Workers: 3/1` (see snapshot below). Also setting the workers manually via the widget "Workers" input box will not produce the number of workers chosen.
  - misleading doc: `cluster.scale?` says `scale` takes an argument `n` which is the number of workers
   - using `FooCluster(n_workers=1, cores=6, processes=2)` will report 2 workers. 
   - other places in Dask that expect a number of workers e.g. `client.wait_for_workers` will be inconsistent. So you'll have to do something like:
     ```py
     cluster = FooCluster(cores=6, processes=2)
     cluster.scale(4) # 4 jobs
     client.wait_for_workers(8) # 8 worker processes
     ```
* `.scale(n_workers)` is what people familiar with dask will expect

Here is the snapshot with the confusing widget and docstring (`dask_jobqueue.local.LocalCluster` is nice for demonstration purposes):

image

And the associated code (note I am using distributed from master):
```py
from dask_jobqueue.local import LocalCluster
cluster = LocalCluster(cores=6, processes=3, memory='1GB')
cluster.scale?
cluster.scale(1)
cluster
```

jhamman:
Just chiming in quickly to say that my preference would be for `scale(n)` to represent the number of workers, not jobs. I think there is a great deal of value in supporting an abstraction beyond a *job*.

BTW, we could implement functionality to do both, e.g. `cluster.scale(jobs=4)`. This has the benefit of being more explicit and articulating the difference between dask-jobqueue and the other cluster managers.

mrocklin:
Sorry to be quiet here the last couple of days.  I got a little bit busy.
Would anyone else be interested in making the scale changes?

jhamman:
FWIW, I gave this a spin today. Things were mostly smooth. I still feel like the final piece here would be to sort out the scale changes. I think this basically means reimplementing the basic ideas in #63.

mrocklin:
Thanks for trying things out @jhamman !  Was there anything else behind the "mostly" adverb above?

> I still feel like the final piece here would be to sort out the scale changes. I think this basically means reimplementing the basic ideas in #63

My hope is that it's simpler than this.  I think that if we just divide by the number of workers per job in `def scale` then everything will work itself out.  I think that all of the adaptive work is already done.

lesteve:
> I think that if we just divide by the number of workers per job in def scale then everything will work itself out.

@mrocklin my understanding was that this would be a change in `distributed` but maybe you had something else in mind?

While working on https://github.com/dask/distributed/pull/3067, I realised the discrepancy of "grouped workers" vs "worker processes" already exists in SpecCluster when the spec generates multiple workers, e.g. `distributed.deploy.tests.test_spec_cluster.MultiWorker`.

```py
from distributed.deploy.tests.test_spec_cluster import MultiWorker
from distributed import Scheduler
from distributed import SpecCluster

cluster = SpecCluster(
    scheduler = {'cls': Scheduler},
    worker={
        "cls": MultiWorker,
        "options": {"n": 2, "nthreads": 6, "memory_limit": "4 GB"},
        "group": ["-0", "-1"],
})

cluster.scale(1)  # will generate 2 worker processes
# I would like .scale to be always in terms of "worker processes" 
# so I would like to use cluster.scale(2) instead  
cluster
```

mrocklin:
I would rather not make that policy decision in distributed just yet.  I'm still of the opinion that things will be a bit more complicated this way.  Instead, I would like to restrict this decision to dask-jobqueue for a while and see how it goes before enforcing it upstream.

lesteve:
> I would rather not make that policy decision in distributed just yet. I'm still of the opinion that things will be a bit more complicated this way

OK I trust your informed opinion on this. I would be interested if you have something in particular in mind that you think would be more complicated. Is it that the relation between `.scale` and `.worker_spec` would not be that straightforward.

> I would like to restrict this decision to dask-jobqueue for a while and see how it goes before enforcing it upstream.

So we would define `JobQueueCluster.scale` and the hope is that everything would just work, right?

FYI, it is very unlikely that I have time to look at this today and I'll be without connectivity from tonight until next Monday. I am going to state my opinion explictly in order to avoid it being a blocker:
- I am fine with this PR can be merged, if `.scale` works in terms of "worker processes".
- there is probably a few things to tidy up but that can be done after this PR is merged. I'd say getting them done before the next release would be ideal but not mandatory. IIRC they were mostly about quirks in the tests.

mrocklin:
> So we would define JobQueueCluster.scale and the hope is that everything would just work, right?

I've just pushed changes that do this.  I *think* that it should work out ok, but review and testing would be welcome.

> I'll be without connectivity from tonight until next Monday.

I hope this means that you will be enjoying life :)

> there is probably a few things to tidy up but that can be done after this PR is merged. I'd say getting them done before the next release would be ideal but not mandatory. IIRC they were mostly about quirks in the tests.

Do you mean these points from earlier?

> it would be great if we could find a way not to have to cd in a different folder (jobqueue scheduler-specific, mostly PBS and SLURM). In an ideal world the /dask-jobqueue volumes would be mounted on the worker images I think but also pbsuser would have the right to write into /dask-jobqueue. Not sure how easy it is to do the second part in docker.
xfail for .adapt(cores=...) is OK I think as long as we do remember to fix the issue in distributed before the next release

jhamman:
> Was there anything else behind the "mostly" adverb above?

Not really. I dusted off a fairly old use case. I noticed that using `startup_cost` in `adapt` ended up throwing a bunch of exceptions later on (details below). I wasn't sure if this ever worked correctly though so I didn't think much of it.


<details>
```python-traceback
distributed.core - ERROR - workers_to_close() got an unexpected keyword argument 'startup_cost'
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/core.py", line 399, in handle_comm
    result = handler(comm, **msg)
TypeError: workers_to_close() got an unexpected keyword argument 'startup_cost'
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <zmq.eventloop.ioloop.ZMQIOLoop object at 0x2b2e33748dd8>>, <Task finished coro=<AdaptiveCore.adapt() done, defined at /glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/deploy/adaptive_core.py:170> exception=TypeError("workers_to_close() got an unexpected keyword argument 'startup_cost'",)>)
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/deploy/adaptive_core.py", line 182, in adapt
    recommendations = await self.recommendations(target)
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/deploy/adaptive_core.py", line 152, in recommendations
    L = await self.workers_to_close(target=target)
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/deploy/adaptive.py", line 138, in workers_to_close
    **self._workers_to_close_kwargs
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/core.py", line 673, in send_recv_from_rpc
    result = await send_recv(comm=comm, op=key, **kwargs)
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/core.py", line 547, in send_recv
    raise exc.with_traceback(tb)
  File "/glade/u/home/jhamman/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/core.py", line 399, in handle_comm
    result = handler(comm, **msg)
```
</details>

I was glad to see that my existing configurations worked straight out of the box, and except for the change in jobs vs. core behavior (I see you've been working on that today), things behaved as the did before.

lesteve:
> I've just pushed changes that do this

Nice, thanks a lot for this!

> I hope this means that you will be enjoying life :)

That's the plan :).

> Do you mean these points from earlier?

Yes

dask_jobqueue/htcondor.py
```diff
@@ -220,3 +194,34 @@ def quote_environment(env):
...
+    >>> from dask_jobqueue.htcondor import HTCondorCluster
+    >>> cluster = HTCondorCluster(cores=24, memory="4GB", disk="4GB")
+    >>> cluster.scale(10)
+
+    >>> from dask.distributed import Client
+    >>> client = Client(cluster)
+
+    This also works with adaptive clusters.  This automatically launches and kill workers based on load.
+
+    >>> cluster.adapt(minimum=5)
```
lesteve: I think using `minimum` is very rarely useful. Probably the most useful example is to use `maximum_jobs`

dask_jobqueue/lsf.py
```diff
@@ -196,3 +150,49 @@ def lsf_detect_units():
...
+    lsf_units : str
+        Unit system for large units in resource usage set by the
+        LSF_UNIT_FOR_LIMITS in the lsf.conf file of a cluster.
+
+    Examples
+    --------
+    >>> from dask_jobqueue import LSFCluster
+    >>> cluster = LSFCluster(queue='general', project='DaskonLSF',
+    ...                      cores=15, memory='25GB')
+    >>> cluster.scale(10)  # this may take a few seconds to launch
```
lesteve: It would be great if all the docstring had the same example. Probably using `.scale(jobs=...)` and `.adapt(max_jobs=...)` is the thing that makes the most sense for `dask-jobqueue` users.

Can be done after this PR is merged.

dask_jobqueue/pbs.py
```diff
@@ -117,26 +102,39 @@ def __init__(
...
-        return "%dMB" % math.ceil(n / (1024 ** 2))
-    if n >= 10 * 1024:
-        return "%dkB" % math.ceil(n / 1024)
-    return "%dB" % n
+    Examples
+    --------
+    >>> from dask_jobqueue import PBSCluster
+    >>> cluster = PBSCluster(queue='regular', project="myproj", cores=24,
+    ...     memory="500 GB")
+    >>> cluster.scale(10)  # Ask for ten jobs
```
lesteve: I don't think the comment is true, right? This should use `cluster.scale(jobs=10)` for the comment to be true.

docs/source/howitworks.rst
```diff
@@ -24,10 +24,15 @@ object is instantiated:
    )
 
 You then ask for more workers using the ``scale`` command:
+You can either specify the number of jobs, or the total number of cores or
```
lesteve: While you are touching this file, I think it would improve the doc significantly if there was a precision about two things:
* `FooCluster` arguments are per-job characteristics, i.e. 24 cores per job, `'100GB'` per job
* creating a `FooCluster` object does not launch any jobs, you need to use `.scale` for launching jobs

That can be done in a separate PR though.

docs/source/index.rst
```diff
@@ -19,7 +19,7 @@ Example
 
    from dask_jobqueue import PBSCluster
    cluster = PBSCluster()
-   cluster.scale(10)         # Ask for ten workers
+   cluster.scale(10)         # Deploy ten single-node jobs
```
lesteve: You need `cluster.scale(jobs=10)` for the comment to be true right?

docs/source/index.rst
```diff
@@ -45,7 +45,8 @@ save resources when not actively computing.
 
 .. code-block:: python
 
-   cluster.adapt(minimum=6, maximum=90)  # auto-scale between 6 and 90 workers
+   cluster.adapt(minimum=6, maximum=90)  # auto-scale between 6 and 90 jobs
```
lesteve: Should be `minimum_jobs=6, maximum_jobs=90` right?

dask_jobqueue/pbs.py
mrocklin: Fixed

dask_jobqueue/lsf.py
mrocklin: Fixed.

dask_jobqueue/htcondor.py
mrocklin: Fixed

mrocklin:
OK, I think that I've addressed all comments.  @lesteve I also removed the xfail on the adaptive test (things work upstream now).  Tests pass.  From my perspective this is good to go.

lesteve:
I think this looks good, in particular the code to convert from `n` to `jobs` in both `scale` and `adapt` seems fine (only read the code though not tried it ).

I am going to merge this PR. I feel this has already been under review long enough and  the only contentious point left was the `scale` in terms of "worker process" vs "grouped workers" which has been resolved in recent commits in favour of "worker processes".

It would be great if people involved in this PR could test `master` on their favourite cluster, in particular whether the `scale(jobs=...)` or `adapt(max_jobs=...)` is working as they expect, and report back if they see any problems!

Thanks a lot to everyone involved in this PR and in particular @mrocklin for doing the bulk of the work!

Another thing to tidy up in another PR is to stop using `distributed` from master in CI, my expectation is that 0.24 should be good enough.

jhamman:
Thanks all! Glad to see this brought to completion.

mrocklin:
Woot!

jacobtomlinson:
🚀

mrocklin:
Next question: what should happen before we release?

It would be nice if people could try out `master` on their own systems and report back.

I *think* that I've covered all of the concerns that @lesteve had (except for the `cd` issue in the docker testing infrastructure).  Are there any other known issues?

lesteve:
I had this mostly in mind before the release:

> It would be great if people involved in this PR could test master on their favourite cluster, in particular whether the scale(jobs=...) or adapt(max_jobs=...) is working as they expect, and report back if they see any problems!

It would have been great since we are introducing breaking changes to potentially think about introducing a few more, e.g. #323 and #205 but I don't think that is critical.

guillaumeeb:
🎆 👍  🥇 🚀 

 Thanks @mrocklin and @lesteve for all the work! I'll try to test this in the next few days.

PR #312, opened 2019-08-14
use f-strings and recommended syntax for Python 3.6

basnijholt:


guillaumeeb:
Thanks @basnijholt. Do we want to drop support for Python 3.5? I'm not sure...

lesteve:
I would be in favour of aligning on Python versions support with dask. Dask supports Python 3.5, see https://docs.dask.org/en/latest/develop.html#python-versions.

I would be more than happy to revisit this PR when dask drops Python 3.5.

basnijholt:
OK, then I'll close this PR.

I just (wrongly) assumed that Python 3.5 had already been dropped because of:

https://github.com/dask/dask-jobqueue/blob/7970f31df6fc4cb41f4adf65d5e088b52cb0b8f2/setup.py#L24

Actually, this means that only 3.5.0 is not supported but 3.5.1 and up is. Maybe it should change into `>=3.5`?

PR #317, opened 2019-08-19
support Python 3.5.0 too

basnijholt:
see https://github.com/dask/dask-jobqueue/pull/312#issuecomment-522478412


lesteve:
Thanks a lot for the PR! Dask does this as well (see https://github.com/dask/dask/blob/master/setup.py#L51) so let's merge this one.

PR #319, opened 2019-08-20
support older HTCondor versions without -file argument and add schedd argument

riedel:
The  -file argument was introduced in recent versions of HTCondor otherwise the script argument is positional. This adds support for older versions commonly in the wild. (Would only generate a problem if more than one job is queued at once)

### DIFF
--- a/dask_jobqueue/htcondor.py
+++ b/dask_jobqueue/htcondor.py
@@ -47,9 +47,11 @@ class HTCondorCluster(JobQueueCluster):
 Environment = "%(quoted_environment)s"
 Arguments = "%(quoted_arguments)s"
 Executable = %(executable)s
+
+Queue
 """.lstrip()
 
-    submit_command = "condor_submit -queue 1 -file"
+    submit_command = "condor_submit"
     cancel_command = "condor_rm"
     job_id_regexp = r"(?P<job_id>\d+\.\d+)"
 
--- a/dask_jobqueue/htcondor.py
+++ b/dask_jobqueue/htcondor.py
@@ -18,6 +18,8 @@ class HTCondorCluster(JobQueueCluster):
     ----------
     disk : str
         Total amount of disk per job
+    schedd : str
+       Submit to the specified condor_schedd
     job_extra : dict
         Extra submit file attributes for the job
     %(JobQueueCluster.parameters)s
@@ -59,7 +61,9 @@ Queue
     # Python (can't find its libs), so we have to go through the shell.
     executable = "/bin/sh"
 
-    def __init__(self, disk=None, job_extra=None, config_name="htcondor", **kwargs):
+    def __init__(
+        self, disk=None, job_extra=None, schedd=None, config_name="htcondor", **kwargs
+    ):
         if disk is None:
             disk = dask.config.get("jobqueue.%s.disk" % config_name)
         if disk is None:
@@ -71,6 +75,8 @@ Queue
             self.job_extra = dask.config.get("jobqueue.%s.job-extra" % config_name, {})
         else:
             self.job_extra = job_extra
+        if schedd is None:
+            schedd = dask.config.get("jobqueue.%s.schedd" % config_name, None)
 
         # Instantiate args and parameters from parent abstract class
         super().__init__(config_name=config_name, **kwargs)
@@ -109,6 +115,9 @@ Queue
             )
         if self.job_extra:
             self.job_header_dict.update(self.job_extra)
+        if schedd:
+            self.submit_command += " -name " + shlex.quote(schedd)
+            self.cancel_command += " -name " + shlex.quote(schedd)
 
     def env_lines_to_dict(self, env_lines):
         """ Convert an array of export statements (what we get from env-extra
--- a/dask_jobqueue/jobqueue.yaml
+++ b/dask_jobqueue/jobqueue.yaml
@@ -155,6 +155,7 @@ jobqueue:
 
     # HTCondor Resource Manager options
     disk: null                  # Total amount of disk per job
+    schedd: null                # Submit to the specified condor_schedd
     extra: []
     env-extra: []
     job-extra: {}               # Extra submit attributes


lesteve:
Thanks for your PR. Full disclosure: I don't know anything about Condor (and I think this is the case of most maintainers in `dask-jobqueue`) and we have very little tests right now for HTCondor. This makes it difficult to merge your PR without confirmation from somebody else that it is a reasonable thing to do.

I am going to ping the people involved in the original HTCondor PR https://github.com/dask/dask-jobqueue/pull/245 and wait for their informed feed-back: @matyasselmeci @cloudronin @mivade.

It would be great if some of you could test it quickly on your HTCondor cluster.
A simple snippet to test it on your HTCondor cluster would look like this:

```py
import time

from dask_jobqueue import HTCondorCluster

cluster = HTCondorCluster(your_arguments_go_here)
client = Client(cluster)

while len(client.scheduler_info()['workers']) < 1:
    print('No workers connected yet')
    time.sleep(5)

future = client.submit(lambda x: x, 1)
print('Got result: ', future.result())
```

lesteve:
@riedel I see you have pushed an additional commit recently so I am guessing you are still working on this PR. I would suggest you let us know when this stabilizes and you think this PR is ready for feed-back.

mivade:
I can test this out once it's ready for review.

dask_jobqueue/htcondor.py
```diff
@@ -107,6 +113,9 @@ def __init__(self, disk=None, job_extra=None, config_name="htcondor", **kwargs):
             )
         if self.job_extra:
             self.job_header_dict.update(self.job_extra)
+        if schedd:
+            self.submit_command += " -name "+schedd
```
matyasselmeci:
```suggestion
            self.submit_command += " -name "+shell_quote(schedd)
```

dask_jobqueue/htcondor.py
```diff
@@ -107,6 +113,9 @@ def __init__(self, disk=None, job_extra=None, config_name="htcondor", **kwargs):
...
+            self.cancel_command += " -name "+schedd
```
matyasselmeci:
```suggestion
            self.cancel_command += " -name "+shell_quote(schedd)
```

lesteve:
> ```
> try:
>     from shlex import quote as shell_quote  # Python 3.3 and newer
> except ImportError:
>     from pipes import quote as shell_quote  # Python 2 to 3.2
> ```

Thanks for the review @matyasselmeci, note that we have dropped Python 2 support, since `dask-jobqueue` 0.6 (released 6 July 2019) so this should be sufficient:

```py
from shlex import quote
```

dask_jobqueue/htcondor.py
riedel: Thanks! I used simply shlex.quote since shlex was already imported as a whole module (see https://github.com/dask/dask-jobqueue/pull/319#issuecomment-523049358) .

lesteve:
@riedel can you look at the build failures? Quickly looking at it:
* one file needs to be formatted with `black`. You can install `pre-commit` (there is a configuration step as well IIRC) so that the files get automatically `black`ed before a commit
* I think you need to add your `schedd` key to `dask_jobqueue/jobqueue.yaml`

mivade:
One issue I'm seeing is that I get a `KeyError` if I don't specify `schedd`:

```pytb
  File "condor.py", line 7, in <module>
    cluster = HTCondorCluster(disk="1 GB", cores=8, memory="1 GB")
  File "/Users/michael.depalatis/tmp/dask-condor/dask-jobqueue/dask_jobqueue/htcondor.py", line 77, in __init__
    schedd = dask.config.get("jobqueue.%s.schedd" % config_name)
  File "/Users/michael.depalatis/miniconda3/envs/dask-condor/lib/python3.7/site-packages/dask/config.py", line 456, in get
    result = result[k]
KeyError: 'schedd'
```

It would be nice to have that as optional for cases where you're submitting to the `schedd` on the machine you're running on.

riedel:
> It would be nice to have that as optional for cases where you're submitting to the `schedd` on the machine you're running on.

Definitively, I set the default to null to get that effect. I forgot to commit the file.

lesteve:
Thanks a lot for this! One question (I could not figure it out in 5 minutes of googling): is it possible to put the `schedd` in the job script? I am guessing probably not since it is needed in the `condor_rm` command but asking just in case because if it was `schedd` would be more similar to all the other `HTCondor` parameters.

It would be really great to add a test in `tests. The simplest thing to do would be to something like:
```py
cluster = HTCondor(cores=1, memory='1GB', schedd='my-schedd-name')
# pretty sure I the quoting is wrong but you see what I mean
assert `-name my-schedd-name' in cluster.submit_command
assert `-name my-schedd-name' in cluster.submit_command
```

> (I could not run them on my development system, so I forgot)

For tests that don't need a cluster to run, you can run them locally (i.e. no need to be on the cluster).
```
git clone https://github.com/dask/dask-jobqueue
cd dask-jobqueue
pytest dask_jobqueue
```

lesteve:
Also I opened #324 to add a CONTRIBUTING.md with "how to run tests".

riedel:
> Thanks a lot for this! One question (I could not figure it out in 5 minutes of googling): is it possible to put the `schedd` in the job script?

My first intuituition would be "no" but there might actually be sth, I have never tested (I am not an condor expert at all):
> For a grid-type-string of condor, the first parameter is the name of the remote condor_schedd daemon. The second parameter is the name of the pool to which the remote condor_schedd daemon belongs.

(What I thought could be done is to change the shebang and call just the script.)

But as you said, the problem then is that you also need to pass it to `condor_rm`  and I like symmetry. It is actually a good reference for all the parameters needed to be passed outside of job_extra.


BTW: the `-addr` and `-pool`parameter is also sth that one might also want to specify. 

Should I add those as well to simply be complete?  (I currently have no use for them, but actually it might be interesting)

I can add a test.

mivade:
> the problem then is that you also need to pass it to condor_rm and I like symmetry.

From reading the `condor_submit` docs with the `grid_resource` option, I think you *can* specify a remote `schedd` without a command-line option, but if my understanding of it is correct, you still need to be submitting to a local `schedd` which then will coordinate with the remote one.

> BTW: the `-addr` and `-pool`parameter is also sth that one might also want to specify.
> 
> Should I add those as well to simply be complete? (I currently have no use for them, but actually it might be interesting)

I would at least add the `-addr`. `-pool` could also be useful but not something that I can easily test.

lesteve:
> Should I add those as well to simply be complete? (I currently have no use for them, but actually it might be interesting)

I would only add a feature if someone has a use for it now rather than because someone may need it one day. In other words: YAGNI (You Aren't Going to Need It).

lesteve:
Thanks for the details! In case it was not clear enough already, I completely defer to people involved in this PR for what kind of parameter makes sense to add for `HTCondor`.

Something I would say I think the minimal requirement for adding a parameter is that:
1. someone can test it on a HTCondor cluster
2. someone actually needs it

Another thing we could do is to have a `condor_command_extra` (better name needed):
```py
cluster = HTCondor(condor_command_extra=['-name my_schedd'])
```

This way we don't have to decide which arguments we need to include in this PR and this is more future-proof.

This is if all the parameters we are talking about symmetric between `condor_submit` and `condor_rm`? If not you could have `submit_command_extra` or `cancel_command_extra` (or better names).

matyasselmeci:
Can we split out 97ff9ab (the -file fix) and merge that for now? I didn't see any controversy over that fix and _is_ actually biting people in the wild.

matyasselmeci:
As for specifying the schedd, I think @lesteve's suggestion of `submit_command_extra` and `cancel_command_extra` is a good way of solving this problem; that way we don't have to argue over which arguments are "important".

lesteve:
> Can we split out 97ff9ab (the -file fix) and merge that for now?

I opened #351 to do this, let me know if that was what you had in mind.

riedel:
I opened a ticket referencing this pull request. Would be too messy to update it (was messed up from the beginning because I accidentally merged two pull requests, one of which was resolved!)

PR #320, opened 2019-08-20
specify condor_schedd

riedel:
While I am at it ( #319 ) and you are looking for testers ( @matyasselmeci @cloudronin @mivade): I also had to add an argument to define the scheduler location in our setup.

Would be great if this also could be merged.

### DIFF
--- a/dask_jobqueue/htcondor.py
+++ b/dask_jobqueue/htcondor.py
@@ -18,6 +18,8 @@ class HTCondorCluster(JobQueueCluster):
     ----------
     disk : str
         Total amount of disk per job
+    schedd : str
+       Submit to the specified condor_schedd
     job_extra : dict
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @jrbourbeau , who I think did the Condor work

lesteve:
@riedel It seems like there this PR has some overlap with #319 which does add `schedd` as well, I am guessing this was not intended?

If you meant two create two separate independent PRs, can you clean up #319 and tackle https://github.com/dask/dask-jobqueue/pull/319#pullrequestreview-277203251 in this PR?

riedel:
My bad, I wanted to make 2 seperate PR. but I did not create a feature branch on the other one and accidentially committed the other one to master, too. 

Since there is already comments on the other one I close this one...

Issue #332, opened 2019-08-29
Should we default to multiple processes (rather than multiple threads)?


lesteve:
```py
from dask_jobqueue import SLURMCluster
cluster = SLURMCluster(cores=4, memory='16GB')
[line for line in cluster.job_script().splitlines() if 'dask_worker' in line]
```

Excerpt of the output:
```
python -m distributed.cli.dask_worker --nthreads 4 ...
```

So by default we use one dask worker with 4 threads. This does not seem ideal.
In a Python world (because of the GIL), using multi-processing by default for parallelism is more likely to increase performance.

Maybe we can use the same heuristic of processes vs threads as `LocalCluster` (`distributed.deploy.local.nprocesses_nthreads`):
https://github.com/dask/distributed/blob/ad0c7c23c384e981e0622c36e7fb132429becc34/distributed/deploy/local.py#L211
summary of the heuristic:
* `if cores <= 4, processes=cores, threads_per_process=1`
* `if cores > 4, processes ~= sqrt(cores), threads_per_process ~= sqrt(cores)` and `processes > threads_per_process`

mrocklin:
That seems like a decent heuristic if `processes=None`

PR #334, opened 2019-09-08
Default to multiprocesses instead of single process

jennirinker:
Resolving #332 .

This configuration change in JobQueueCluster changes the default processes per job so that they follow the same heuristic as dask (proc=cores for cores <= 4, otherwise it's the smallest factor greater than the square root of cores).

PR also includes modified documentation (though it doesn't explain the heuristic -- perhaps that should be done on dask's docs and we link it?).

### DIFF
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -25,7 +25,7 @@ def test_errors():
 
 
 def test_command_template():
-    with PBSCluster(cores=2, memory="4GB") as cluster:
+    with PBSCluster(cores=2, processes=1, memory="4GB") as cluster:
         assert (
             "%s -m distributed.cli.dask_worker" % (sys.executable)
             in cluster._command_template
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -189,3 +189,10 @@ def test_cluster_has_cores_and_memory(Cluster):
 
     with pytest.raises(ValueError, match=cls_name + r"cores=4, memory='\d+GB'"):
         Cluster(cores=4)
+
+
+def test_default_workers():
+    """For 6 cores, default is 1 workers with 3 procs, 2 threads/proc"""
+    with PBSCluster(cores=6, memory="4GB") as cluster:
+        assert " --nprocs 3" in cluster._command_template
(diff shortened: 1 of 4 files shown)


mrocklin:
Note that there is a pending rewrite of this project in
https://github.com/dask/dask-jobqueue/pull/307 .  It might be a good idea
to wait for that;

jennirinker:
Ack, I already wrote and pushed the new test. Well, that's ok if this never gets merged, it was a good test-run for me on contributing.

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -189,3 +189,10 @@ def test_cluster_has_cores_and_memory(Cluster):
 
     with pytest.raises(ValueError, match=cls_name + r"cores=4, memory='\d+GB'"):
         Cluster(cores=4)
+
+
+def test_default_workers():
```
willirath: We could modify this further to cover all clusters:

```python
@pytest.mark.parametrize(
    "Cluster",
    [PBSCluster, MoabCluster, SLURMCluster, SGECluster, LSFCluster, OARCluster],
)
def test_default_workers(Cluster):
    """For 6 cores, default is 1 workers with 3 procs, 2 threads/proc"""
    with Cluster(cores=6, memory="4GB") as cluster:
        assert " --nprocs 3" in cluster._command_template
        assert " --nthreads 2" in cluster._command_template
```

But let's only do this if there's agreement that this will be merged before the big rewrite.

lesteve:
> Well, that's ok if this never gets merged, it was a good test-run for me on contributing.

Thanks a lot for the PR @jennirinker! I am confident that this will get merged since this is something we want to do as #332 indicates.

I agree with @mrocklin that this is a good idea to first merge #307. Once #307 is merged this PR may need to be tweaked a bit but I expect that it should not be too much work.

lesteve:
@jennirinker side-comment: Please use "Fix #issueNumber" in your PR description, this way the associated issue gets closed automatically when the PR is merged. For more details, look at [this](https://github.com/blog/1506-closing-issues-via-pull-requests).

Using anything more fancy e.g. "resolving issue #issueNumber" does not work unfortunately. I updated your PR description.

jennirinker:
@lesteve Ah, thanks. I was close -- "resolve" instead of "resolving" and I would've had it. Next time.

lesteve:
No worries, this is the kind of caveats that trip everyone up once in a while ..

Issue #344, opened 2019-09-30
Try out dask-jobqueue development version on different clusters


lesteve:
There has been a non-trivial refactoring in #307 to use `SpecCluster` from `distributed`. It would be great if people could test this on their cluster to avoid unexpected problems when we released the next version of `dask-jobqueue` (probably a matter of one week or two):
* [x] `PBSCluster`: @guillaumeeb @andersy005. This has already been tested while #307 was being worked on I think.
* [x] `HTCondor`: @matyasselmeci @cloudronin @mivade
* [x] `LSFCluster`: tested by @mrocklin (on Summit I think). Not sure who else would be a good person.
* [x] `SLURMCluster`: tested by @mrocklin. Not sure who else would be a good person.
* [ ] `SGECluster`: I can test this. Maybe @ericmjl can try it out as well.
* [ ] `OARCluster`: I can test this.

To install the `dask-jobqueue` development version:
```
pip install `git+https://github.com/dask/dask-jobqueue`
```
You want to make sure that you have `distributed >= 2.5.1` (see #342 for more details).

There is a list of known breaking changes:
https://github.com/dask/dask-jobqueue/blob/master/docs/source/changelog.rst#changelog

If you spot some more breaking changes, let us know, so we can improve the changelog!

mivade:
Looks good on HTCondor. I mainly checked that I can scale up and down the number of workers and submit to them.

lesteve:
@mivade glad to hear that, thanks a lot for trying it out!

If you could also check simple adaptative use cases, that would be great, e.g. something like this:
```py
cluster.adapt(minimum_jobs=0, maximum_jobs=2)
client.map(func, list_of_args) # func should take a few seconds
# make sure that the cluster scales up with 2 jobs at most and then down to 0 jobs
```

mivade:
Adaptive mostly works. The only issue I see is that I end up with a single worker after all jobs are complete even with `minimum_jobs=0`.

mrocklin:
Do you perhaps still have a future lying around?  If so Dask needs to keep
a worker around to hold the results

>>> client.futures

mivade:
Rookie mistake. That was exactly it, @mrocklin.

@lesteve let me know if there are any other tests I can do.

mrocklin:
We must all be rookies then :)

andersy005:
> SLURMCluster: tested by @mrocklin. Not sure who else would be a good person.

[...]

PR #346, opened 2019-10-01
Update `.scale()` and `.adapt()` docstrings

andersy005:
Addresses https://github.com/dask/dask-jobqueue/issues/344#issuecomment-537092345

I am not sure what's the right way to include the information from `distributed's SpecCluster` without copying and pasting it into the new docstrings:

- `adapt()`: Updated
```python
In [4]: cluster.adapt?                                                                                          
Signature:
cluster.adapt(
    *args,
    minimum_jobs: int = None,
    maximum_jobs: int = None,
    **kwargs,
)
Docstring:
Scale Dask cluster automatically based on scheduler activity.

Parameters
----------

minimum_jobs : int
   Minimum number of jobs
maximum_jobs : int
   Maximum number of jobs
**kwargs :
   Extra parameters to pass to dask.distributed.Adaptive
```

- `adapt()`: Old version

```python
In [4]: cluster.adapt?                                                                                          
Signature:
cluster.adapt(
    minimum_cores=None,
    maximum_cores=None,
    minimum_memory=None,
    maximum_memory=None,
    **kwargs,
)
Docstring:
Turn on adaptivity
For keyword arguments see dask.distributed.Adaptive
Instead of minimum and maximum parameters which apply to the number of
worker, If Cluster object implements jobqueue_worker_spec attribute, one can
use the following parameters:
Parameters
----------
minimum_cores: int
    Minimum number of cores for the cluster
maximum_cores: int
    Maximum number of cores for the cluster
minimum_memory: str
    Minimum amount of memory for the cluster
maximum_memory: str
    Maximum amount of memory for the cluster
Examples
--------
>>> cluster.adapt(minimum=0, maximum=10, interval='500ms')
>>> cluster.adapt(minimum_cores=24, maximum_cores=96)
>>> cluster.adapt(minimum_memory='60 GB', maximum_memory= '1 TB')
```

- scale
```python
                                                                      
In [3]: cluster.scale?                                                                                          
Signature: cluster.scale(n=None, jobs=0, memory=None, cores=None)
Docstring:
Scale cluster to specified configurations.

Parameters
----------
n : int
   Target number of workers
jobs : int
   Target number of jobs
memory : str
   Target amount of memory
cores : int
   Target number of cores
```

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -516,10 +516,10 @@ class JobQueueCluster(SpecCluster):
 
         Parameters
         ----------
-        minimum_cores : int
-           Minimum number of cores for the cluster
-        maximum_cores : int
-           Maximum number of cores for the cluster
+        minimum : int
+           Minimum number of workers to keep around for the cluster
+        maximum : int
+           Maximum number of workers to keep around for the cluster
         minimum_memory : str
            Minimum amount of memory for the cluster
         maximum_memory : str
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -516,7 +516,14 @@ class JobQueueCluster(SpecCluster):
 
         Parameters
         ----------
-
+        minimum_cores : int
+           Minimum number of cores for the cluster
+        maximum_cores : int
+           Maximum number of cores for the cluster
+        minimum_memory : str
+           Minimum amount of memory for the cluster
+        maximum_memory : str
+           Maximum amount of memory for the cluster
         minimum_jobs : int
            Minimum number of jobs
         maximum_jobs : int
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -490,6 +490,20 @@ class JobQueueCluster(SpecCluster):
         return self._dummy_job.job_name
 
     def scale(self, n=None, jobs=0, memory=None, cores=None):
+        """ Scale cluster to specified configurations.
+
+        Parameters
+        ----------
+        n : int
+           Target number of workers
+        jobs : int
+           Target number of jobs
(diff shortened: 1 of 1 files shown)


dask_jobqueue/core.py
```diff
@@ -498,6 +512,23 @@ def scale(self, n=None, jobs=0, memory=None, cores=None):
     def adapt(
         self, *args, minimum_jobs: int = None, maximum_jobs: int = None, **kwargs
     ):
+        """ Scale Dask cluster automatically based on scheduler activity.
+
+        Parameters
+        ----------
+
```
jhamman: I think we should also include the minimum/maximum arguments here. I suspect most users will interact with these regularly and I think chasing the arguments through to dask.distributed.Adaptive will end up being annoying.

andersy005:
@jhamman, 

When you get time, this is ready for another look.

dask_jobqueue/core.py
```diff
@@ -498,6 +512,30 @@ def scale(self, n=None, jobs=0, memory=None, cores=None):
...
         self, *args, minimum_jobs: int = None, maximum_jobs: int = None, **kwargs
     ):
+        """ Scale Dask cluster automatically based on scheduler activity.
+
+        Parameters
+        ----------
+        minimum_cores : int
+           Minimum number of cores for the cluster
+        maximum_cores : int
+           Maximum number of cores for the cluster
```
jhamman: I don't think we want the `_cores` in these names. These will be passed through to https://distributed.dask.org/en/latest/api.html#adaptive which just uses `minimum`/`maximum`.

andersy005: 👍 Fixed it ( I think :))

PR #347, opened 2019-10-04
Support stdin with LSF bsub

mrocklin:


lesteve:
Looks good to me, @SimonBoothroyd, can you confirm that this PR works for you?

@mrocklin I assume you have tested this on Summit for LSF 10?

SimonBoothroyd:
@lesteve I tested the config route and it works great for me! But the `lsf_version` route raises a `TypeError: cannot use a string pattern on a bytes-like object`. I think also the regex pattern will match `10.1.0.7` as `0.1.0.7` but something more like
```
@toolz.memoize
def lsf_version():
    out, _ = subprocess.Popen("lsid", stdout=subprocess.PIPE).communicate()
    version = re.search(r"(\d+\.)+\d+", out.decode()).group()
    return LooseVersion(version)
```

should work. From a quick test of this it also seems like `use_stdin` should just compare `< "10"` rather than `<= "9"` to handle cases like `9.1.XXX`, but otherwise this is great - thanks for making this change @mrocklin !

mrocklin:
> @mrocklin I assume you have tested this on Summit for LSF 10?

I didn't before (my apologies) but I have now, and yes, things work with the recent changes suggested by @SimonBoothroyd

Issue #350, opened 2019-10-11
HTCondor cluster cannot scale up


simone-codeluppi:
Hi
I finally had the chance to test dask-jobqueue on our HTCondor cluster. Thanks @matyasselmeci for creating the functionality!

As reported in the API guide [HTCondor_API](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.HTCondorCluster.html#) 
i run:

```python
from dask_jobqueue.htcondor import HTCondorCluster
cluster = HTCondorCluster(cores=24, memory="4GB", disk="4GB")
```

When I scale up to  10 jobs 

```python
cluster.scale(jobs=10)  # ask for 10 jobs
```

 I get the following error :  

```python
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOLoop object at 0x7fcaf8bb2780>>, <Task finished coro=<
SpecCluster._correct_state_internal() done, defined at /fish/simone_processing/anaconda3/envs/dask_htc/lib/python3.7/site-packages/distributed/deploy/spec.py:284> exception=RuntimeError('Command exited wi
th non-zero exit code.\nExit code: 1\nCommand:\ncondor_submit -queue 1 -file /tmp/tmpvytzt0ez.sh\nstdout:\n\nstderr:\nUsage: condor_submit [options] [<attrib>=<value>] [- | <submit-file>]\n    [options] a
re\n\t-terse  \t\tDisplay terse output, jobid ranges only\n\t-verbose\t\tDisplay verbose output, jobid and full job ClassAd\n\t-debug  \t\tDisplay debugging output\n\t-append <line>\t\tadd line to submit 
file before processing\n\t              \t\t(overrides submit file; multiple -a lines ok)\n\t-queue <queue-opts>\tappend Queue statement to submit file before processing\n\t                   \t(submit fi
le must not already have a Queue statement)\n\t-batch-name <name>\tappend a line to submit file that sets the batch name\n\t-disable\t\tdisable file permission checks\n\t-dry-run <filename>\tprocess submi
t file and write ClassAd attributes to <filename>\n\t        \t\tbut do not actually submit the job(s) to the SCHEDD\n\t-maxjobs <maxjobs>\tDo not submit if number of jobs would exceed <maxjobs>.\n\t-sing
[...]

PR #351, opened 2019-10-14
Support older HTCondor versions without -file argument

lesteve:
Split off from #319.

@matyasselmeci is that what you had in mind?


matyasselmeci:
Yep, exactly.

lesteve:
OK I am going to merge this then, thanks for your feed-back!

@simone-codeluppi since you had the issue, it would be great if you could confirm that using `dask-jobqueue` master (e.g. through `pip install git+https://github.com/dask/dask-jobqueue`) fixes the issue you saw in #350.

simone-codeluppi:
@lesteve the `dask-jobqueue` master fixed my issue. Thanks for fix it!

lesteve:
Great thanks for the confirmation!

PR #357, opened 2019-10-21
Create CONTRIBUTING.md

jacobtomlinson:
Ref dask/community#17


lesteve:
That looks good to me, thanks a lot for this.

My current plan leave is to leave it open for a day or so, in case the other Dask-related projects have some generic feed-back that you may want to apply across all the projects.

Issue #328, opened 2019-08-25
LSFCluster may be overly specific?


mrocklin:
I was trying out dask-jobqueue on the Summit supercomputer at Oak Ridge National Labs.  I ran into a number of problems with our current configuration that seem to be special cases.  I propose un-special-casing these, but I would like to get feedback from others who use LSF.

### bsub command

We currently do some odd things with bsub.  This made things fail for me on a login node (although they did work on a compute node).  Removing this special-cased behavior made things work well for me in both places.

```diff
diff --git a/dask_jobqueue/lsf.py b/dask_jobqueue/lsf.py
index 95042a8..12126ee 100644
--- a/dask_jobqueue/lsf.py
+++ b/dask_jobqueue/lsf.py
@@ -53,7 +53,7 @@ class LSFJob(Job):
     """,
         4,
     )
-    submit_command = "bsub <"
+    submit_command = "bsub"
     cancel_command = "bkill"

     def __init__(
@@ -134,10 +134,6 @@ class LSFJob(Job):

         logger.debug("Job script: \n %s" % self.job_script())

-    def _submit_job(self, script_filename):
-        piped_cmd = [self.submit_command + " " + script_filename + " 2> /dev/null"]
-        return self._call(piped_cmd, shell=True)
```

cc'ing @raybellwaves and @guillaumeeb , who show up under `git blame` for this code

### `-R "span[hosts=1]"`

My particular deployment didn't like these lines.  I don't know if these are very useful generally though, and I can work around them.

```python
            if ncpus > 1:
                # span[hosts=1] _might_ affect queue waiting
                # time, and is not required if ncpus==1
                header_lines.append('#BSUB -R "span[hosts=1]"')
```

cc @louisabraham (also pointed to by `git blame`)

raybellwaves:
Not at UM anymore. Pinging a colleague who may use this or know someone who can give feedback @milancurcic

louisabraham:
Hi,

```diff
-    submit_command = "bsub <"
+    submit_command = "bsub"
```

I know `bsub` can read the command from the args (not sure about stdin). Is this what this is about?


```
-R "span[hosts=1]"
```

[...]

PR #360, opened 2019-10-24
LSF: Accept use_stdin in the constructor

stuarteberg:
Right now, all LSF options can be specified either in the config OR in the constructor arguments, except the new `use-stdin` option (introduced in #347). Unlike all the others, that option may only be specified in the config (not the constructor).

I don't see why we'd want `use-stdin` to be different from all the other LSF options, so this PR allows the user to pass `use_stdin` to the `LSFCluster` constructor if she wants to.  (As usual, the config is used if no value was passed in the constructor.)

```python
cluster = LSFCluster(cores=15, memory='25GB',
                     use_stdin=True) # <-- now allowed
```

Side note: I suspect this new setting will be needed by many, if not most, LSF users, so I added some verbose documentation for it.

FWIW, I tested these changes on my LSF cluster, and they work as expected.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,8 @@ from distributed.utils_test import loop  # noqa: F401
 
 import pytest
 
+import dask_jobqueue.lsf
+
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -27,3 +29,18 @@ def pytest_runtest_setup(item):
     if envnames:
         if item.config.getoption("-E") not in envnames:
             pytest.skip("test requires env in %r" % envnames)
+
+
+@pytest.fixture(autouse=True)
+def mock_lsf_version(monkeypatch, request):
+    # Monkey-patch lsf_version() UNLESS the 'lsf' environment is selected.
+    # In that case, the real lsf_version() function should work.
+    markers = list(request.node.iter_markers())
+    if any("lsf" in marker.args for marker in markers):
+        return
+
+    try:
+        dask_jobqueue.lsf.lsf_version()
+    except OSError:
+        # Provide a fake implementation of lsf_version()
+        monkeypatch.setattr(dask_jobqueue.lsf, "lsf_version", lambda: "10")
--- a/dask_jobqueue/jobqueue.yaml
+++ b/dask_jobqueue/jobqueue.yaml
@@ -140,7 +140,7 @@ jobqueue:
     job-extra: []
     log-directory: null
     lsf-units: null
-    use-stdin: null
+    use-stdin: null             # (bool) How jobs are launched, i.e. 'bsub jobscript.sh' or 'bsub < jobscript.sh'
 
   htcondor:
     name: dask-worker
--- a/dask_jobqueue/lsf.py
+++ b/dask_jobqueue/lsf.py
@@ -29,6 +29,7 @@ class LSFJob(Job):
         job_extra=None,
         lsf_units=None,
         config_name="lsf",
+        use_stdin=None,
         **kwargs
     ):
         if queue is None:
@@ -45,6 +46,12 @@ class LSFJob(Job):
             job_extra = dask.config.get("jobqueue.%s.job-extra" % config_name)
         if lsf_units is None:
             lsf_units = dask.config.get("jobqueue.%s.lsf-units" % config_name)
+
+        if use_stdin is None:
+            use_stdin = dask.config.get("jobqueue.%s.use-stdin" % config_name)
+        if use_stdin is None:
+            use_stdin = lsf_version() < "10"
+        self.use_stdin = use_stdin
 
         # Instantiate args and parameters from parent abstract class
         super().__init__(*args, config_name=config_name, **kwargs)
@@ -97,7 +104,7 @@ class LSFJob(Job):
         logger.debug("Job script: \n %s" % self.job_script())
 
     async def _submit_job(self, script_filename):
-        if use_stdin():
+        if self.use_stdin:
             piped_cmd = [self.submit_command + "< " + script_filename + " 2> /dev/null"]
             return self._call(piped_cmd, shell=True)
         else:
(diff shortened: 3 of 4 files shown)


stuarteberg:
OK, I got the tests passing, but I'm seeing intermittent failures from Travis that are unrelated to this PR.

FWIW, here's the error:

```
PREFIX=/opt/anaconda

Unpacking payload ...

  0%|          | 0/35 [00:00<?, ?it/s]

No output has been received in the last 10m0s, this potentially
indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on:
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

At first this error appeared in the `JOBQUEUE=pbs` build, so I triggered a new build. Now `pbs` passes, but I see the same error in the `JOBQUEUE=slurm` build.

I'll force-push one more time to see if I can get lucky with a successful build.

stuarteberg:
>I'll force-push one more time to see if I can get lucky with a successful build.

Darn -- no luck.  I think this PR is ready for merge anyway.

lesteve:
> Side note: I suspect this new setting will be needed by many, if not most, LSF users, so I added some verbose documentation for it.

@stuarteberg do you think we should default to `use_stdin=True`? As mentioned in https://github.com/dask/dask-jobqueue/issues/328#issuecomment-542486173 it feels like we may have made the wrong choice mostly because it was based on @mrocklin's experience on Summit which may not be representative of other LSF clusters.

Also for further reference could you answer this two questions:
* what is the output of `lsid | head -n1` 
* does `bsub < job_script` (rather than `bsub job_script`) work for you?

stuarteberg:
@lesteve

**Disclaimer:** I am not an LSF expert, and I only have experience with one LSF cluster, of which I am merely a user, not an administrator.

>do you think we should default to use_stdin=True? 

Yes, I think we should.  If `use_stdin=False`, then `LSFCluster` will [write the jobscript to `/tmp`][1] (or whichever directory `distributed.utils.tmpfile()` chooses), and then launch the command as follows:

[1]: https://github.com/dask/dask-jobqueue/blob/b4f94c05bb94a90c85e733646272bcd18a5b0529/dask_jobqueue/core.py#L260

```
bsub /tmp/jobscript.sh
```

Unless `tmpfile()` chooses a location on the shared file system (which seems unlikely), that command is going to fail, because the LSF host that actually executes the job has a completely different `/tmp` directory.

When using `bsub < /tmp/jobscript.sh`, (i.e. `use_stdin=True`) then `jobscript.sh` is fed to the LSF scheduler and then passed to the LSF host when the job is executed.  There is no need for the execution host to have access to the original jobscript file, so it doesn't matter where it was written to originally.

Again, I'm not an expert, but if `bsub < /tmp/jobscript.sh` didn't work when @mrocklin tried it at Summit, then my hunch is that there's something weird about Summit's configuration.  It's even stranger that `bsub /tmp/jobscript.sh` apparently worked.

In any case, I don't think our [current default heuristic][2] is correct, because `bsub <` is still supported in LSF 10 ([as documented in the LSF 10 manual][3]).

[2]: https://github.com/dask/dask-jobqueue/blob/b4f94c05bb94a90c85e733646272bcd18a5b0529/dask_jobqueue/lsf.py#L218

[3]: https://www.ibm.com/support/knowledgecenter/en/SSWRJV_10.1.0/lsf_command_ref/bsub.zz1argument.command.1.html

---

>Also for further reference could you answer these two questions:
>
>    - what is the output of `lsid | head -n1`

```
$ lsid | head -n1
IBM Spectrum LSF Standard 10.1.0.8, May 10 2019
```

>    - does `bsub < job_script` (rather than `bsub job_script`) work for you?

Yes, both of those work for me, as long as `job_script` is located in a shared location, such as my home directory.  But (as explained above) if `job_script` is located in `/tmp/`, then only `bsub < /tmp/job_script` works.

lesteve:
> Disclaimer: I am not an LSF expert, and I only have experience with one LSF cluster, of which I am merely a user, not an administrator.

None of us are LSF experts even less LSF administrators ... as someone who has access to a LSF cluster and from your earlier comments I think you qualify as the dask-jobqueue LSF expert ;-).

Thanks a lot your feed-back, it is extremely useful! Also it aligns very much with my understanding of the problem so I think we should switch to use_stdin=True by default.

stuarteberg:
>I think we should switch to `use_stdin=True` by default.

OK, would you like me to open a separate PR for that, or simply change it now, as part of this PR?

conftest.py
```diff
@@ -27,3 +29,18 @@ def pytest_runtest_setup(item):
     if envnames:
         if item.config.getoption("-E") not in envnames:
             pytest.skip("test requires env in %r" % envnames)
+
+
+@pytest.fixture(autouse=True)
```
lesteve: Would it be possible to not use `autouse` here, so that the fixture is explicitly used in test_use_stdin? My preference would be to avoid pytest magic if possible.

lesteve:
> OK, assuming @guillaumeeb agrees, would you like me to open a separate PR for that, or simply change it now, as part of this PR?

If you don't mind, a separate PR would be preferrable. This PR is adding the use_stdin parameter, which was an oversight of #307.

Changing to `use_stdin=True` is a behaviour change that we may want to understand again in six months time. As such I feel it deserves its PR with its dedicated discussion (rather than plenty of comments in separate issues).

conftest.py
stuarteberg: Since `lsf_version()` is called by default (unless `use-stdin` is specified in the config), then this monkey-patch is needed by all tests that instantiate `LSFCluster()`.  That includes every test in `test_lsf.py` and also half of the tests in `test_jobqueue_core.py`.

I'm not a `pytest` expert, but IIUC, we need to use `autouse=True` or we need to add this fixture to every test that needs it in those two files.  Is there some better mechanism I'm missing?

lesteve: Right I missed that. I think we can keep it like this for this PR.

When we switch to use_stdin=True, we should remove the `lsid` logic (and so this autouse fixture). Basically, we thought there was a change in behaviour linked to LSF 10 and my current understanding is that this is not the case but is linked to some quirks on Summit ...

stuarteberg:
>If you don't mind, a separate PR would be preferrable.

I'll open a PR once this one is merged.  (It will touch the same files as this one.)

lesteve:
Thanks I'll merge this one, since this seems perfectly reasonable to me. @guillaumeeb don't hesitate to comment if you think we missed something.

lesteve:
My current thinking is that there are a few fixes that would be nice to include in a release in the near future (say 1-2 weeks) and `use_stdin=True` by default is one of them.

PR #361, opened 2019-10-25
Fix jobqueue system config_name not correctly passed through

### LINKED ISSUE #358: Interface not loaded from config file
Hello!
I've been using dask-jobqueue for quite some time without messing with the default configuration file localted in `~/.config/dask`. Recently. I decided to change the config file `jobqueue.yaml` to avoid copy/pasting the same jupyter cell across notebooks and just call

```python

cluster = SLURMCluster()
```

with this configuration

```yaml
jobqueue:
  slurm:
    name: dask-worker
    cores: 12
    memory: 60GB 
    processes: 1
    interface: 'ib0' 
    local-directory: /home/grivera/scratch
    queue: mpi_short2
    walltime: '01:00:00'
    log-directory: /home/grivera/slurm_logs
```

instead of

```python

cluster = SLURMCluster(
    queue="mpi_short2",
    cores=12,
    memory="60GB",
    processes=1,
    interface="ib0",
    dashboard_address=":6767",
)
```

When using the config file, my workers don't seem to connect at all and eventually they get killed due to time out. To debug this, I printed `cluster.job_script()` for both cases and found the ip used is different, even thought I'm using the same values in both cases.

<details>
<summary> output of job_script when using kwargs </summary>

```bash
#!/usr/bin/env bash

#SBATCH -J dask-worker
#SBATCH -e /home/grivera/slurm_logs/dask-worker-%J.err
#SBATCH -o /home/grivera/slurm_logs/dask-worker-%J.out
#SBATCH -p mpi_short2
#SBATCH -n 1
#SBATCH --cpus-per-task=12
#SBATCH --mem=56G
#SBATCH -t 01:00:00

JOB_ID=${SLURM_JOB_ID%;*}

/home/grivera/miniconda3/envs/Work/bin/python -m distributed.cli.dask_worker tcp://192.168.0.15:33289 --nthreads 12 --memory-limit 60.00GB --name name --nanny --death-timeout 60 --local-directory /home/grivera/scratch --interface ib0
```
</details>

<details>
<summary> output of job_script when using the yaml file </summary>

```bash
#!/usr/bin/env bash

#SBATCH -J dask-worker
#SBATCH -e /home/grivera/slurm_logs/dask-worker-%J.err
#SBATCH -o /home/grivera/slurm_logs/dask-worker-%J.out
#SBATCH -p mpi_short2
#SBATCH -n 1
#SBATCH --cpus-per-task=12
#SBATCH --mem=56G
#SBATCH -t 01:00:00

JOB_ID=${SLURM_JOB_ID%;*}

/home/grivera/miniconda3/envs/Work/bin/python -m distributed.cli.dask_worker tcp://127.0.0.1:43852 --nthreads 12 --memory-limit 60.00GB --name name --nanny --death-timeout 60 --local-directory /home/grivera/scratch --interface ib0
```
</details>

If I omit the `interface` kwarg, the result is the same for both cases (workers can't connect and ip is 127.0.0.1) so I think this parameter is not being loaded from the config file.

guillaumeeb:
Closes #358

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -391,6 +391,7 @@ class JobQueueCluster(SpecCluster):
     )
 
     job_cls = None
+    config_name = None
 
     def __init__(
         self,
@@ -421,9 +422,12 @@ class JobQueueCluster(SpecCluster):
                 "or SGEJob with the job_cls= argument."
             )
 
-        if config_name:
+        if config_name is not None:
+            self.config_name = config_name
+
+        if self.config_name:
             if interface is None:
-                interface = dask.config.get("jobqueue.%s.interface" % config_name)
+                interface = dask.config.get("jobqueue.%s.interface" % self.config_name)
 
         scheduler = {
             "cls": Scheduler,  # Use local scheduler for now
@@ -435,8 +439,8 @@ class JobQueueCluster(SpecCluster):
                 "security": security,
             },
         }
-        if config_name:
-            kwargs["config_name"] = config_name
+        if self.config_name:
+            kwargs["config_name"] = self.config_name
         kwargs["interface"] = interface
         kwargs["protocol"] = protocol
         kwargs["security"] = security
--- a/dask_jobqueue/tests/test_jobqueue_core.py
+++ b/dask_jobqueue/tests/test_jobqueue_core.py
@@ -3,9 +3,11 @@ import shutil
 import socket
 import sys
 import re
+import psutil
 
 import pytest
 
+import dask
 from dask_jobqueue import (
     JobQueueCluster,
     PBSCluster,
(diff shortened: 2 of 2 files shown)


guillaumeeb:
I tried to think how to test this, but nothing really clean comes to my mind, only something involving checking `cluster._kwargs` to check `interface` ... Should we do that?

mrocklin:
Thanks for resolving this @guillaumeeb .  

For testing, maybe we should check the scheduler_spec or worker_spec, to verify that the interface is somewhere in there?

```python
async def test_foo(cleanup);
    with dask.config.set(...):
        cluster = PBSCluster(asynchronous=True)  # no need to await it
        assert "my-interface" in str(cluster.scheduler_spec)
        assert "my-interface" in str(cluster.worker_spec)
```

guillaumeeb:
Thanks for the hint @mrocklin, I tried to add something. Did not pu the second assert because it caused problems as `cluster.worker_spec` was None and also because the original problem arised in Scheduler spec.

dask_jobqueue/core.py
```diff
@@ -421,9 +422,12 @@ def __init__(
                 "or SGEJob with the job_cls= argument."
             )
 
-        if config_name:
+        if config_name is not None:
+            self.config_name = config_name
```
lesteve: Not convinced this is the right fix, not 100% sure though. I need to look into it more ...

My instincts is that there is some unnecessary complicated things with config_name both a class variable at the Cluster level and the Job level ... a bit vague but I'll try to look at it tomorrow in more details.

guillaumeeb: Yeah I agree this might thought about it a little more, but maybe we could do this after fixing this problem? (I know, we'll probably never look at it again ...)

Maybe we should merge this and open another issue to indicate that we parse the config twice and create duplicate class variables?

lesteve: Wow it has been 19 days already ... I have a WIP PR in #366. Let me know if you have any comments.

PR #362, opened 2019-11-04
Improve error message when cores or memory is not specified in cluster class

lesteve:
Small side-effect of #307.

```py
from dask_jobqueue.sge import SGECluster 
cluster = SGECluster() 
```

The error is confusing because it mentions `SGEJob` rather than `SGECluster` ...

```

ValueError: You must specify how much cores and memory per job you want to use, for example:
cluster = SGEJob(cores=8, memory='24GB')
```


mrocklin:
Thanks @lesteve !

PR #366, opened 2019-11-19
Simplify default config_name and job_cls mechanism

### LINKED ISSUE #358: Interface not loaded from config file

lesteve:
Supersedes and closes #361.
Fixes #358.

The idea is to use a similar mechanism for `config_name` and `job_cls` (a class variable is the default value, but it can also be passed as a parameter).

There are also some cleaning up of a few quirks that remained after #307.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -143,14 +143,9 @@ class Job(ProcessInterface, abc.ABC):
 
         super().__init__()
 
-        # TODO move if cores is None exception here too ...
         if config_name is None:
             raise ValueError(
-                "Looks like you are trying to create a class that inherits from dask_jobqueue.core.JobQueueCluster. "
-                "If that is the case, you need to:\n"
-                "- set the 'config_name' class variable to a non-None value\n"
-                "- create a section in jobqueue.yaml with the value of 'config_name'\n"
-                "If that is not the case, please open an issue in https://github.com/dask/dask-jobqueue/issues."
+                "Job needs a config_name"
             )
 
         if job_name is None:
@@ -190,7 +185,7 @@ class Job(ProcessInterface, abc.ABC):
                 )
             )
 
-        # This attribute should be overridden
+        # This attribute should be set in the derived class
         self.job_header = None
 
         if interface:
@@ -391,8 +386,6 @@ class JobQueueCluster(SpecCluster):
         cluster_parameters=cluster_parameters
     )
 
-    job_cls = None
-
     def __init__(
         self,
         n_workers=0,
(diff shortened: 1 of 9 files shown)


dask_jobqueue/core.py
```diff
@@ -414,18 +412,33 @@ def __init__(
...
+                "- you are inheriting from JobQueueCluster (most likely): you need to have a class variable"
+                " in your JobQueueCluster-derived class\n"
+                "- you are using JobQueueCluster directly (less likely, only useful for tests), "
+                "pass a  job_cls= parameter."
             )
 
-        if config_name:
-            if interface is None:
-                interface = dask.config.get("jobqueue.%s.interface" % config_name)
+        default_config_name = getattr(self.job_cls, "config_name", None)
```
guillaumeeb: So the key is here, all the informaton is in the Job class, not in the JobQueueCuster one.

lesteve:
> Okay, this made me a headache to properly understand all the implications here.

I agree this is the kind of PR where it may be easier to look at the code, rather than the diff, because the diff is quite tricky to follow.

My thinking behind this was to simplify the default value mechanism for both `job_cls` and `config_name`. Now they are both coming from a class variable (`job_cls` is a class variable in `*Cluster` and `config_name` is a class variable in `*Job`). Also the `*Cluster` uses the `*Job` default `config_name`. The fact that `config_name` passed in `*Cluster` was not the same than the one in `*Job` was the root cause of #358.

Previously (i.e. in master before this PR is merged):
- config_name was set in `*Cluster` objects but not used. What was used was the default value of `config_name` in the `*Job.__init__`.

I am hoping that this PR cleans up a few things. This could probably be cleaned up further so suggestions more than welcome !

guillaumeeb:
Would have need this explanations!

> Also the *Cluster uses the *Job default config_name

I understand the idea, but it feels slightly weird.

This is a really small concern though, I'm happy to keep it so as I've no good suggestion to offer.

lesteve:
OK let's merge this one then!

PR #375, opened 2019-12-02
Use a mix of processes and threads when processes is not specified

### LINKED ISSUE #332: Should we default to multiple processes (rather than multiple threads)?

lesteve:
Fixes #332, closes #334.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -25,8 +25,10 @@ job_parameters = """
     memory: str
         Total amount of memory per job
     processes : int
-        Cut the job up into this many processes.
-        Good for GIL workloads or for nodes with many cores.
+        Cut the job up into this many processes. Good for GIL workloads or for
+        nodes with many cores.
+        By default, ``process ~= sqrt(cores)`` so that the number of processes
+        and the number of threads per process is roughly the same.
     interface : str
         Network interface like 'eth0' or 'ib0'.
     nanny : bool
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -11,9 +11,10 @@ import abc
 
 import dask
 from dask.utils import ignoring
-from distributed.deploy.spec import ProcessInterface, SpecCluster
-from distributed.scheduler import Scheduler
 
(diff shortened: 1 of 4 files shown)


lesteve:
@jennirinker I took over your PR #334, any comments let me know!

lesteve:
Good point, I will add some documentation in the docstring for the default value of `processes`.

I have to say I don't remember why/how the mix of processes/threads was chosen for `LocalCluster` but I trust there was a good reason behind it.

My main worry is that the default (when specifying only `cores` and not `processes`) is going to use only multithreading and will make `dask-jobqueue` look bad in most cases (Python functions which don't release the GIL).

mrocklin:
The heuristic was intended to find a good balance between the two.  My experience is that machines with 20+ cores don't get full parallelism with only threads, and that separating these into a few processes tends to help.  It could be though that workloads that run on HPC centers tend to be more GIL friendly (more numpy, less text).  I don't know though.

I've never really regretted having large nodes cut up into a few processes, so In general I'm mildly in favor of this change.  I can understand not wanting it too though for explicitness.

lesteve:
I pushed an entry in changelog. I am going to merge this one.

Issue #333, opened 2019-09-06
For loop never ending


eXpensia:
Hello,

I'm using dask to analyse some images, my code looks like this :

```
somefunc(n):
    try:
        long calculation
        return result
    except:
         return to be filtered result

def main():
    num_image = 268828
    X = np.empty((num_image,7), dtype=np.float32)
    cluster = SLURMCluster(cores=32,
                           processes=32,
                           interface ="ib0",
                           walltime="04:10:00",
                           memory = "80GB")
    cluster.scale(600)
    client = Client(cluster)

    with tqdm (total = num_image) as pbar:
        future = client.map(somefunc, range(num_image))
        for future,result in as_completed(future,with_results = True):
            X[result[0],:] = result
            pbar.update(1)
            future.release()

    return X


if __name__ == '__main__':

    x = main()
    io.savemat('/gpfs/home/savedata.mat', {'data': x})

```
The scripts very well excepts for some iterations and it stays stuck without doing some iterations

`100%|█████████▉| 268711/268828 [3:40:11<00:31,  3.68it/s]`

Looking at the log files I found that it first looped on a first error :

```
Traceback (most recent call last):
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/distributed/worker.py", line 875, in heartbeat
    metrics=await self.get_metrics(),
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/distributed/core.py", line 747, in send_recv_from_rpc
    comm = await self.pool.connect(self.addr)
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/distributed/core.py", line 874, in connect
    connection_args=self.connection_args,
  File "/gpfs/home/maxlavaud/miniconda3/envs/env_dask/lib/python3.6/site-packages/distributed/comm/core.py", line 227, in connect
    _raise(error)
[...]

PR #380, opened 2020-02-05
OAR: add async to _submit_job method

msimonin:
_submit_job is awaited by the `Job.start`.

### DIFF
--- a/dask_jobqueue/oar.py
+++ b/dask_jobqueue/oar.py
@@ -71,7 +71,7 @@ class OARJob(Job):
 
         logger.debug("Job script: \n %s" % self.job_script())
 
-    def _submit_job(self, fn):
+    async def _submit_job(self, fn):
         # OAR specificity: the submission script needs to exist on the worker
         # when the job starts on the worker. This is different from other
         # schedulers that only need the script on the submission node at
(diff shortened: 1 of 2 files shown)


lesteve:
Thanks a lot for fixing this! This bug was very likely added with the SpecCluster refactor (#307) and not noticed since the 0.7 release ...

I added an entry in the changelog, so merging this!

lesteve:
As discussed IRL, if you ever get the chance to add CI for OAR, it would be super appreciated!

I would also make `OAR` more of a first citizen in `dask-jobqueue` (as `SLURM`, `SGE` and `PBS`) and help avoid similar problems in the future. We use docker-compose for your CI, you can have a look how it is done for SLURM, SGE and PBS: https://github.com/dask/dask-jobqueue/tree/master/ci.

While I am at it, one of the limitation of `OARCluster`: in the `OAR` cluster I have access to, `oarsub` is not installed on the compute nodes. My understanding is that this is the default setup in `OAR`.

If you create your `OARCluster` on a compute node (rather than on the login node, which is easier but not completely best practice compliant), a work-around as the one mentioned https://github.com/dask/dask-jobqueue/issues/333#issuecomment-530263090 is probably necessary.

If it turns out that `oarsub` is not installed on the compute nodes on your OAR cluster either, we could add a parameter to `OARCluster` e.g. `login_node_host` or something similar to make `OARCluster` easier to use.

PR #384, opened 2020-03-04
Add scheduler_options to pass scheduler-specific parameters

### LINKED ISSUE #355: Specify a cluster scheduler listener port?
I'm trying to use dask-jobqueue on an SGE cluster, and by default our worker nodes cannot communicate back to the login node. The admins have set aside a small range of ports us to use for this, but I can't seem to specify a port/port range for the scheduler to listen on. I see with `dask.distributed`'s `LocalCluster`, you can specify the argument `scheduler_port`, but `SGECluster` does not seem to take the keyword (though it doesn't complain):

```
cluster = SGECluster(queue='short.q',
                     cores=10,
                     memory='20GB',
                     walltime='00:20:00',
                     scheduler_port=8804)
```
`cluster.scheduler` shows the scheduler still being assigned a random port (always around 40,000):

```
<Scheduler: "tcp://169.230.126.85:39008" processes: 0 cores: 0>
```

Is there a way to configure this?

Thanks.

### LINKED ISSUE #382: Using different interfaces for scheduler and workers
I am currently trying to setup a jobqueue cluster in a HPC environment in which available network interfaces on the compute nodes (where Dask workers live) and the login nodes (where the scheduler lives) are not the same.

Available network interfaces

* on the compute nodes there are `lo`, `ib0`, but no `ib1` and no `eth{0,1}`
* whereas on the login nodes there are `lo`, `ib1`, `eth0`, and `eth1`

Starting a jobqueue cluster with a basic workflow of

```python
import dask_jobqueue
import dask.distributed as dask_distributed
jobqueue_cluster = dask_jobqueue.SLURMCluster(cores=6, memory='24GB',
                                              project='esmtst', queue='devel',
                                              interface='ib0')
client = dask_distributed.Client(jobqueue_cluster)
jobqueue_cluster.scale(jobs=1)

```
causes the following error

```python-traceback
ValueError                                Traceback (most recent call last)
<ipython-input-2-b93ee7f2ce8e> in <module>
----> 1 jobqueue_cluster = dask_jobqueue.SLURMCluster(cores=6, memory='24GB',
      2                                               project='esmtst', queue='devel',
      3                                               interface='ib0')

/p/project/cesmtst/hoeflich1/miniconda3/envs/Dask-jobqueue_v2020.02.10/lib/python3.8/site-packages/dask_jobqueue/core.py in __init__(self, n_workers, job_cls, loop, security, silence_logs, name, asynchronous, interface, host, protocol, dashboard_address, config_name, **kwargs)
    446             worker["group"] = ["-" + str(i) for i in range(kwargs["processes"])]
    447
--> 448         self._dummy_job  # trigger property to ensure that the job is valid
    449
    450         super().__init__(

/p/project/cesmtst/hoeflich1/miniconda3/envs/Dask-jobqueue_v2020.02.10/lib/python3.8/site-packages/dask_jobqueue/core.py in _dummy_job(self)
    473         except AttributeError:
    474             address = "tcp://<insert-scheduler-address-here>:8786"
--> 475         return self.job_cls(
    476             address or "tcp://<insert-scheduler-address-here>:8786",
    477             name="name",

/p/project/cesmtst/hoeflich1/miniconda3/envs/Dask-jobqueue_v2020.02.10/lib/python3.8/site-packages/dask_jobqueue/slurm.py in __init__(self, queue, project, walltime, job_cpu, job_mem, job_extra, config_name, *args, **kwargs)
     39             job_extra = dask.config.get("jobqueue.%s.job-extra" % config_name)
     40
---> 41         super().__init__(*args, config_name=config_name, **kwargs)
     42
     43         header_lines = []

/p/project/cesmtst/hoeflich1/miniconda3/envs/Dask-jobqueue_v2020.02.10/lib/python3.8/site-packages/dask_jobqueue/core.py in __init__(self, scheduler, name, cores, memory, processes, nanny, interface, death_timeout, local_directory, extra, env_extra, header_skip, log_directory, shebang, python, job_name, config_name, **kwargs)
    195         if interface:
    196             extra = extra + ["--interface", interface]
--> 197             kwargs.setdefault("host", get_ip_interface(interface))
    198         else:
    199             kwargs.setdefault("host", "")

/p/project/cesmtst/hoeflich1/miniconda3/envs/Dask-jobqueue_v2020.02.10/lib/python3.8/site-packages/distributed/utils.py in get_ip_interface(ifname)
    181     if ifname not in net_if_addrs:
    182         allowed_ifnames = list(net_if_addrs.keys())
--> 183         raise ValueError(
    184             "{!r} is not a valid network interface. "
    185             "Valid network interfaces are: {}".format(ifname, allowed_ifnames)

ValueError: 'ib0' is not a valid network interface. Valid network interfaces are: ['lo', 'ib1', 'eth0', 'eth1']
```

Workaround as suggested in https://github.com/dask/dask-jobqueue/issues/207#issuecomment-566397581 doesn't work, because Dask distributed doesn't like interface and host to be set at the same time.

lesteve:
fix #355 (regression in 0.7 where `scheduler port` was ignored), fix #382 (different interfaces on scheduler and workers).

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -69,8 +69,12 @@ cluster_parameters = """
         Whether or not to run this cluster object with the async/await syntax
     security : Security
         A dask.distributed security object if you're using TLS/SSL
-    dashboard_address : str or int
-        An address like ":8787" on which to host the Scheduler's dashboard
+    scheduler_options : dict
+        Used to pass additional arguments to Dask Scheduler. For example use
+        ``scheduler_options={'dasboard_address': ':12435'}`` to specify which
+        port the web dashboard should use or ``scheduler_options={'host': 'your-host'}``
+        to specify the host the Dask scheduler should run on. See
+        :class:`distributed.Scheduler` for more details.
 """.strip()
 
 
@@ -420,13 +424,15 @@ class JobQueueCluster(SpecCluster):
         silence_logs="error",
         name=None,
         asynchronous=False,
-        # Scheduler keywords
-        interface=None,
+        # Scheduler-only keywords
+        dashboard_address=None,
         host=None,
+        scheduler_options=None,
+        # Options for both scheduler and workers
+        interface=None,
         protocol="tcp://",
-        dashboard_address=":8787",
-        config_name=None,
         # Job keywords
+        config_name=None,
         **kwargs
     ):
         self.status = "created"
@@ -447,22 +453,47 @@ class JobQueueCluster(SpecCluster):
                 )
             )
 
+        if dashboard_address is not None:
+            raise ValueError(
+                "Please pass 'dashboard_address' through 'scheduler_options': use\n"
+                'cluster = {0}(..., scheduler_options={{"dashboard_address": ":12345"}}) rather than\n'
+                'cluster = {0}(..., dashboard_address="12435")'.format(
+                    self.__class__.__name__
+                )
+            )
+
+        if host is not None:
+            raise ValueError(
+                "Please pass 'host' through 'scheduler_options': use\n"
+                'cluster = {0}(..., scheduler_options={{"host": "your-host"}}) rather than\n'
+                'cluster = {0}(..., host="your-host")'.format(self.__class__.__name__)
+            )
+
         default_config_name = self.job_cls.default_config_name()
         if config_name is None:
             config_name = default_config_name
 
         if interface is None:
             interface = dask.config.get("jobqueue.%s.interface" % config_name)
+        if scheduler_options is None:
+            scheduler_options = {}
+
+        default_scheduler_options = {
+            "protocol": protocol,
+            "dashboard_address": ":8787",
+            "security": security,
+        }
+        # scheduler_options overrides parameters common to both workers and scheduler
+        scheduler_options = dict(default_scheduler_options, **scheduler_options)
+
+        # Use the same network interface as the workers if scheduler ip has not
+        # been set through scheduler_options via 'host' or 'interface'
+        if "host" not in scheduler_options and "interface" not in scheduler_options:
+            scheduler_options["interface"] = interface
 
         scheduler = {
             "cls": Scheduler,  # Use local scheduler for now
-            "options": {
-                "protocol": protocol,
-                "interface": interface,
-                "host": host,
-                "dashboard_address": dashboard_address,
-                "security": security,
-            },
+            "options": scheduler_options,
         }
 
         kwargs["config_name"] = config_name
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -15,7 +15,7 @@ from dask.utils import ignoring
 from distributed.deploy.spec import ProcessInterface, SpecCluster
 from distributed.deploy.local import nprocesses_nthreads
 from distributed.scheduler import Scheduler
-from distributed.utils import format_bytes, parse_bytes, tmpfile, get_ip_interface
+from distributed.utils import format_bytes, parse_bytes, tmpfile
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


dask_jobqueue/core.py
```diff
@@ -447,22 +453,47 @@ def __init__(
                 )
             )
 
+        if dashboard_address is not None:
```
lesteve: Not allowing to pass `host` and `dashboard_address` is a breaking change. I guess it feels cleaner to be able to pass all the scheduler options through `scheduler_options`. Also if we keep `host` for example, you can pass different things at different levels, e.g. `host` + `interface` in `scheduler_options` which would make the error hard to make user-friendly (and/or the code a bit ugly).

If you think that's too harsh, I can make it a warning until the 0.8 release!

lesteve:
Maybe @willirath you could have a quick look if you are not too busy with other things?

dask_jobqueue/core.py
guillaumeeb: What feel's weird here is to keep the `host` and `dashboard_address` in the constructor kwargs, but thrown an exception when used. Maybe we should add some doc or comment, or simply remove them and catch them through `**kwargs` argument? Or else we should handle it and log a deprecation warning.

I'm not really familliar with Python best practices in this case.

dask_jobqueue/core.py
```diff
@@ -447,22 +453,47 @@ def __init__(
...
+            "dashboard_address": ":8787",
+            "security": security,
+        }
+        # scheduler_options overrides parameters common to both workers and scheduler
+        scheduler_options = dict(default_scheduler_options, **scheduler_options)
+
+        # Use the same network interface as the workers if scheduler ip has not
+        # been set through scheduler_options via 'host' or 'interface'
+        if "host" not in scheduler_options and "interface" not in scheduler_options:
+            scheduler_options["interface"] = interface
```
guillaumeeb: Should'nt we just put interface in the `default_scheduler_options`? Won't it be overriden if defined in the `scheduler_options`?

Or is this because of a problem in distributed if host and interface do not match?

dask_jobqueue/core.py
lesteve: I can see two solutions:
1. raise an informative error if people are using `host` or `dasboard_address` to tell them to pass this through `scheduler_options`. What I had in mind was to have this user-friendly error in 0.7.1 and remove `host` and `dashboard_address` in 0.8.
2. warn if people are using `host` or `dashboard_address` telling them that this will not be supported in 0.8 and then combine them with `scheduler_options` to do the right thing.

I tend to think 1. is easier because you don't need the combining logic between `host` `dashboard_address` and `scheduler_options` which can be tricky. For example, if you use `host` and use `scheduler_options={'interface': eth0}` which one should have the priority and should we raise an error in this case.

Also my feeling is that people ignore warnings until it breaks so that 2. is not that helpful for users in the end.

> I'm not really familliar with Python best practices in this case.

It completely depends on the project. For example, in scikit-learn we have a strong backward-compatibility policy: we would have warnings for two major releases before we can break user code. I feel in `dask-jobqueue` we can afford to be less conservative.

dask_jobqueue/core.py
lesteve: I should probably test this edge case indeed. My line of reasoning is like this (as I tried to explain in the comment):
- `interface` parameter applies primarily to workers. It should apply to the scheduler only if the scheduler IP is not already set by either `host` or `interface` in the `scheduler_options` dict.

dask_jobqueue/core.py
```diff
@@ -202,9 +206,6 @@ def __init__(
 
         if interface:
             extra = extra + ["--interface", interface]
-            kwargs.setdefault("host", get_ip_interface(interface))
```
lesteve: After having looked a bit more in details, I am reasonably sure that `host` in kwargs is not used at all, i.e. this is another manifestation of #386. 

This code was introduced in https://github.com/dask/dask-jobqueue/pull/135. At that time the `kwargs` were passed to `LocalCluster`. Nowadays the kwargs are passed to the `job_cls`. This does not make any sense to pass `host=something` to the worker arguments where `something` is computed where on the node the scheduler runs (similar comment as https://github.com/dask/dask-jobqueue/issues/382#issuecomment-596915307) ...

lesteve:
I think this one should be ready to merge, if someone has some spare time to look at it, this would be much appreciated!

If I don't hear back, I plan to merge this one in a few days.

lesteve:
OK I am going to merge this one, do feel free to comment on this PR (even after the fact) if you have some improvements to suggest!

PR #388, opened 2020-03-05
Switch to use_stdin=True by default for LSFCluster.

### LINKED ISSUE #372: Switch to use_stdin=True by default in LSFCluster
This was discussed in https://github.com/dask/dask-jobqueue/pull/360#issuecomment-547937121, and other places for example https://github.com/dask/dask-jobqueue/issues/328#issuecomment-542486173.

The summary is:
- use_stdin=False does `bsub job_script` rather than `bsub < job_script`
- `use_stdin=False` was decided based on using `dask-jobqueue` on Summit (part of #307). This does not seem to be representative of other LSF clusters.
- at one point, we thought this was related to LSF 10 but actually not since `bsub < job_script` is mentioned in https://www.ibm.com/support/knowledgecenter/en/SSWRJV_10.1.0/lsf_command_ref/bsub.zz1argument.command.1.html (search for "bsub <" to find the exact place it is mentioned)

lesteve:
Fix #372.

cc @d-v-b, @stuarteberg.

Note: this kind of config file defaults change, needs an action from the user, either:
- edit `jobqueue.yaml` by hand
- delete `jobqueue.yaml` if you have nothing useful there and it will be recreated from scratch with the new defaults

I believe this is good enough but let me know if you have any comments.

### DIFF
--- a/dask_jobqueue/jobqueue.yaml
+++ b/dask_jobqueue/jobqueue.yaml
@@ -140,7 +140,7 @@ jobqueue:
     job-extra: []
     log-directory: null
     lsf-units: null
-    use-stdin: null             # (bool) How jobs are launched, i.e. 'bsub jobscript.sh' or 'bsub < jobscript.sh'
+    use-stdin: True             # (bool) How jobs are launched, i.e. 'bsub jobscript.sh' or 'bsub < jobscript.sh'
 
   htcondor:
     name: dask-worker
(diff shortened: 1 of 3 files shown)


d-v-b:
looks good to me

lesteve:
OK thanks, I am going to merge this one!

Issue #186, opened 2018-10-30
ClusterManager: Reflexion on how to start a remote Scheduler


guillaumeeb:
One of the goal of ClusterManager object is to be able to launch a remote scheduler. In dask-jobqueue scope, this probably means submitting a job which will start a Scheduler, and then connect to it.

We probably still lacks some remote interface between ClusterManager and scheduler object for this to work, so it will probably mean to extend APIs upstream.

Identified Scheduler method to provide:

- retire_workers(n, minimum, key)
- scheduler_info(), already existing, see if sufficient,
- add_diagnostic_plugin(), and mostly retrive plugin information remotely

I suspect that adaptive will need to change significantly too, this will maybe lead to having a transitional adaptive logic in dask-jobqueue, and other remote function to add in scheduler.

This is in the scope of #170.

guillaumeeb:
With #306 we've got a lot of what is needed to start a Scheduler in a dedicated batch job. 

I also believe we've got all what is needed in Distributed side too, as dask-kubernetes don't use a LocalCluster anymore, see https://github.com/dask/dask-kubernetes/pull/162.

mrocklin:
Yes, everything should be set up.  We probably need to abstract out the
FooJob classes to split between scheduler and workers.  All of the
SpecCluster infrastructure is there though.

lesteve:
Just curious could you elaborate a bit more about why this would be useful. I have some guesses but I just want to make sure they are somewhat accurate.
- I am guessing that the main advantage is that the scheduler does not run on the cluster login node anymore (most users around me run their own jupyter notebook server on the login node so the scheduler is on the login node as well, this is probably different with a JupyterHub setup). I do understand that doing that would be considered best practices by HPC sys-admin. I'd be curious if you have met some cases where the scheduler goes wild and consumes a lot more than expected or maybe there are a lot of people having their own Dask scheduler on the login node and it starts to add up.
- IMO rather than the scheduler consuming more resources than it should what is a lot more likely to happen is that the client script will, for example because once you have your results from your job you do some not so light post-processing and you forget you are on the login node and not really supposed to do that.
[...]

PR #390, opened 2020-03-05
[WIP] Scheduler in Cluster

quasiben:
This is really hacky and ongoing work to move the scheduler into the cluster.  By doing this we _should_ be able to more easily run other communication protocols for the dask cluster such as UCX:

Current idea for how this might look:
```python
cluster = SLURMCluster(
    deploy_mode="remote",
    cores=1,
    memory="5GB",
    protocol="ucx",
    env_extra=[
        "UCX_MEMTYPE_CACHE=n",
        "UCX_SOCKADDR_TLS_PRIORITY=sockcm",
        "UCX_TLS=tcp,cuda_copy,sockcm,cuda_ipc",
    ],
)


cluster.scale(1)
client = Client(cluster)
```

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -8,6 +8,7 @@ import subprocess
 import sys
 import weakref
 import abc
+import time
 
 import dask
 from dask.utils import ignoring
@@ -139,8 +140,12 @@ class Job(ProcessInterface, abc.ABC):
         python=sys.executable,
         job_name=None,
         config_name=None,
+        deploy_mode='local',
+        deploy_command='distributed.cli.dask_worker', # can also be distributed.cli.dask_worker
+        protocol='tcp',
         **kwargs
     ):
+
         self.scheduler = scheduler
         self.job_id = None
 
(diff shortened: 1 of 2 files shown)


lesteve:
Interesting so I guess this is related to https://github.com/dask/dask-jobqueue/issues/186. Independently of ucx this might be something that we want at one point ...

quasiben:
Yes. The slurm cluster i have been using does not have Infiniband on the edge node so it forced the issue of the scheduler needing to start in the cluster as part of the job

muammar:
I am having problems with the scheduler being on the login node. It is always killed by admin rules. How can I test this? I would be very happy to try it.

quasiben:
You could checkout this PR and try something like the above but it likely will not work at this stage.

muammar:
> On Mar 25, 2020, at 06:43, Benjamin Zaitlen <notifications@github.com> wrote:
> 
> 
> You could checkout this PR and try something like the above but it likely will not work at this stage.
> 

Thanks for your reply. I think I will just use some of the recommendations in the issue report. 

Best, 

-- 
Muammar El Khatib.
http://muammar.me

guillaumeeb:
@quasiben I know this is quite of an old PR, but this would be very cool to pursue!

quasiben:
This is fairly outdated.  I'm going to close and come back to this another time

PR #392, opened 2020-03-16
Mention HTCondor in main documentation page

matyasselmeci:
Looks like the front page lists all the supported batch systems -- please add HTCondor to the list now that there is support for it.


lesteve:
Thanks, merging this one!

PR #395, opened 2020-03-17
MNT: small CI cleanup

lesteve:
- switch to xenial (current Travis default). This should not have any impact since most of our tests run inside docker
- sudo is true by default, see https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration for more details
- DOCKER_COMPOSE_VERSION and OS env variables are not used
- use latest Dask release rather than development version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
-dist: trusty
-sudo: required
+dist: xenial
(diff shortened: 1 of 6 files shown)


lesteve:
CI is green, merging!

PR #396, opened 2020-03-20
Remove unused JOB_ID variable.

lesteve:
`JOB_ID` was added in https://github.com/dask/dask-jobqueue/pull/63. The idea behind it was that the Dask worker name would contain `JOB_ID` so that a scheduler plugin could link Dask workers to a JOB_ID.


lesteve:
Merging this one!

Issue #386, opened 2020-03-05
Cluster classes accept any named parameters without error


lesteve:
```py
from dask_jobqueue import SLURMCluster
cluster = SLURMCluster(cores=1, memory='4GB', this_parameter='does_not_make_sense')
```

I would expect an error somewhere saying that `this_parameter` is not a valid parameter. Instead it is silently ignored.

I think the cluster classes accept a `**kwargs` that is generally not used. This is probably done like this so that the parameters of `JobQueueCluster` can be used in the derived class. My feeling is that the `**kwargs` can be removed either from `JobQueueCluster` or the derived cluster classes. Even if I am wrong, my hope is that there must be a way to have a good error message.

lesteve:
Quickly jotting down some notes before I forget. The best thing I can see for now:
- have derived Job classes have explicit parameter names (i.e. no kwargs). That means copying all the parameters from Job (base class) in each cluster-specific class
- in `JobQueueCluster` add a mechanism to have an error at the cluster level that the user understands. Mechanism: either with inspect.signature or `try: except TypeError:` to detect that some params in the cluster **kwargs do not exist in the `job_cls`.

PR #397, opened 2020-03-23
JobQueueCluster: pass security along

bfis:
`JobQueueCluster.__init__` needs to pass the `security` argument to it's super's (`SpecCluster`) `__init__` since it is needed there too.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -514,6 +514,7 @@ class JobQueueCluster(SpecCluster):
             loop=loop,
             silence_logs=silence_logs,
             asynchronous=asynchronous,
+            security=security,
             name=name,
         )
 


lesteve:
Thanks for your PR!

I have noticed that recently too. I have some ongoing branch with a broader scope trying to tackle #386 and fixing some quirks while doing that. You can see the changes: 
https://github.com/dask/dask-jobqueue/compare/master...lesteve:simplify-job-kwargs

It may take a week before I manage to clean it up a bit and add tests though, plus maybe one week of review.

Here are a few of the options that you have available:
* you manage to extract the Security fix from my branch and add them to this PR with some tests
* if you have some spare bandwidth fixing `protocol` as well would be very welcome (probably a bit easier than adding `security`)
* you help me on my branch by doing PRs on my branch, for example by testing it on your cluster and/or adding more tests

By the way if you have a use case for using `security` in a HPC context I would be quite interested to hear more about it!

PR #403, opened 2020-04-01
Dask JobQueue docs

jerowe:
I think it would be helpful for people coming from an HPC background to have some "this is how dask submits your HPC jobs and runs Dask tasks" docs.


guillaumeeb:
Hi @jerowe,

Not sure if you're still thinking it's work adding things in the docs?

We tried to do something along what you describe in https://jobqueue.dask.org/en/latest/howitworks.html. Do you thing some more should be added?

PR #421, opened 2020-04-30
Improve consistency of SGE cluster object with other clusters

kathoef:
I tried to access `*Cluster.config_name` for all available clusters during a test scenario currently developed [here](https://github.com/dask/dask-jobqueue/pull/405#discussion_r418159072). This was possible for all clusters, except SGE. To have the cluster objects consistent the attribute should be added, I think.

### DIFF
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -120,3 +120,4 @@ class SGECluster(JobQueueCluster):
         job=job_parameters, cluster=cluster_parameters
     )
     job_cls = SGEJob
+    config_name = "sge"


willirath:
Looks like an opportunity to try team mentions:

@dask/sge, this one is for you.

lesteve:
Looks like this was some forgotten clean-up from #366 and cluster classes are not supposed to have the `config_name` class parameter ...

If you need it from the cluster class, you can use `self.job_cls.default_config_name()`.

I am going to close this one and merge #426 instead.

kathoef:
Great, that looks like a better thing to do!

PR #420, opened 2020-04-25
Add minimal htcondor CI support

riedel:
I set up a really minimal ci/htcondor.sh to address #247. No other files needed sofar.

A test fails, but this seems to be actually a real issue: 
> 'HTCondorCluster' object has no attribute 'pending_jobs'

Please give it a try!

I thought I'd rather do sth if you add me to the team ;)

PR #423, opened 2020-05-02
Fix mysql version to get Slurm CI green

riedel:
I just checked what happened when suddenly the slurm CI stopped working. The only relevant thing I saw was a version bump in the mysql docker image.

I'll close this PR immediately if Travis does not work...

### DIFF
--- a/ci/slurm/docker-compose.yml
+++ b/ci/slurm/docker-compose.yml
@@ -2,7 +2,7 @@ version: "2.2"
 
 services:
   mysql:
-    image: mysql:5.7
+    image: mysql:5.7.29
     hostname: mysql
     container_name: mysql
     environment:


riedel:
qed ;)

willirath:
I suggest we merge this immediately and have a closer look at making these environment changes easier to spot in a separate PR. (See #424.)

willirath:
> The only relevant thing I saw was a version bump in the mysql docker image.

Just so I understand wether I'm kind of blind here or not: Did you get this info directly from the Travis logs? Or did you check what happened with the relevant images on Docker hub?

riedel:
I checked the Travis log because I wanted to get #420 done and just looked at the date when it broke. I first thought it was the slurm image but it only changed the README. Because the other stuff (dependencies) seemed too difficult and I checked the only other image referenced and there was a version bumb that exactly coincided with the the time Travis broke. So I gave it a quick shot  and then pushed it. I actually did not know anyone was working on it until I pushed it: I was just frustrated that my PR wasn't green   :) .

riedel:
BTW: I still have no clue what actually happens. It was just naive luck and someone should probably look into it. But maybe it is rather an issue for the docker image (don't think it depends on anything specific).

lesteve:
I am very happy to see this green, thanks a lot!

I am also very happy not to look into what changed in mysql and to ignore this for now ... there is also something related to Travis-CI because I could not reproduce locally ...

PR #425, opened 2020-05-03
Make CI more robust

willirath:
- [x] make SLURM CI more inspectable
- [x] pin slurm docker images more aggressively
- ~~[x] pin other docker images more aggressively~~
- [x] apply to other schedulers as well

Things I could not address here:
- Pinning of <https://github.com/giovtorres/slurm-docker-cluster> / <https://hub.docker.com/r/giovtorres/slurm-docker-cluster> wasn't done, because there's only the `latest` tag and I'm not sure if I can pin a digest here and just hope for it being available on Dockerhub forever.
- Pinning of the `ubuntu:14.04` images used in SGE wasn't done, because I could not find the exact tag on Dockerhub.


willirath:
With b9fb801, the [SLURM job in Travis](https://travis-ci.org/github/dask/dask-jobqueue/jobs/682557218) will fail, but will tell us more about the env before it does so.

riedel:
I tried similar stuff for the HTCondor CI (will wait with a PR), after looking at what you proposed. A few comments from my experience.

Doing docker run before listing the images can be avoided by doing docker(-compose) pull (and build) before doing the listing. After starting up docker ps can be done. I would not group things to much as a standard as it overcomplicates the startup procedure (at least for me). Things could already go wrong when doing the docker-compose up (to me that would be the natural place to actually start up the whole cluster).

Why not put the set -x outside the functions in travis? Then it would be default for all CIs (can still turn it off for anoying stuff).

willirath:
Fair points, looking at the code now, it seems my changes did make CI logs more verbose, but also didn't really improve readability of the CI code... :/

I'll transfer to draft stage again to prevent accidental merging and try and straighten out the logic first.

lesteve:
I tried ages ago to add `set -e` everywhere and failed, this kind of reminds me of it a little bit (I could not find the PR but I still have an old branch https://github.com/lesteve/dask-jobqueue/tree/add-set-e-in-bash-scripts) ...

Anything that can help debug CI problems is more than welcome!

With the recent pain from the Slurm CI, docker-compose logs where the most useful thing to look at IMO. So maybe they could be printed at the end if we did not manage to start the cluster ?

Also we should not wait for 50 minutes before the build is marked as failed ... Maybe 1-2 minutes without being able to start the toy cluster should be enough ...

guillaumeeb:
@willirath do you think this PR should be kept open?

willirath:
No, let's close this.

(And thanks @guillaumeeb and @andersy005 for the housekeeping session!)

Issue #356, opened 2019-10-17
List of common cluster specific (but not job scheduler specific) quirks with work-arounds if available


lesteve:
I wish there were some list of cluster configuration quirks (that are not job scheduler specific) and possible work-arounds (when there are some) somewhere in the doc (I was not aware of limitation of TCP/IP connection limitations between login and compute nodes in some clusters until a few days ago). Here are a rough list off the top of my head:

* submit_command not available on the compute nodes, e.g. #333. Possible work-around: https://github.com/dask/dask-jobqueue/issues/333#issuecomment-530263090 (I never tried it myself). This is the case for all the OAR clusters I know about, i.e. the submit command is never available on the compute nodes 
* TCP/IP restrictions between login and compute nodes e.g. #354 and #355. Possible work-around: start the main script / notebook in an interactive node with all the additional pain and limitations this entails, see https://github.com/dask/dask-jobqueue/issues/354#issuecomment-542879534 for the one I know about.
* non uniform network interfaces on login and compute nodes. I guess same work-around as TCP/IP restriction would work but not a great work-around.

Please add more if you know more off the top of your head.

cc @mrocklin @guillaumeeb @jhamman

mrocklin:
There are a few here: https://blog.dask.org/2019/08/28/dask-on-summit

lesteve:
> There are a few here: blog.dask.org/2019/08/28/dask-on-summit

Good point!

PR #430, opened 2020-05-18
Kick-off doc section about common work-arounds.

lesteve:
This is about https://github.com/dask/dask-jobqueue/issues/356 and also because recently #196 led me to realise there was no nice place where to fit his proposed work-around for job arrays


lesteve:
I am going to merge this one to kick off this common work-arounds page.

@guillaumeeb or @willirath if you happen to have five minutes to look at this page and have some suggestions, this would be more than welcome!

guillaumeeb:
Thanks @lesteve.

I was also wondering were to put the documentation for #122. But I'm not sure this is a "work-around", this sounds more like a "how to that" chapter. Do you think we should add another section, or just rename this new one to something more general?

lesteve:
#122 feels like a work-around to me but I agree work-around is not a great name.

This could be made more general like cookbook or advanced tips and tricks or a better name if you can think of one ...

guillaumeeb:
Yeah, I prefer something like "Advanced tips and tricks" or something like that,

"Advanced pratices" ?
just "Tips and Tricks" ?
"Advanced use cases" ?

lesteve:
Advanced tips and tricks works for me, let me do the renaming.

lesteve:
Renaming done in https://github.com/dask/dask-jobqueue/commit/bb89969d9c6cabcae66c900409872bb29e060e62.

PR #434, opened 2020-05-26
Add submit_command_extra and cancel_command_extra arguments

### LINKED ISSUE #411: Add `submit_command_extra` and `cancel_command_extra` to HTCondor
Primarily for HTCondor it might be important to pass extra parameters, such as schedd.

As for specifying the schedd, I think @lesteve's suggestion of `submit_command_extra` and `cancel_command_extra` is a good way of solving this problem; that way we don't have to argue over which arguments are "important".

_Originally posted by @matyasselmeci in https://github.com/dask/dask-jobqueue/pull/319#issuecomment-541160495_

This probably should be added across all schedulers: see also #202

matyasselmeci:
`submit_command_extra` and `cancel_command_extra` (config options
`jobqueue.X.submit-command-extra` and `jobqueue.X.cancel-command-extra`)
are lists of strings that are arguments passed on as-is to `condor_submit`
and `condor_rm` when submitting or removing HTCondor jobs.

Fixes #411.


dask_jobqueue/htcondor.py
```diff
@@ -94,6 +96,24 @@ def __init__(
...
+            HTCondorJob.submit_command + " " + shell_quote_list(submit_command_extra)
+        )
+
+        if cancel_command_extra is None:
+            cancel_command_extra = dask.config.get(
+                "jobqueue.%s.cancel-command-extra" % self.config_name, []
+            )
+
+        self.cancel_command = (
+            HTCondorJob.cancel_command + " " + shell_quote_list(cancel_command_extra)
```
riedel, line 114: not so sure if I see the benefit over directly calling `" ".join(shlex.quote(arg) for arg in cancel_command_extra)`

dask_jobqueue/tests/test_htcondor.py
```diff
@@ -26,6 +27,8 @@ def test_job_script():
         disk="100MB",
         env_extra=['export LANG="en_US.utf8"', 'export LC_ALL="en_US.utf8"'],
         job_extra={"+Extra": "True"},
+        submit_command_extra=["-verbose"],
```
riedel, line 30: I would actualy use one test with multiple arguments like ["-batchname","foo"] (I know its kinda rediculous, but it would test the join"

riedel:
I tested a bit more and have a problem I cannot understand.
```
cluster = HTCondorCluster(cores=1, memory="1GB", disk="100MB", submit_command_extra=["-name","schedd2"] ,cancel_command_extra=["-name", "schedd2"])

cluster.scale(1)
cluster.scale(0)
```
strangely leaves stale jobs. Any clue?

riedel:
@matyasselmeci , can you reproduce my problem with the "-name" argument?  

Any clue how to effectively check that the params have been passed? Your idea from https://github.com/dask/dask-jobqueue/issues/411#issuecomment-613032232 was actually really good I think. But this won't do for the cancel_command I guess.

matyasselmeci:
`Job._close_job()` appears to be a class method: https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py#L343-L348
so it's not using the instance-specific cancel_command attributes.

lesteve:
> `Job._close_job()` appears to be a class method: [/dask_jobqueue/core.py@`master`#L343-L348](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py?rgh-link-date=2020-05-27T21%3A45%3A21Z#L343-L348)
> so it's not using the instance-specific cancel_command attributes.

Sorry I missed this one (don't hesitate to ping me if you need input in general), no objection to make it a normal method rather than a class method.

matyasselmeci:
Sounds good, thanks!

One idea for a test case for cancel_command is to add `-name {wrong_name}` as an argument and watch condor_rm fail; then replace that with `-name {right_name}` and watch it succeed.

riedel:
thx!! Ping me or remove the draft status, if you like me to review/test it .

riedel:
One thought: It might make sense to split this PR and merge it in 2 batches to see that the tests are working as expected, but I can also manually revert the class variable part and test it locally. I think it might be a bit of a hassle otherwise, as all this is logically connected.  

Maybe the title of the PR should mention that this only adds the commands to HTCondor (should maybe add more tickets to remind the other groups to add this to their schedulers if it makes sense or track the status across schedulers in the ticket and ask for comments,)

lesteve:
> should maybe add more tickets to remind the other groups to add this to their schedulers if it makes sense or track the status across schedulers in the ticket and ask for comments,

It seems to me that this is very HTCondor specific and I have not seen anyone express a  need for more flexibility along this axis in other job schedulers.

I agree there it would be great to keep some uniformity between schedulers (mostly for stability/maintenance reasons) but at the same time I think it is fine having some scheduler-specific developments if there is a strong use case behind it.

riedel:
LGTM. Tested it on our cluster with the sched arg manually. Anything left to do? If you remove the draft status the CI should also run. Can't wait for the merge...

dask_jobqueue/tests/test_htcondor.py
```diff
@@ -75,6 +83,92 @@ def test_basic(loop):
...
+        with Client(cluster) as client:
+
+            cluster.scale(2)
+
+            client.wait_for_workers(2)
+
+            cluster.scale(0)
+
+            start = time()
+            while client.scheduler_info()["workers"]:
```
riedel, line 158: one last thing: does this loop really sense? I first thought I understand but now I am unsure: It checks until the time out that the workers remain in the queue? But why does it assert `not workers`?

matyasselmeci: You're right, I got the test backwards.

riedel:
seems like there are some side effects to making the method an instance method: #442 (I just cherry-picked your commit, anyone working on this should probably do the same to distangle the problem). 

There seem to be some unwanted leftovers: I didn't see that when manually testing if the arguments work (calling the cancel command seemed to work fine for me...)

BTW: @lesteve: can we turn on github actions for draft PRs?

lesteve:
> BTW: @lesteve: can we turn on github actions for draft PRs?

It seems like the CI ran fine. Did you mean "turn off" rather than "turn on"?

riedel:
> It seems like the CI ran fine. Did you mean "turn off" rather than "turn on"?
it ran after @matyasselmeci marked it as ready for review. That is why the problem with the instance method only became apparent now. It did not run until deccacb but on ee54cd6. 

But don't mind, I looked at the CI and this seems to be a problem with Github actions recently: https://github.community/t/github-actions-not-being-triggered-on-draft-pr/16390 . I guess it also OK to put [WIP] in the title if you want CI but no merge until this is fixed.

guillaumeeb:
Anybody still planning to work on this?

riedel:
Just to document it, the workaround why I personally gave up was that HTCondor can actually handle crucial extra-args that cannot be passed via the job description also via the the environment variable so I am doing this instead:

`os.environ["_condor_SCHEDD_HOST"] = "login-l.sdil.kit.edu"`

Still the problem that @matyasselmeci  ran into are really strange and it seems like some kind of bug that should be documented (maybe even in dask distributed).

It for sure needs some in depth debugging effort.  Should I open a ticket regarding the problem of making job_close an instance method? And then rather close this PR for now (can be reopened if ticket is then fixed)

guillaumeeb:
> Still the problem that @matyasselmeci ran into are really strange and it seems like some kind of bug that should be documented (maybe even in dask distributed).

You think it could come from dask.distributed?

> It for sure needs some in depth debugging effort. Should I open a ticket regarding the problem of making job_close an instance method? And then rather close this PR for now (can be reopened if ticket is then fixed)

Do we need this now for anything else that this problem? Or is this just because you think it would be cleaner?

riedel:
>You think it could come from dask.distributed?

Actually it was good that you asked! It triggered me to look into the design. I actually now figured out a way to do this in #514 . It seems that the `_close_job `is a class method by design, probably because it is used in the weakref finalizer (guessing) . The actual interface was `close `, so it was rather easy to move the cancel_command derefferencing into this. 

I however now have the huge problem of getting @matyasselmeci tests to run properly. Everything seems to be just fine, however, the new test design that actually uses failing submit_commands is really a challenge, because I somehow cannt figure out a way how to really catch that.

riedel:
I hope it is OK if I close this one in favour of #514 , which includes the majority of changes and has been reviewed.

PR #439, opened 2020-06-06
Tweak name in cluster.job_script().

lesteve:
I was quite surprised by this in https://github.com/dask/dask-jobqueue/issues/196#issuecomment-630292957 at one point and forgot that `cluster.job_script()` is not exactly the same script that is submitted for each submission script that runs `dask-worker`.

Also dummy-name is easier to grep than name, so easier to reach the comment as well next time I forget about this and need to be remembered.


lesteve:
Merging this one (I just wanted to make sure that this would not break anything), thanks for your approval @willirath!

PR #441, opened 2020-06-12
Add docker-compose htcondor setup

### LINKED ISSUE #427: Make HTCondor CI a bit closer to the other CI
In an ideal world, it would be great if HTCondor CI was a bit closer to the other CI, in particular:
- have a Dockerfile with the dependencies (`dask`, `distributed` etc ...) installed in the Dockerfile
- maybe follow `start-slurm.sh` etc ... convention but maybe this is overkill since HTCondor use a single Docker image rather than a `docker-compose` setup so I am a bit undecided on this last point.

_Originally posted by @lesteve in https://github.com/dask/dask-jobqueue/pull/420#issuecomment-624047047_

(following up on #247)

riedel:
fixes #427

### DIFF
--- a/ci/htcondor/docker-compose.yml
+++ b/ci/htcondor/docker-compose.yml
@@ -1,0 +1,53 @@
+version: "3.4"
+
+services:
+  cm:
+    hostname: cm.htcondor
+    environment:
+      - USE_POOL_PASSWORD=yes
+    volumes:
+      - secrets:/root/secrets
+    image: htcondor/cm:el7
+
+  submit:
+    build:
+      context: .
+      target: submit
+    hostname: submit.htcondor
+    environment:
+      - CONDOR_HOST=cm
+      - USE_POOL_PASSWORD=yes
+    depends_on:
+      - cm
+    volumes:
+      - secrets:/root/secrets
+      - ../..:/dask-jobqueue
+
+  execute1:
+    build:
+      context: .
+      target: execute
+    hostname: execute1.htcondor
+    environment:
+      - CONDOR_HOST=cm
+      - USE_POOL_PASSWORD=yes
+    depends_on:
+      - cm
+    volumes:
+      - secrets:/root/secrets
+
+  execute2:
+    build:
+      context: .
+      target: execute
+    hostname: execute2.htcondor
+    environment:
+      - CONDOR_HOST=cm
+      - USE_POOL_PASSWORD=yes
+    depends_on:
+      - cm
+    volumes:
+      - secrets:/root/secrets
+
+volumes:
+  secrets:
--- a/ci/htcondor/docker-compose.yml
+++ b/ci/htcondor/docker-compose.yml
@@ -2,15 +2,16 @@ version: "3.4"
 
 services:
   cm:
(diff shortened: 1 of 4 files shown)


riedel:
@matyasselmeci : hope I used the images in the way they were supposed to be used ;) . 

I wonder if it might be better to go with conda and a shared filesystem instead of preinstalling python3 and dask on the exec nodes.

Its mainly still WIP because the docker images should be uploaded.

riedel:
@lesteve  Can you build the images (I left out this commit on purpose: 273911b24804ccd281db7376d74df13c774e15ab) and merge this? One remaining question: I left the master untouched and use the official image: change this for stability?

ci/htcondor/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+version: "3.4"
+
+services:
+  cm:
```
riedel, line 4: might want to introduce a dummy build here

ci/htcondor/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    hostname: cm.htcondor
+    environment:
+      - USE_POOL_PASSWORD=yes
+    volumes:
+      - secrets:/root/secrets
+    image: htcondor/cm:el7
```
riedel, line 10: might want to do a dummy build and use our own image here

guillaumeeb (approved):
> @lesteve Can you build the images (I left out this commit on purpose: 273911b) and merge this?

@riedel is there something else to do here? This looks good to me!

riedel:
yes, it's WIP because it does not include pre-build images yet. I think for bootstrapping they need to be uploaded once (not sure if one could trigger the autobuild). To each image a line needs to be added like: https://github.com/dask/dask-jobqueue/blob/69f27ac71db953ce50b3dd9bbe86c48ef17e5b89/ci/slurm/docker-compose.yml#L19 . The images are missing here: https://hub.docker.com/r/daskdev/dask-jobqueue/tags . Easiest would be if someone with push rights for docker would add the image lines and then put it into this PR and merge it.

riedel:
@lesteve : I removed [WIP] because it was a bit misleading and I wanted to highlight this is ready for merge. Can you look at the changes and add the docker image lines and push them.

I can be merged without that change, however then the images are simply not cached (this is also the current state:so)

I could do the change but then the CI would give a false positive.

lesteve:
@riedel thanks for pushing this forwar and sorry I don't have too much time to devote to dask-jobqueue these days.

I pushed the images in dockerhub: https://hub.docker.com/r/daskdev/dask-jobqueue/tags?ordering=last_updated, `daskdev/dask-jobqueue:htcondor-submit` and `daskdev/dask-jobqueue:htcondor-execute`.

I adapted the `docker-compose.yml` to use these images in my last commit [`9e7dfba` (#441)](https://github.com/dask/dask-jobqueue/pull/441/commits/9e7dfba7463695b0920767cc9b3f870d0447649c).

Let me know if you have any comments!

lesteve:
OK, merging this one, thanks!

PR #442, opened 2020-06-16
[WIP] cherry-pick Turn core.Job._close_job() into an instance method so we can override…

riedel:
… cancel_command in the instance

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -353,11 +353,10 @@ class Job(ProcessInterface, abc.ABC):
         logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
         self._close_job(self.job_id)
 
-    @classmethod
-    def _close_job(cls, job_id):
+    def _close_job(self, job_id):
         if job_id:
             with ignoring(RuntimeError):  # deleting job when job already gone
-                cls._call(shlex.split(cls.cancel_command) + [job_id])
+                self._call(shlex.split(self.cancel_command) + [job_id])
             logger.debug("Closed job %s", job_id)
 
     @staticmethod


lesteve:
Hmmm not sure why this is not working, I would suggest debugging this locally ...

There are instructions there (they could probably be improved a lot):
https://jobqueue.dask.org/en/latest/develop.html#test-with-job-scheduler

I don't really do it like this if I am honest but more like this:
```bash
cd ci/slurm
./start-slurm.sh

docker exec -it slurmctld /bin/bash -c "pytest /dask-jobqueue/dask_jobqueue --verbose -E slurm -s"
```

riedel:
Since the schedulers are a bit complex for simple debugging what do you think about using the really old unix atq (https://en.wikipedia.org/wiki/At_(command)) for such debugging. If there is general interest I would quickly write a driver that should the also fail and would be easy to use for anyone and good for debugging such problems since the LocalCluster does not fail.

lesteve:
I have to admit I don't really follow sorry ...

A few suggestions hoping they may help although maybe you have tried this already:
- use the good old `print` debugging method as a starting point and make sure the `_close_job` method called
- remove the `with ignoring(RuntimeError)` because it may swallow a problem with the killing of the job ...
- why are tests using `dask_jobqueue.local.LocalCluster` not failing? Is that because it still uses `_close_job` as a classmethod? Try to make it a normal method instead.
- why would the cluster object not end up in a closed state? No idea but if you figured that out by looking at the code this would be another way to attack this problem


Side-comments:
- something I try my best to do (and still probably forgot from time to time) is to always specify "Dask scheduler" vs "job scheduler" (i.e. SLURM, PBS, etc ...), this name collisision is quite confusing otherwise.
- another confusing thing is `LocalCluster` that can be either the `dask_jobqueue` or the `distributed` one. We should probably change the name in `dask_jobqueue` to `MockCluster`, `MockLocalCluster` or a better name ...

guillaumeeb:
@riedel are you still planning to do something here?

> what do you think about using the really old unix atq job scheduler as an additional (Mock)LocalCluster Backend

As a quick thought, I would rather not add another MockCluster just for testing...

riedel:
The problem where I stopped was, that actually the none build suceeds, so this is really difficult to debug. Which is truly strange. My guessing was that it had something todo with the forking proceedure: https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/local.py#L57 . However, I have my doubts by now since there seems to be an exec() call triggered according to the docs. I'll close this one to clean up, since the PR is more like a debugging thing anyways...

PR #445, opened 2020-06-17
Fix math typo in GiB definition

rwest:
While reading the documentation at https://jobqueue.dask.org/en/latest/configuration-setup.html
I noticed a strange "1 GiB = 2³0 bytes" and realized it should be "1 GiB = 2³⁰ bytes"
This fixes that typo.


willirath:
Thanks @rwest!

PR #447, opened 2020-06-25
Include Github release in release procedure

### LINKED ISSUE #446: New Github design shows oldest release as latest?
The new side bar seems to get the order of releases wrong. It lists the v0.1 as the latest release:

image

(Not sure if this is something that can be fixed in this repo or if this is a bug with Github.)

willirath:
Closes #446

As Github distinguishes releases and tags, we need to explicitly create a release from a tagged version.

### DIFF
--- a/docs/release-procedure.md
+++ b/docs/release-procedure.md
@@ -21,6 +21,9 @@ git tag -a x.x.x -m 'Version x.x.x'
 git push --tags upstream
 ````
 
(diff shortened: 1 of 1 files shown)


willirath:
This would to it: https://github.com/actions/create-release#example-workflow---create-a-release

guillaumeeb:
Closing this as per https://github.com/dask/dask-jobqueue/issues/446#issuecomment-650115549.

We're not using github releases anymore.

PR #453, opened 2020-07-26
Updating start_workers to scale in examples

vsoch:
The dask-jobqueue library, for the slurm executor, has been updated to use `scale` instead of `start_workers` and it looks like the docs don't reflect that yet.


lesteve:
Thanks a lot merging!

PR #432, opened 2020-05-21
Use pre-built docker images to speed up CI

riedel:
addressing #424, started with htcondor just to check travis, then will add others

PR #455, opened 2020-07-27
Add GH action to periodically build docker images

### LINKED ISSUE #433: Periodically build images for CI
see  discussion in #432

et up a Travis deploy stage that will be triggered e.g. weekly

andersy005:
Fixes #433 and amends https://github.com/dask/dask-jobqueue/pull/432

For this action to work properly, we need to set both the DOCKER_USERNAME, DOCKER_PASSWORD secrets in the repo settings. However, I currently don't have admin rights on this repo.

### DIFF
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jobqueue: ["htcondor", "pbs", "sge", "slurm", "none"]
+        jobqueue: ["pbs", "sge", "slurm"]
 
     steps:
       - name: Checkout
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -1,6 +1,8 @@
 name: docker-images-build
 
 on:
+  push:
+    branches: "master"
   schedule:
     - cron: "0 0 * * *" # Daily “At 00:00”
 
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -2,7 +2,7 @@ name: docker-images-build
 
 on:
   schedule:
-    - cron: "0 0 * * 0" # Weekly “At 00:00 on Sunday.”
+    - cron: "0 0 * * *" # Daily “At 00:00”
 
 jobs:
   build:
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -1,10 +1,8 @@
 name: docker-images-build
 
 on:
-  push:
-    branches: "*"
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "0 0 * * 0" # Weekly “At 00:00 on Sunday.”
  
 
 jobs:
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -3,7 +3,6 @@ name: docker-images-build
 on:
   schedule:
     - cron: "0 0 * * 0" # Weekly “At 00:00 on Sunday.”
- 
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker-images.yaml
```diff
@@ -0,0 +1,46 @@
+name: docker-images-build
+
+on:
+  schedule:
+    - cron: "0 0 * * 0" # Weekly “At 00:00 on Sunday.”
```
jacobtomlinson, line 5: Why weekly? Why not daily?

andersy005: Good catch. I updated it to daily

jacobtomlinson:
> I wasn't sure whether there was already a specific user or people were using their user credentials for this

I think we have a range of credentials being used in various places. I'm slowly trying to standardise things to try and drop reliance on specific maintainer accounts.

> once it is merged is there a way to trigger the cron GH action manually or should we just wait until the next scheduled time?

It may be good to add `push:` to the list of triggers so that this runs on merge to master too.

lesteve:
I think we have a bit of a weird setup and I am not sure how to simplify it and how to make it play well with GH actions.
- the docker images are only used in our CI (those are the docker-compose to have toy clusters for Slurm, PBS, etc ...). Originally we were building the Docker images inside the CI (so on each commit) which was a bit slow (especially in a PR) and also it is easier for running the test locally to do `docker pull` rather than building the images locally.
- on a push to the master branch you probably want to generate the Docker files and push them to DockerHub before running the CI. AFAIU having separate `ci.yml` and `build-docker-images.yaml` does not make that straightforward so maybe we want a single `.yml`?
- even then you can not really test any change to the toy clusters docker-compose setup (i.e. `ci/slurm`, `ci/pbs`) in a PR but I think that's fine because they change very rarely ...

README.rst
```diff
@@ -23,3 +23,5 @@ New BSD. See `License File <https://github.com/dask/dask-jobqueue/blob/master/LI
    :target: https://gitter.im/dask/dask?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
 .. |Version Status| image:: https://img.shields.io/pypi/v/dask-jobqueue.svg
    :target: https://pypi.python.org/pypi/dask-jobqueue/
+.. |Docker Images Build Status| image:: https://img.shields.io/github/workflow/status/dask/dask-jobqueue/docker-images-build?logo=docker
```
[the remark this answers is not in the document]
andersy005: I removed the badge

andersy005:
> on a push to the master branch you probably want to generate the Docker files and push them to DockerHub before running the CI. AFAIU having separate ci.yml and build-docker-images.yaml does not make that straightforward so maybe we want a single .yml?

It's my understanding that 

- decoupling the docker image build from the CI has a few advantages including having different event triggers for jobs defined in `ci.yml` and `build-docker-images.yaml`. Combining `ci.yml` and `build-docker-images.yaml` into a single workflow will likely make the process more complicated since the GitHub action event triggers will be the same across the entire workflow. 
- since we will be building these docker images on a daily basis, running the CI with the latest image (even if it's a day or a few hours old) won't cause any problem,

andersy005:
Checking in here. I am curious whether there are any additional issues/tasks this PR needs to address before getting merged?

guillaumeeb:
Sorry @andersy005 for the delay, I did not follow this change, and it's a bit complicated to get into...

I'll wait one more week for @lesteve, and then I'll try to look closer.

At one point, we might want to make you a maintainer of this repo too.

lesteve:
Sorry I didn't have much time to devote to dask-jobqueue unfortunately.

I am going to merge this one and see what happens.

> At one point, we might want to make you a maintainer of this repo too.

100% agreed, I sent you an invitation @andersy005!

lesteve:
AFAICT this seems to be working fine, see https://github.com/dask/dask-jobqueue/actions?query=workflow%3Adocker-images-build and https://hub.docker.com/r/daskdev/dask-jobqueue/tags

lesteve:
Actually I noticed that the images don't get pushed for scheduled events, e.g. https://github.com/dask/dask-jobqueue/runs/1190757827.

I am guessing the condition needs to be changed:
 `if: github.ref == 'refs/heads/master' && github.event_name == 'push'`

PR #466, opened 2020-09-29
Skip building docker images when the image is present

andersy005:
Now that #455 has been merged, this PR is an attempt at making sure to avoid rebuilding images. The image are currently built via the CI, and we just pull the latest images via ` docker-compose pull`. I got carried away and ended up adding a `mamba` bit in this PR :) With mamba, the conda environment creation takes half the time it used to take:

- Without mamba: `~ 2min 8s`

Screen Shot 2020-09-29 at 3 05 09 PM

- With mamba: `~ 1min 13s`

Screen Shot 2020-09-29 at 3 09 02 PM

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     defaults:
       run:
         shell: bash -l {0}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,9 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
     strategy:
       fail-fast: false
       matrix:
@@ -17,32 +20,32 @@ jobs:
       - name: Setup Conda Environment
         uses: goanpeca/setup-miniconda@v1
         with:
-          miniconda-version: "latest"
-          python-version: 3.6
-          environment-file: ci/environment.yml
+          channels: conda-forge
+          mamba-version: "*"
           activate-environment: dask-jobqueue
           auto-activate-base: false
+         
+      - name: Setup conda environment 
+        run: |
+          mamba env update -f ci/environment.yml
+          mamba list
 
       - name: Setup
-        shell: bash -l {0}
         run: |
           source ci/${{ matrix.jobqueue }}.sh
           jobqueue_before_install
 
       - name: Install
-        shell: bash -l {0}
         run: |
           source ci/${{ matrix.jobqueue }}.sh
           jobqueue_install
 
       - name: Test
-        shell: bash -l {0}
         run: |
           source ci/${{ matrix.jobqueue }}.sh
           jobqueue_script
 
       - name: Cleanup
-        shell: bash -l {0}
         run: |
           source ci/${{ matrix.jobqueue }}.sh
           jobqueue_after_script
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,10 @@ jobs:
         jobqueue: ["htcondor", "pbs", "sge", "slurm", "none"]
 
     steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.7.0
+        with:
+            access_token: ${{ github.token }}
       - name: Checkout source
         uses: actions/checkout@v2
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v2
 
-      - name: Setup Conda Environment
+      - name: Setup miniconda
         uses: goanpeca/setup-miniconda@v1
         with:
           channels: conda-forge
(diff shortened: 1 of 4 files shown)


andersy005:
@dask/dask-jobqueue-sge, 

I could use some help to diagnose the most recent SGE CI failures... The CI script keeps getting stuck at `setup` with unending output along these lines:

```bash
aiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
Waiting for SGE slots to become available
```

andersy005:
Never mind... The issue magically disappeared, and the CI is now passing :)

andersy005:
> @dask/dask-jobqueue-sge,
> 
> I could use some help to diagnose the most recent SGE CI failures... The CI script keeps getting stuck at `setup` with unending output along these lines:
> 
> ```shell
> aiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> Waiting for SGE slots to become available
> ```

The failures appear to be intermittent... I will see if I can figure out what's going on...

lesteve:
The intermittent SGE error is weird. I tried reproducing locally and did not manage to. Maybe you can try as well following: https://jobqueue.dask.org/en/latest/develop.html. Otherwise on the CI you can try to look at docker logs and printing the output of `docker exec sge_master qhost` (this tells you whether the SGE cluster is properly setup i.e. the SGE master node see the two SGE worker nodes).

I think the `docker-compose --no-build is fine` since in most cases people should `docker pull` before running `./start-sge.sh`.

I'd be curious to hear about your experience with mamba. It is great to be able to reduce the CI time to be able to iterate faster. At the same time, I am wondering whether this is really worth it. I heard (a while ago so maybe not that true anymore) that mamba solver would give versions potentially very different than the conda one.

andersy005:
> Otherwise on the CI you can try to look at docker logs and printing the output of docker exec sge_master qhost (this tells you whether the SGE cluster is properly setup i.e. the SGE master node see the two SGE worker nodes).

I tried debugging the issue locally, and I got inconsistent results i.e. some attempts resulted into the cluster not being set up properly, some other attempts succeeded. The CI is now failing with a different error :( 

> I'd be curious to hear about your experience with mamba. It is great to be able to reduce the CI time to be able to iterate faster. At the same time, I am wondering whether this is really worth it. I heard (a while ago so maybe not that true anymore) that mamba solver would give versions potentially very different than the conda one.

mamba is relatively stable these days... So far, the solver appears to respect the version requirements whenever they are specified. However, I've noticed that whenever versions aren't specified, conda and mamba solvers may give you different results

lesteve:
The error was weird : my guess is that the SGE cluster was  not properly set up. It is probably not trivial, but it would be really good to get to the bottom of the SGE timeout. If you managed to reproduce it locally, maybe you can look at docker logs, to figure out whether there is some piece of information that could help us troubleshooting further ...

Another thing to try would be to docker-compose restart avec say 15-30 seconds if the cluster was not able to start properly (i.e. still printing `Waiting for SGE slots to become available`).

guillaumeeb:
@andersy005 @lesteve is there something preventing us to merge this one? Except the current conflict?

I understand that SGE CI issue has to be solved somewhere else?

.github/workflows/ci.yaml
```diff
@@ -5,44 +5,52 @@ on: [push, pull_request]
...
-          environment-file: ci/environment.yml
+          channels: conda-forge
+          mamba-version: "*"
           activate-environment: dask-jobqueue
           auto-activate-base: false
+         
+      - name: Setup conda environment 
+        run: |
+          mamba env update -f ci/environment.yml
+          mamba list
```
andersy005, line 36: As I understand it, the current setup doesn't use this new environment. We end up using the environment created during the docker container creation:

https://github.com/dask/dask-jobqueue/blob/127470780ca82cfabecff247ee5e306867606a83/ci/slurm/Dockerfile#L9-L11

I'm wondering.. should we just include  schedulers(pbs, slurm, etc..) _only_ in the docker containers and then create the conda environment (during CI runs) and mount conda and its envs directories into the docker images during test runs? 

It's my understanding that this would:

- make it easy to make changes to the conda environment in a PR without having to mess with the docker container. 
- reduce the docker container size (as a bonus)... 

Any thoughts?

andersy005:
The merge conflict is fixed. 

> I understand that SGE CI issue has to be solved somewhere else?

I think so

.github/workflows/ci.yaml
guillaumeeb: Yep, if you can make that works, I think this would be a nice change to decouple Docker images (base infrastructure) and conda env (real dask-jobqueue use case).

Should we just merge this one and work on this later?

andersy005: Great! 

> Should we just merge this one and work on this later?

👍🏽 Yes, let's merge this as is. I will submit a separate PR to address the docker/conda decoupling.

lesteve:
OK lest's merge this one, thanks @andersy005!

I opened https://github.com/dask/dask-jobqueue/issues/484 about the SGE problem.

PR #468, opened 2020-10-01
Build docker images for scheduled runs

andersy005:
Fixes https://github.com/dask/dask-jobqueue/pull/455#issuecomment-702244825

Cc @lesteve


lesteve:
Thanks a lot!

PR #472, opened 2020-10-27
Add new ConductorCluster

jennalc:
Here is a new cluster type ConductorCluster which enables Dask to run on IBM Spectrum Conductor.


jennalc:
The ConductorCluster uses REST API to communicate with Spectrum Conductor, so it requires the requests module.
When adding ConductorCluster to __init__.py, I saw it broke the htcondor test due to this new requirement, so I have removed it again.

guillaumeeb:
Thanks @jennalc for you contribution here. I've been very busy for several months now, and it seems to be the case of other maintainers here. So please forgive us if we can't take a closer look at it during weeks to come.

lesteve:
Same thing here ... personally my feeling would be to encourage you to publish your code as a separate package and see where that leads you. I would be more than happy to add a list of similar packages if more arise in the dask-jobqueue doc.

As far as I am concerned, the HPC scheduler world is some kind of wild jungle, and from a maintainability perspective supporting a cluster without having access to one is going to be quite a tough sell. Of course, this is my personal opinion, and I am more than willing to hear others opinion on this.

Kudos to having done the work and sorry if that doesn't quite go in the direction you were hoping for. I am hoping this was useful for you in any case since you are now able to use Dask on your cluster.

jennalc:
If it makes any difference, there is a free community version of Spectrum Conductor. I know there is also an LSFCluster type in this repository, which falls into the same category, but not sure the history on that one.

guillaumeeb:
> As far as I am concerned, the HPC scheduler world is some kind of wild jungle, and from a maintainability perspective supporting a cluster without having access to one is going to be quite a tough sell. Of course, this is my personal opinion, and I am more than willing to hear others opinion on this.

I find that dask-jobqueue can accommodate some non well maintained interfaces to job schedulers. Even if the JobQueueCluster implementation does not work on every systems, it gives others insights on how to do it.

If things get broken for a job scheduling system at one point, so be it. We just need to be clear for which systems are really maintained.

jennalc:
@lesteve @guillaumeeb - Guessing this probably fell off the radar at the end of last year.
Just updated with some fixes for compatibility with Dask 2.23.0+ where some of the TCP handshake protocol was changed & the custom handler methods here stopped working.

dask_jobqueue/__init__.py
```diff
@@ -1,15 +1,15 @@
...
-from .htcondor import HTCondorCluster
-
-from ._version import get_versions
-
-__version__ = get_versions()["version"]
-del get_versions
+# flake8: noqa
+from . import config
+from .core import JobQueueCluster
+from .moab import MoabCluster
```
guillaumeeb, line 4: I'm not sure I understand what has changed in this file, and I see no ConductorCluster?

guillaumeeb (commented):
Hi @jennalc,

First of all, really sorry for the time to get back to you. I know this can be frustrating when trying to contribute to something and nothing comes back.

I've browsed the code you propose to add to dask-jobqueue, and the more I read into it, the more I feel the Conductor mechanism is not well suited to this repo. I'm under the impression that you've made this work, but looking at the code, it seems Conductor does not provide any job queueing system like interface (e.g. qsub, qstat, or sbatch or anything).

This implementation seems to rely on some kind of Dask managed service provided by Conductor cluster, that you query by a REST service. Am I understanding correctly? 

It also looks like you are re-implementing the most part if not all the JobQueueCluster base class.

If my analysis is right, I would encourage you to keep this code and make it a separate repository, say dask-conductor for example. You should directly implement SpecCluster, this would probably be simpler.

Let me know if I'm wrong or if I am misunderstanding something.

jennalc:
@guillaumeeb Thank you for taking the time to look at the code. You are correct that the ConductorCluster is interacting with a service provided by Conductor via REST API. That service is acting as a client to request resources on the cluster. So the Dask jobs do effectively queue until they are granted resources on the cluster, but because it uses REST API in an asynchronous way, it is not as straight forward as the other clusters in this repo.

Anyways, understand your reasoning for not wanting to accept the code into this repo. Appreciate your time nonetheless.

guillaumeeb:
> but because it uses REST API in an asynchronous way, it is not as straight forward as the other clusters in this repo.

Yeah this is exeactly it. But please, propose your ConductorCluster as a separate repo. Ideally, you should refactor it to directly extend `SpecCluster` class. We could link to it in our documentation or main Dask doc.

Issue #470, opened 2020-10-16
Make job submission asynchronous


bbockelm:
I have noticed that execution of commands (e.g., `condor_submit` for the condor backend) appear to be synchronous.  In fact, there's a small note about this in the code itself:

https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py#L305

We've started to notice this particularly at very busy batch schedulers.  For example, when dask labextension (https://github.com/dask/dask-labextension) is used in a Jupyter notebook, it will spawn the Dask scheduler inside the jupyter hub process (I think I got this terminology right?) and not the notebook.  Because it's in the hub itself, if dask jobqueue is non-responsive then the entire UI freezes (as no I/O is done in the event loop).  This triggers user complaints of "Jupyter stops working when we use Dask".

The impact of the blocking behavior can be easily seen by replacing the submit executable with a shell script that does a `sleep 20` before invoking the real submit executable.

mivade:
Or to not have to change anything else using the `_call` method with something like:

```python
async def _submit_job(self, script_filename):
    return await asyncio.get_running_loop().run_in_executor(None, self._call(...))
```

oshadura:
@bbockelm I tested both solutions proposed by @mivade in our environment, but I still see some non-responsiveness (I will investigate).

guillaumeeb:
> it will spawn the Dask scheduler inside the jupyter hub process (I think I got this terminology right?) 

For the terminology, it's on the Jupyterlab UI process (the notebook server) that dask-labextension will run, and not in the Kernel (another process were the code gets executed, e.g. the notebook cells).

> Because it's in the hub itself, if dask jobqueue is non-responsive then the entire UI freezes (as no I/O is done in the event loop)

So if I understand, the `condor_submit` command is taking time to run, and so it block all Jupyterlab UI through dask-labextension.

An easy solution is stop using jupyterlab-extension for starting Dask cluster for the time being 😄. I understand this can be seen as a regression for users... For my part I've never use dask-labextension to launch Dask clusters on our job scheduling system, I'm always doing it inside a notebook cell (so the Kernel). I only use the extension to watch my computations.

I also think that this might be a Condor issue (job submission should be almost immediate in job queueing systems), or that this can be handled in dask-labextention maybe?

Anyway, if you find a simple way to make things asynchronous here, this would be welcomed too!

guillaumeeb:
Related to #567

jrueb:
[...]

PR #473, opened 2020-11-13
[WIP] Attempt to make job submission async

oshadura:
Probably fixes: #470 
cc: @bbockelm

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -390,7 +390,10 @@ class Job(ProcessInterface, abc.ABC):
         )
 
         proc = await asyncio.create_subprocess_exec(
-            cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE, **kwargs
+            cmd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            **kwargs
         )
 
         out, err = await proc.communicate()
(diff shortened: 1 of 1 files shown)


dask_jobqueue/core.py
```diff
@@ -302,8 +302,8 @@ def job_file(self):
             yield fn
 
     async def _submit_job(self, script_filename):
-        # Should we make this async friendly?
-        return self._call(shlex.split(self.submit_command) + [script_filename])
+        call = shlex.split(self.submit_command + [script_filename])
+        return await asyncio.get_running_loop().run_in_executor(None, self._call(call))
```
mivade, line 306: If I'm reading this right, I don't think there's a need to run this in an executor as you already made `_call` async.

lesteve:
Some CI error looks genuine for example this one: https://github.com/dask/dask-jobqueue/pull/473/checks?check_run_id=1406501917.

I am not an async expert, but I would guess you need to add some `await` where `_call` is called, maybe changing the functions calling `_call` to `async`. One such function seems to be `_job_id_from_submit_output`, there may be others. In an ideal world, the main user-facing functions would stay synchronous so that if you get an exception it blows up in an obvious way rather than having some log message saying that an exception was never retrieved.

Full disclosure : unfortunately I don't really understand the original issue #470 ...

guillaumeeb:
This PR is a bit old... @oshadura are you still around and up for this change?

I'm going to close this one next time I'll take a look at the activity here, unless someone updates it.

oshadura:
@guillaumeeb sorry I was working on other things and thanks for reminder I will definitely try to take a look in next couple of days and update/close PR.

jacobtomlinson:

PR #476, opened 2020-11-25
Use Status enum

mrocklin:


lesteve:
Great, thanks a lot for this! The SGE timeout is something I haven't had time to look into unfortunately!

PR #501, opened 2021-06-16
WIP: Allow walltime as None to be given during initialization

### LINKED ISSUE #500: Why is the walltime is set by default?
I was wondering why the default walltime for workers is set to 30 minutes instead of None?

For example, see https://github.com/dask/dask-jobqueue/blob/1f9ae1ecc79a5b76930e56c8801f3b8ace659877/dask_jobqueue/jobqueue.yaml#L71

I kept having failed jobs for 2 days until I realized that it's actually this library is setting a hard limit for my jobs, why?
This was very hard to debug among all the other errors that I was trying to fix.

If you want your users to be aware of walltime, isn't there a better solution than setting it to a random number?
Can't you make it a required variable?
Can't the default value be something like `"not-set-by-user"` and complain if it's not set?

183amir:
Distinguish between None (set by user) and the default value
so `walltime=None` can be set during initilization

Fixes #500

### DIFF
--- a/dask_jobqueue/sge.py
+++ b/dask_jobqueue/sge.py
@@ -5,6 +5,7 @@ import dask
 from .core import Job, JobQueueCluster, job_parameters, cluster_parameters
 
 logger = logging.getLogger(__name__)
+DEFAULT = "loaded-from-config"
 
 
 class SGEJob(Job):
@@ -19,7 +20,7 @@ class SGEJob(Job):
         queue=None,
         project=None,
         resource_spec=None,
-        walltime="not-set-by-user",
+        walltime=DEFAULT,
         job_extra=None,
         config_name=None,
         **base_class_kwargs
@@ -36,7 +37,7 @@ class SGEJob(Job):
             resource_spec = dask.config.get(
                 "jobqueue.%s.resource-spec" % self.config_name
             )
-        if walltime == "not-set-by-user":
+        if walltime == DEFAULT:
             walltime = dask.config.get("jobqueue.%s.walltime" % self.config_name)
         if job_extra is None:
             job_extra = dask.config.get("jobqueue.%s.job-extra" % self.config_name)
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
I think my preference here would be to leave the default as `None` but add support for a walltime of `0` to represent unbounded.

mivade:
Wouldn't it be easier to understand if the default was set to the actual default value and `None` meant no wall time? Another advantage to that approach is that it makes documentation of what's going on more straightforward and automatic. I'm also not sure that we'd want to change defaults for just one implementation.

183amir:
> I'm also not sure that we'd want to change defaults for just one implementation.

This is just a draft. I am trying to see how the maintainers would like this to be implemented.

guillaumeeb (commented):
The problem I see with the proposed approach is that if you do this with walltime, you'll have to do it with every parameters, leading to a lot of change. I'm not really fond of this "DEFAULT different of None" approach...

I think we've got to make it clear if kwargs are None, then they are loaded from the config, and if different of None then, they are used in the job_script.

Anyway, I'm fine with setting more parameters in the default config files to None. With a quick glance, I see that almost only the walltime is set by default. We can set it to null and add some advice in the documentation.

andersy005:
> I think we've got to make it clear if kwargs are None, then they are loaded from the config, and if different of None then, they are used in the job_script. 

I agree with @guillaumeeb! I think is reasonable to keep the existing defaults and point users to the `~/.config/dask/jobqueue.yaml` as a recommended way to set their preferred, permanent defaults.

183amir:
> Anyway, I'm fine with setting more parameters in the default config files to None. With a quick glance, I see that almost only the walltime is set by default. We can set it to null and add some advice in the documentation.

I also think this is the best approach. With a bit of documentation and not setting walltime by default, you let users decide what's best for them.

guillaumeeb:
I'm going to close this one. 

I think there is already enough information about how config file works here: https://jobqueue.dask.org/en/latest/configuration.html.

We just need to clarify the default values.

PR #504, opened 2021-07-05
feat: Add RemoteSLURMCluster & RemoteSLURMJob #502

### LINKED ISSUE #502: Add SlurmRESTJob to connect SLURMCluster to a remote Slurm cluster
Hello 👋  Thank you for considering this feature request :) I have been looking over dask-jobqueue (together with prefect) to allocate resources on a Slurm cluster I have access to. dask-jobqueue seems exactly what we'd need for this, thank you for maintaining it 🙇 

### Context

In the case where Slurm and a python process (script or notebook) are not running on the same host SlurmCluster will not be able to spawn any jobs and error.

Slurm added a REST API: https://slurm.schedmd.com/rest_api.html

### Feature

Could we add a `RemoteSlurmCluster` and `RemoteSlurmJob` that largely extend SlurmCluster/SlurmJob and instead of using `subprocess.Popen` we'd do an HTTP request instead?

```
class RemoteSLURMJob(SLURMJob):
    async def _submit_job(self, script_filename):
        # formatting should be according to:
        # https://slurm.schedmd.com/rest_api.html#slurmctldSubmitJob
        return requests.post('slurm-url/jobs/submit')  # reach out to API

    def _job_id_from_submit_output(self, out):
        # out is the JSON output from _submit_job request.post
        # See https://slurm.schedmd.com/rest_api.html#v0.0.36_job_submission_response
        return out['job_id']

    @classmethod
    def _close_job(cls, job_id):
        # See: https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
        requests.delete(f'slurm-url/job/{job_id}')


class RemoteSLURMCluster(SLURMCluster):
    job_cls = RemoteSLURMJob
```

As far as I can tell this should be a drop-in replacement.

Thoughts?

alexandervaneck:
### What does this PR change?

It adds support for RemoteSLURMCluster which spins up dask-workers through the SLURM REST API instead of through subprocess.Popen.

### Why is this useful?

In environments where the python process doesn't have direct access to the SLURM CLI (f.e. a docker container) we still want to be able to use dask-jobqueue to interface with SLURM.

### Open questions

1. I've been able to write some minimal end-to-end test. The implementation largely inherits from SLURMCluster, which is sufficiently tested. Where could we invest more in tests?
2. I've not been able to test this with a workin SLURM REST API cluster, since I am still waiting on my colleagues to activate it. I'm opening this PR early to receive early reviews.

Thank you for allowing me to contribute 🙇‍♂️  @jacobtomlinson @guillaumeeb 

Resolve #502

### DIFF
--- a/ci/slurm/start-slurm.sh
+++ b/ci/slurm/start-slurm.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-docker-compose up -d --no-build
+docker-compose up -d --build
 
 while [ `./register_cluster.sh 2>&1 | grep "sacctmgr: error" | wc -l` -ne 0 ]
   do
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -62,20 +62,20 @@ class RemoteSLURMJob(SLURMJob):
 
     async def _submit_job(self, script):
         # https://slurm.schedmd.com/rest_api.html#slurmctldSubmitJob
-        client_session = aiohttp.ClientSession(
+        async with aiohttp.ClientSession(
             raise_for_status=True, **self.api_client_session_kwargs
-        )
-        # Unfortunately for certain message (like Authentication Failure) SLURM
-        # will respond with a 'Content-Type' of application/json but then the response will
-        # not be parsable JSON.
-        # This now only works if we set `AIOHTTP_NO_EXTENSIONS=1`
-        # See https://github.com/aio-libs/aiohttp/issues/1843
-        response = await client_session.post(
-            f"{self.api_url}slurm/v0.0.36/job/submit",
-            json={"script": script, "job": self._job_configuration},
-        )
-        async with response:
-            return await response.json()
+        ) as client_session:
+            # Unfortunately for certain message (like Authentication Failure) SLURM
+            # will respond with a 'Content-Type' of application/json but then the response will
+            # not be parsable JSON.
+            # This now only works if we set `AIOHTTP_NO_EXTENSIONS=1`
+            # See https://github.com/aio-libs/aiohttp/issues/1843
+            response = await client_session.post(
+                f"{self.api_url}slurm/v0.0.36/job/submit",
+                json={"script": script, "job": self._job_configuration},
+            )
+            async with response:
+                return await response.json()
 
     def _job_id_from_submit_output(self, out):
         # out is the JSON output from _submit_job request.post
@@ -85,14 +85,14 @@ class RemoteSLURMJob(SLURMJob):
     async def close(self):
         logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
         # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
-        client_session = aiohttp.ClientSession(
+        async with aiohttp.ClientSession(
             raise_for_status=True, **self.api_client_session_kwargs
-        )
-        response = await client_session.delete(
-            f"{self.api_url}slurm/v0.0.36/job/{self.job_id}",
-        )
-        async with response:
-            return await response.json()
+        ) as client_session:
+            response = await client_session.delete(
+                f"{self.api_url}slurm/v0.0.36/job/{self.job_id}",
+            )
+            async with response:
+                return await response.json()
 
     @classmethod
     def _close_job(cls, job_id):
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -38,10 +38,10 @@ class RemoteSLURMJob(SLURMJob):
         if self.log_directory is not None:
             configuration[
                 "standard_error"
-            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%%J.err"
+            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%J.err"
             configuration[
                 "standard_output"
-            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%%J.out"
+            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%J.out"
         if self.queue is not None:
             configuration["partition"] = self.queue
         if self.project is not None:
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -169,4 +169,27 @@ class RemoteSLURMCluster(SLURMCluster):
             **kwargs,
         )
 
+    @classmethod
+    def with_http_api_client(
+        cls,
+        http_api_url: str,
+        http_api_user_name: str,
+        http_api_user_token: str,
+        *args,
+        **kwargs,
+    ) -> "RemoteSLURMCluster":
+        kwargs["api_url"] = (
+            http_api_url if http_api_url.endswith("/") else f"{http_api_url}/"
+        )
+        return cls(
+            api_client_session_kwargs=dict(
+                headers={
+                    "X-SLURM-USER-NAME": http_api_user_name,
+                    "X-SLURM-USER-TOKEN": http_api_user_token,
+                }
+            ),
+            *args,
+            **kwargs,
+        )
+
     job_cls = RemoteSLURMJob
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -1,7 +1,6 @@
 import logging
 from contextlib import contextmanager
-from typing import Any, Dict, Generator
-from urllib.parse import urljoin
+from typing import Any, Dict
 
 import aiohttp
 
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -68,7 +68,6 @@ class RemoteSLURMJob(SLURMJob):
         response = await client_session.post(
             f"{self.api_url}slurm/v0.0.36/job/submit",
             json={"script": script, "job": self._job_configuration},
-            raise_for_status=False,
         )
         async with response:
             return await response.json()
@@ -86,7 +85,6 @@ class RemoteSLURMJob(SLURMJob):
         )
         response = await client_session.delete(
             f"{self.api_url}slurm/v0.0.36/job/{self.job_id}",
-            raise_for_status=False,
         )
         async with response:
             return await response.json()
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -1,6 +1,6 @@
 import logging
 from contextlib import contextmanager
-from typing import Dict, Generator
+from typing import Any, Dict, Generator
 from urllib.parse import urljoin
 
 import aiohttp
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 
 class RemoteSLURMJob(SLURMJob):
     def __init__(self, *args, **kwargs):
-        self.api_client_session = kwargs.pop("api_client_session")
+        self.api_client_session_kwargs = kwargs.pop("api_client_session_kwargs")
         self.api_url = kwargs.pop("api_url")
         self.remote_job_extra = (
             kwargs.pop("remote_job_extra") if "remote_job_extra" in kwargs else {}
@@ -65,16 +65,15 @@ class RemoteSLURMJob(SLURMJob):
 
     async def _submit_job(self, script):
         # https://slurm.schedmd.com/rest_api.html#slurmctldSubmitJob
-        try:
-            response = await self.api_client_session.post(
-                f"{self.api_url}slurm/v0.0.36/job/submit",
-                json={"script": script, "job": self._job_configuration},
-            )
-            response.raise_for_status()
-        except aiohttp.ClientError as e:
-            logger.exception("SLURMJob request failed.")
-            raise RuntimeError from e
-        else:
+        client_session = aiohttp.ClientSession(
+            raise_for_status=True, **self.api_client_session_kwargs
+        )
+        response = await client_session.post(
+            f"{self.api_url}slurm/v0.0.36/job/submit",
+            json={"script": script, "job": self._job_configuration},
+            raise_for_status=False,
+        )
+        async with response:
             return await response.json()
 
     def _job_id_from_submit_output(self, out):
@@ -85,15 +84,14 @@ class RemoteSLURMJob(SLURMJob):
     async def close(self):
         logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
         # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
-        try:
-            response = await self.api_client_session.delete(
-                f"{self.api_url}slurm/v0.0.36/job/{self.job_id}"
-            )
-            response.raise_for_status()
-        except aiohttp.ClientError as e:
-            logger.exception("SLURMJob request failed.")
-            raise RuntimeError from e
-        else:
+        client_session = aiohttp.ClientSession(
+            raise_for_status=True, **self.api_client_session_kwargs
+        )
+        response = await client_session.delete(
+            f"{self.api_url}slurm/v0.0.36/job/{self.job_id}",
+            raise_for_status=False,
+        )
+        async with response:
             return await response.json()
 
     @classmethod
@@ -150,12 +148,12 @@ class RemoteSLURMCluster(SLURMCluster):
 
     def __init__(
         self,
-        api_client_session: aiohttp.ClientSession,
+        api_client_session_kwargs: Dict[str, Any],
         *args,
         **kwargs,
     ):
         # Set into kwargs, they'll be passed through `**job_kwargs` into RemoteSLURMJob
-        kwargs["api_client_session"] = api_client_session
+        kwargs["api_client_session_kwargs"] = api_client_session_kwargs
         super().__init__(*args, **kwargs)
 
     @classmethod
@@ -164,7 +162,7 @@ class RemoteSLURMCluster(SLURMCluster):
     ) -> "RemoteSLURMCluster":
         kwargs["api_url"] = "http://localhost/"
         return cls(
-            api_client_session=aiohttp.ClientSession(
+            api_client_session_kwargs=dict(
                 connector=aiohttp.UnixConnector(path=socket_api_path)
             ),
             *args,
--- a/dask_jobqueue/remote_slurm.py
+++ b/dask_jobqueue/remote_slurm.py
@@ -1,19 +1,22 @@
 import logging
 from contextlib import contextmanager
+from typing import Dict, Generator
 from urllib.parse import urljoin
 
 import aiohttp
 
 from .core import cluster_parameters, job_parameters
-from .slurm import SLURMCluster, SLURMJob
+from .slurm import SLURMCluster, SLURMJob, slurm_format_bytes_ceil
 
 logger = logging.getLogger(__name__)
 
 
 class RemoteSLURMJob(SLURMJob):
     def __init__(self, *args, **kwargs):
-        self.api_url = kwargs.pop("api_url")
-        self.__class__.api_url = self.api_url
+        self.api_socket_path = kwargs.pop("api_socket_path")
+        self.remote_job_extra = (
+            kwargs.pop("remote_job_extra") if "remote_job_extra" in kwargs else {}
+        )
         super().__init__(*args, **kwargs)
 
     @contextmanager
@@ -24,19 +27,57 @@ class RemoteSLURMJob(SLURMJob):
         """
         yield self.job_script()
 
+    @property
+    def _job_configuration(self) -> Dict:
+        configuration = {
+            "environment": {"name": "test"},
+            "tasks": 1,
+            "cpus_per_task": self.job_cpu or self.worker_cores,
+        }
+
+        if self.job_name is not None:
+            configuration["name"] = self.job_name
+        if self.log_directory is not None:
+            configuration[
+                "standard_error"
+            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%%J.err"
+            configuration[
+                "standard_output"
+            ] = f"{self.log_directory}/{self.job_name or 'worker'}-%%J.out"
+        if self.queue is not None:
+            configuration["partition"] = self.queue
+        if self.project is not None:
+            configuration["account"] = self.project
+
+        self.memory = self.job_mem
+        if self.job_mem is None:
+            self.memory = slurm_format_bytes_ceil(self.worker_memory)
+        if self.memory is not None:
+            configuration["memory_per_node"] = self.memory
+
+        if self.walltime is not None:
+            configuration["time_limit"] = self.walltime
+
+        configuration.update(self.remote_job_extra)
+
+        return configuration
+
     async def _submit_job(self, script):
         # https://slurm.schedmd.com/rest_api.html#slurmctldSubmitJob
         try:
-            async with aiohttp.ClientSession() as session:
+            async with aiohttp.ClientSession(
+                connector=aiohttp.UnixConnector(path=self.api_socket_path)
+            ) as session:
                 response = await session.post(
-                    urljoin(self.api_url, "jobs/submit"), json={"script": script}
+                    "http://localhost/slurm/v0.0.36/job/submit",
+                    json={"script": script, "job": self._job_configuration},
                 )
                 response.raise_for_status()
         except aiohttp.ClientError as e:
             logger.exception("SLURMJob request failed.")
             raise RuntimeError from e
         else:
-            return response.json()
+            return await response.json()
 
     def _job_id_from_submit_output(self, out):
         # out is the JSON output from _submit_job request.post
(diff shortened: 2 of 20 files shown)


dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,114 @@
...
+        Amount of memory to request in SLURM. If None, defaults to worker
+        processes * memory
+    job_extra : list
+        List of other Slurm options, for example -j oe. Each option will be prepended with the #SBATCH prefix.
+
+    Examples
+    --------
+    >>> from dask_jobqueue import SLURMCluster
+    >>> cluster = SLURMCluster(
+    ...     queue='regular',
```
andersy005, line 88:
```suggestion
    >>> cluster = RemoteSLURMCluster(
    ...     api_url='http://a-url.com/slurm/',
    ...     queue='regular',
```

andersy005 (commented):
Thank you for putting this together, @AlexanderVanEck

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,114 @@
...
+    """.format(
+        job=job_parameters, cluster=cluster_parameters
+    )
+
+    def __init__(self, *args, **kwargs):
+        if "api_url" not in kwargs:
+            raise ValueError("api_url is missing from RemoteSlurmCluster.")
+
+        if not kwargs["api_url"].endswith("/"):
+            raise ValueError("api_url on RemoteSlurmCluster should end with /.")
```
andersy005, line 110: The lack of `/` is likely going to be common among users. I was wondering if we could normalize the URL for the user instead of raising an error?

AlexanderVanEck: You're right. I thought about using https://github.com/gruns/furl for this, quite fond of it. but it seemed overkill to add a library for 1 url.

We could also do;

```
if not kwargs["api_url"].endswith("/"):
    kwargs["api_url"] == f"{kwargs['api_url']}/"
```

?

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Added

dask_jobqueue/remote_slurm.py
andersy005: 👍🏽 for the simple approach without a new dependency

AlexanderVanEck: Added

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,115 @@
...
+         gives output directly to SLURMJob._submit_job().
+        """
+        yield self.job_script()
+
+    async def _submit_job(self, script):
+        # https://slurm.schedmd.com/rest_api.html#slurmctldSubmitJob
+        try:
+            response = requests.post(
+                f"{self.api_url}jobs/submit", json={"script": script}
+            )
```
jacobtomlinson, line 29: Given that this call is being made within an `async` method it will block the event loop.

It would be better to use `aiohttp` here.

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,115 @@
...
+    def _job_id_from_submit_output(self, out):
+        # out is the JSON output from _submit_job request.post
+        # See https://slurm.schedmd.com/rest_api.html#v0.0.36_job_submission_response
+        return out["job_id"]
+
+    async def close(self):
+        logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
+        # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
+        try:
+            response = requests.delete(f"{self.api_url}job/{self.job_id}")
```
jacobtomlinson, line 46: Same here. This should be an async HTTP call with `aiohttp`.

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,115 @@
...
+    """.format(
+        job=job_parameters, cluster=cluster_parameters
+    )
+
+    def __init__(self, *args, **kwargs):
+        if "api_url" not in kwargs:
+            raise ValueError("api_url is missing from RemoteSlurmCluster.")
+
+        if not kwargs["api_url"].endswith("/"):
+            kwargs["api_url"] = f"{kwargs['api_url']}/"
```
jacobtomlinson, line 111: Instead of trying to handle trailing slashes and things here you may find it better to use `urllib.parse.urljoin` to join parts of a URL together.

dask_jobqueue/tests/test_remote_slurm.py
```diff
@@ -0,0 +1,48 @@
+import sys
+
+import pytest
+
+from dask_jobqueue.remote_slurm import RemoteSLURMCluster
+
+TEST_API_URL = "http://a-url.com/slurm"
+
+
+def test_end_to_end(requests_mock):
```
jacobtomlinson, line 10: It would be good to have an async test here too to ensure the `asyncio` functionality also works (required for the Dask Jupyter Lab Extension to function).

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,115 @@
...
+        return out["job_id"]
+
+    async def close(self):
+        logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
+        # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
+        try:
+            response = requests.delete(f"{self.api_url}job/{self.job_id}")
+            response.raise_for_status()
+        except requests.HTTPError as e:
+            logger.exception("SLURMJob request failed.")
```
jacobtomlinson, line 49: The parent `SLURMCluster` (specifically the root object `Cluster`) has an `_log` method which would be preferable here as we store and reuse that info in other places. 

I wonder if we could pass a pointer to the cluster object to the `RemoteSLURMJob` objects so it can be used.

jacobtomlinson (commented):
Thanks so much for raising this. I had a quick pass of the code but do not have a SLURM setup to test against.

dask_jobqueue/remote_slurm.py
AlexanderVanEck: You're right that's probably an improvement. However since Job currently logs to it's own logger I'm inclined to keep that behavior here and we should figure out logging in a separate PR. (Trying to keep the scope tight).

dask_jobqueue/tests/test_remote_slurm.py
AlexanderVanEck: Just found a `test_job` function to which I can add RemoteSLURMCluster and RemoteSLURMJob. Then async should be fully tested. :v:

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Added

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Added

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Thank you for the recommendation. :) I've implemented aiohttp now, however the test is failing with;

```
E               RuntimeError: Task <Task pending coro=<SpecCluster._correct_state_internal() running at distributed/deploy/spec.py:326> cb=[<TaskWakeupMethWrapper object at 0x1126262e8>()]> got Future <Future pending> attached to a different loop
```

I suppose this is because the loop SpecCluster does not get passed to the job and thus the job cannot add tasks to the loop. Jobs also do not get passed the loop currently.

I have very limited knowledge of asyncio, rarely work with it... Would you be able to assist here @jacobtomlinson?

dask_jobqueue/tests/test_remote_slurm.py
```diff
@@ -0,0 +1,56 @@
...
+
+    async with RemoteSLURMCluster(
+        api_url=TEST_API_URL,
+        walltime="00:02:00",
+        processes=4,
+        cores=8,
+        memory="28GB",
+        n_workers=1,
+        job_extra=["--gpus 2"],
+        loop=loop,
```
jacobtomlinson, line 30: You probably need the `asynchronous=True` kwarg here also.

dask_jobqueue/remote_slurm.py
jacobtomlinson: They shouldn't need to be passed the loop. Just implement async methods.

AlexanderVanEck:
TODO: Add SLURM REST API to ci/slurm and remove mocking from test_end_to_end. We can use the slurm cluster instead.

(cc @jacobtomlinson I think this is a better idea)

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Got it to work :v:

dask_jobqueue/tests/test_remote_slurm.py
AlexanderVanEck: Thanks, missed that one.

jacobtomlinson:
Given that you've already written the mocked test there's probably little harm in leaving it in, but adding a CI test would be good.

(Also don't worry about squashing and force pushing, we will squash on merge. It actually makes it harder to review because we can't easily see what has changed between reviews)

dask_jobqueue/tests/test_remote_slurm.py
```diff
@@ -0,0 +1,57 @@
...
+            "#SBATCH --cpus-per-task=8\n"
+            "#SBATCH --mem=27G\n"
+            "#SBATCH -t 00:02:00\n"
+            "#SBATCH --gpus 2\n"
+            "\n"
+            f"{sys.executable} -m "
+            f"distributed.cli.dask_worker {cluster.scheduler.address} --nthreads "
+            "2 --nprocs 4 --memory-limit 7.00GB --name RemoteSLURMCluster-0 "
+            "--nanny --death-timeout 60 --protocol tcp://\n",
+        }
```
jacobtomlinson, line 48: I assume this could subtly change over time and the test would need to be updated each time.

It may be better to replace this with a series of assertions testing some of the content.

```suggestion
        request = aresponses.history[0].request.json()
        assert  "#SBATCH -J dask-worker" in request
        assert  "#SBATCH --gpus 2" in request
        assert  sys.executablein request
        assert  "--name RemoteSLURMCluster-0" in request
        # etc, etc
```

AlexanderVanEck: Good point. I'll probably completely remove this test as the `aresponses` dependency cannot be installed by conda and therefore it will never execute on CI.

AlexanderVanEck:
> Given that you've already written the mocked test there's probably little harm in leaving it in, but adding a CI test would be good.
> 
> (Also don't worry about squashing and force pushing, we will squash on merge. It actually makes it harder to review because we can't easily see what has changed between reviews)

I unfortunately have to remove this test because the `aresponses` dependency cannot be installed by conda. In any case I'm currently working on making the tests work with the ci/slurm docker environment. That should provide plenty of test coverage for us to accept this PR in good conscience.

AlexanderVanEck:
@lesteve Would you have any tips on how I can update the ci/slurm integration test docker setup and have it be used during integration testing of slurm?

For this PR to pass we would need;

1. Update slurm version to 20.11 (the version the REST API was released in)
2. Add `slurmrestd` somehow to have an endpoint we can send requests to in the tests.

> Update slurm
I pushed an image to dockerhub and added it to the FROM section in the ci/slurm/Dockerfile

> slurmrestd
Perhaps you know, off the top of your head, where I should look for configuration examples to configure this? any direction you'd prefer would be appreciated.

I'll make time to look at this today 👍

docs/source/examples.rst
```diff
@@ -180,3 +180,22 @@ specified for certain steps in the processing. For example:
     result_stage_2 = client.compute(stage_2,
                                     resources={tuple(stage_1): {'GPU': 1},
                                                tuple(stage_2): {'ssdGB': 100}})
+Remote SLURM Deployment
+-----------------
+
+All the above SLURMCluster examples are also possible over the SLURM REST API with the use of
+RemoteSlurmCluster. See https://slurm.schedmd.com/rest_api.html for more information.
```
ocaisa, line 187: This is only likely to work under pretty restricted circumstances (like you mentioned in https://github.com/dask/dask-jobqueue/issues/502#issuecomment-874697650) so it is probably good to document that. Something like
```
Remote SLURM deployment is only going to work when the scheduler has access to the SLURM REST API _and_ the workers that are launched on the (remote) cluster are able to communicate with the scheduler over HTTP (e.g., the scheduler is inside a container that is running on the cluster where the jobs will be launched).
```

AlexanderVanEck: Added

jacobtomlinson:
Please stop force pushing 😂. I can't see what changes you are making.

image

AlexanderVanEck:
> Please stop force pushing 😂. I can't see what changes you are making.
> 
> image

I'll try my absolute best. 😬 it's part of my normal workflow to rebase/retouch commits to make them look nice and to the point.

guillaumeeb:
> Update slurm version to 20.11 (the version the REST API was released in)

As you might see, we were using images from https://github.com/giovtorres/slurm-docker-cluster, but it has not been updated on dockerhub for two years. The simplest thing to do here would probably to copy paste the Dockerfile from this project, update Slurm version, and to build and push the image through our CI to daskdev docker user. It could then be used for running Python tests.

> Add slurmrestd somehow to have an endpoint we can send requests to in the tests.

Never configured a Slurm cluster, and I'm not sure any maintainer here has... So you'll have to search or try, since https://slurm.schedmd.com/rest.html does not seem to give all needed information... In the end this is probably some lines of configuration, and one more daemon to launch in the docker-compose setup.

> Alternatively I could update this PR with unit tests only (with mocking) and we tackle those integration tests in another PR (due to ci.yml action needing to build the newer slurm docker image later).

This is an alternative, as updating CI here might be a bit complicated, and we've already done it. But It is a really important thing to do though.

AlexanderVanEck:
@jacobtomlinson I got it to work with Unix sockets as authentication. Slurm does not accept unauthenticated requests. Let me know what you think of this implementation.

Something inside Slurm is making it impossible to share the unix socket via the docker-compose file (it errors as permission error), therefore we have to run those tests in the slurmrestd container directly. 

JWT authentication does not seem to work yet, tomorrow I'll investigate if the latest version 20-11-8-1 can make it work. 20-11-4-1 (current version in Dockerfile) seems to be broken.

@guillaumeeb thank you for your insights.

> The simplest thing to do here would probably to copy paste the Dockerfile from this project, update Slurm version, and to build and push the image through our CI to daskdev docker user. It could then be used for running Python tests.

I had the same thought and made a new slurm docker image definition for us. :)

> In the end this is probably some lines of configuration, and one more daemon to launch in the docker-compose setup.

You were right 👍 however I must say that the slurm docs leave a lot to be desired. I ended up reading their code to make sense of the configuration/requests.

docs/source/examples.rst
```diff
@@ -180,3 +180,27 @@ specified for certain steps in the processing. For example:
...
+This only works for environments where the scheduler is able to communicate with the SLURM REST API
+and the workers get scheduled on SLURM nodes that also can communicate with the scheduler.
+(f.e. the scheduler is inside a docker container but still within the HPC cluster network)
+
+.. code-block:: python
+
+   from dask_jobqueue import RemoteSLURMCluster
+
+   cluster = RemoteSLURMCluster(
+                          api_url='http://<your-slurm-http-rest-api-url>/',
```
AlexanderVanEck, line 198: TODO

- [ ] Update the documentation for RemoteSLURMCluster.from_socket...

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,174 @@
...
+    @contextmanager
+    def job_file(self):
+        """
+        Override the job_file and pass back the job_script. job_file() is used by Job.start and
+         gives output directly to SLURMJob._submit_job().
+        """
+        yield self.job_script()
+
+    @property
+    def _job_configuration(self) -> Dict:
```
AlexanderVanEck, line 32: Seeing as this function is the exact same form as in SlurmJob it's probably worth it to create a dataclass which can then be transformed to ways (as dict and as sbatch script).

@jacobtomlinson what do you think?

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,174 @@
...
+    def _job_id_from_submit_output(self, out):
+        # out is the JSON output from _submit_job request.post
+        # See https://slurm.schedmd.com/rest_api.html#v0.0.36_job_submission_response
+        return out["job_id"]
+
+    async def close(self):
+        logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
+        # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
+        try:
+            response = await self.api_client_session.delete(
```
AlexanderVanEck, line 89: @jacobtomlinson I'm not super familiar with aiohttp. Now when I run pytest it passes succesfully (after the last test) but then prints;

```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fc5a2477e10>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7fc5a20f25f8>, 1007195.255574382)]']
connector: <aiohttp.connector.UnixConnector object at 0x7fc5a2477898>
Task was destroyed but it is pending!
task: <Task pending coro=<AdaptiveCore.adapt() done, defined at /opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive_core.py:179> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fc5a24e0c48>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py:688]>
```

Any ideas how to terminate the session properly when it's being passed in like this?

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,174 @@
...
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def with_socket_api_client(
+        cls, socket_api_path: str, *args, **kwargs
+    ) -> "RemoteSLURMCluster":
+        kwargs["api_url"] = "http://localhost/"
+        return cls(
+            api_client_session=aiohttp.ClientSession(
+                connector=aiohttp.UnixConnector(path=socket_api_path)
```
jacobtomlinson, line 168: This session never gets closed. This is the warning you are seeing.

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Figured it out :)

dask_jobqueue/remote_slurm.py
AlexanderVanEck: Passing the kwargs instead of the ClientSession did the trick. Also reformatted the calls to be a bit less verbose with the try...except. :)

AlexanderVanEck:
I've managed to also add the JWT HTTP configuration - however the integration tests fail. Given that the posted payload is correct it leads me to believe that slurmrestd doesn't actually have JWT support yet.

The NEWS in slurm mentions a couple of bugfixes in version later than 20-11-4-1 (the version currently installable in ubuntu) https://github.com/SchedMD/slurm/blob/master/NEWS. So I'll have to investigate if something is wrong with the token generation/authentication or this is a bug in Slurm.

AlexanderVanEck:
> I've managed to also add the JWT HTTP configuration - however the integration tests fail. Given that the posted payload is correct it leads me to believe that slurmrestd doesn't actually have JWT support yet.
> 
> The NEWS in slurm mentions a couple of bugfixes in version later than 20-11-4-1 (the version currently installable in ubuntu) https://github.com/SchedMD/slurm/blob/master/NEWS. So I'll have to investigate if something is wrong with the token generation/authentication or this is a bug in Slurm.

I figured out JWT authentication, now still need to figure out why we're seeing;

```
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f39d33c7048>, 1059921.899742882)]']
connector: <aiohttp.connector.UnixConnector object at 0x7f3a0578d7f0>
Task was destroyed but it is pending!
task: <Task pending coro=<AdaptiveCore.adapt() done, defined at /opt/anaconda/lib/python3.6/site-packages/distributed/deploy/adaptive_core.py:179> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f39d34bc5e8>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /opt/anaconda/lib/python3.6/site-packages/tornado/ioloop.py:688]>
```

🤔

AlexanderVanEck:
@jacobtomlinson @lesteve this is now ready for a good review, all the features explained in the issue are in. Please let me know your thoughts 🙇 your thoughts help immensely.

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,195 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RemoteSLURMJob(SLURMJob):
+    def __init__(self, *args, **kwargs):
+        self.api_client_session_kwargs = kwargs.pop("api_client_session_kwargs")
+        self.api_url = kwargs.pop("api_url")
+        self.remote_job_extra = (
+            kwargs.pop("remote_job_extra") if "remote_job_extra" in kwargs else {}
+        )
```
andersy005, line 20:
```suggestion
        self.remote_job_extra = kwargs.pop("remote_job_extra", {})
        )
```

AlexanderVanEck: Added

dask_jobqueue/remote_slurm.py
```diff
@@ -0,0 +1,195 @@
...
+        return out["job_id"]
+
+    async def close(self):
+        logger.debug("Stopping worker: %s job: %s", self.name, self.job_id)
+        # https://slurm.schedmd.com/rest_api.html#slurmctldCancelJob
+        client_session = aiohttp.ClientSession(
+            raise_for_status=True, **self.api_client_session_kwargs
+        )
+        response = await client_session.delete(
+            f"{self.api_url}slurm/v0.0.36/job/{self.job_id}",
```
andersy005, line 88: Aren't there cases in which someone could be running a different version other than `v0.0.36`? So, should we be hard coding this here?

AlexanderVanEck: Theoretically yes, if we want to open it up like that then we should probably have users include slurm/v0.0.36 into the api_url itself.

In reality slurmrestd's implementations are nearly identical between 0.0.35/36/37

andersy005: Sounds good... Thank you for the clarification

ci/slurm/start-slurm.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-docker-compose up -d --no-build
```
andersy005, line 3: @AlexanderVanEck, this option was introduced as a way to avoid building images for each push event (it was making the CI take too long to run). Since you want to be able to use the up-to-date docker images in this PR, I recommend separating this PR into two PRs (1) for the existing `RemoteSLURMCluster` and (2) for updating the docker images. PR(2) would get merged before merging this PR. What do you think?

AlexanderVanEck: I understand where you are coming from - however I think we can get away with it.

1. In the scenario where there's is no change in the Dockerfile
- We do docker-compose pull first, which pulls the latest from dockerhub
- docker-compose up --build recognises that there's no changes and uses cache for all layers

2. In the scenario where there's is a change in the Dockerfile
- We do docker-compose pull first, which pulls the latest from dockerhub
- docker-compose up --build recognises that there's some changes and uses cache for all layers that are unaffected

Drawbacks:
1. It will try to build every service in docker-compose, but it will use cache for all but the first one. f.e. https://github.com/dask/dask-jobqueue/pull/504/checks?check_run_id=3069548195#step:6:4178

Is this not what is expected?

> PR into two PRs

The "problem" here would be that the (2) PR does not test the Dockerfile before it is merged because the CI uses the old 'latest' one. and thus it is very likely that I'll push a dockerfile that won't work. 

F.e. see the failed tests in this PR which for some reason made the cluster not come up. Locally this worked.

AlexanderVanEck:
I found a bit of a snag working with aiohttp. Due to SLURM responses sometimes (most of the time?) not being parsable the http-parser library fails on reading the response. Unfortunately I have not been able to deduce why these responses cannot be read.

In any case setting AIOHTTP_NO_EXTENSIONS as suggested by https://github.com/aio-libs/aiohttp/issues/1843 fixes this. I have however no idea how we should control for this in _submit_job. 🤔

guillaumeeb:
@AlexanderVanEck I see you push some commits about a month ago. Do you still feel this PR is ready to review?

ci/slurm/start-slurm.sh
guillaumeeb: @andersy005 I've not yet fully understood the new CI process, do you agree with @AlexanderVanEck answer?

dask_jobqueue/remote_slurm.py
guillaumeeb: Is this really resolved? I'd say we need a way to configure the version in the URL for future compatibility?

dask_jobqueue/tests/test_remote_slurm.py
```diff
@@ -0,0 +1,122 @@
...
+    return RemoteSLURMCluster.with_http_api_client(
+        http_api_url=TEST_API_URL,
+        http_api_user_name=api_username,
+        http_api_user_token=api_token,
+        cores=2,
+        processes=1,
+        memory="2GB",
+        job_mem="2GB",
+        loop=loop,
+        log_directory="/var/log/slurm",
```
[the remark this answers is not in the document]
guillaumeeb: @AlexanderVanEck, is this true for other tests without remote APIs?

sanjeev-one:
+1

jacobtomlinson:
This PR hasn't has any activity for a long time so I'm going to close it out.

PR #507, opened 2021-07-13
JobQueueCluster: add option to change used scheduler class

bfis:
The current `JobQueueCluster` implementation does not allow to change the (effective) value of `scheduler_spec["class"]`, although the underlying `SpecCluster` does support it.

The proposed change adds the keyword argument `scheduler_cls` to the `JobQueueCluster.__init__`, with it's default set to the currently used `Scheduler` (from `distributed.scheduler`) thus retaining the original behavior.
However, it makes the very same keyword argument (`scheduler_cls`) unavailable to the `**job_kwargs`. Currently, this does not interfere with any implementation within this repository and is reasonably expected to not change in the future. If, this is still an undesired effect, one could also retrieve the value from a specific entry (e.g. "cls") in the `scheduler_options` (i.e. via `scheduler_options.pop("cls", Scheduler)`).

In our particular use case, this would enable us to properly change parts of the schedulers behavior (in particular through `worker_objective` and `check_idle_saturated`) through subclassing instead of method patching which is favorable, especially in the light of the new Cython accelerated `SchedulerState`.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -81,6 +81,9 @@ cluster_parameters = """
         port the web dashboard should use or ``scheduler_options={'host': 'your-host'}``
         to specify the host the Dask scheduler should run on. See
         :class:`distributed.Scheduler` for more details.
+    scheduler_cls : type
+        Changes the class of the used Dask Scheduler. Defaults to  Dask's
+        :class:`distributed.Scheduler`.
 """.strip()
 
 
@@ -444,6 +447,7 @@ class JobQueueCluster(SpecCluster):
         dashboard_address=None,
         host=None,
         scheduler_options=None,
+        scheduler_cls=Scheduler,  # Use local scheduler for now
         # Options for both scheduler and workers
         interface=None,
         protocol="tcp://",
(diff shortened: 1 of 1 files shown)


guillaumeeb (approved):
Thanks @bfis for this PR!

Sorry for the delay here (as always...), This looks perfectly really fine to me.

Just one question on:
> However, it makes the very same keyword argument (scheduler_cls) unavailable to the **job_kwargs. Currently, this does not interfere with any implementation within this repository and is reasonably expected to not change in the future. If, this is still an undesired effect, one could also retrieve the value from a specific entry (e.g. "cls") in the scheduler_options (i.e. via scheduler_options.pop("cls", Scheduler)).

I don't see any reason we would need this keyword in the worker options either, it seems you don't too. If we ever need it, could we also add an entry somewhere there: https://github.com/dask/dask-jobqueue/blob/main/dask_jobqueue/core.py#L520?
```python
job_kwargs["scheduler_cls"] = scheduler_cls
```

bfis:
> [...]
> I don't see any reason we would need this keyword in the worker options either, it seems you don't too. If we ever need it, could we also add an entry somewhere there: https://github.com/dask/dask-jobqueue/blob/main/dask_jobqueue/core.py#L520?
> 
> ```python
> job_kwargs["scheduler_cls"] = scheduler_cls
> ```

Yes, we could make it accessible there if need be, however it would probably be easier to just use the equivalent `type(self.scheduler)`.

PR #514, opened 2021-09-04
New attempt at #411 (adding extra argument to condor_submit)

riedel:
I think I might have found the problems in #434


riedel:
@matyasselmeci : do you want to incorporate the changes into #434 ? Seems to work now. However as mentioned in #434 I did not manage to keep all your tests. However, IMHO they are still better than before. Maybe one could do something less invasive than letting condor_submit actually fail.

matyasselmeci:
I'm OK with dropping `test_extra_args_broken_submit`.

I just thought of another way to test submit_command_extra: `condor_submit` lets you pass arguments like `key=value` that will end up as values in the submit file. If you do `MY.key="value"`, then `key` will end up as a classad attribute in the resulting job.

So the test would be to call `HTCondorCluster` with `submit_command_extra=['MY.attr1="hello"', 'MY.attr2="world"']` and later run `["condor_q", "-af", "attr1", "attr2"]` and look for `hello world` in the output.

guillaumeeb (approved):
Not a strong opinion from me on this one.

I'd say it doesn't modify the core package a lot, so if it's useful for Condor users we can probably merge.

If @lesteve or @andersy005 have an opinion, I'd be glad to hear it.

andersy005 (approved):
I am not familiar with HTCondor details well enough to give this a very thorough review, but I concur with Guillaume's statement:

> I'd say it doesn't modify the core package a lot, so if it's useful for Condor users we can probably merge.

riedel:
Thanks for reviewing. IMHO (biased), the "critical" part is not the HTCondor but the small change in passing the cancel_command to the class method. But I think it makes sense beyond HTCondor, because this would be the expected behaviour, when modifying it (i.e. that the modification  acts symetric to the submit_command which is is passed to an instance method).  I first thought of making this more extensible, but I kept this minimal (which was acknowledged :) )

I would opt for postponing adding a better additional test case, so I would be greatful if someone could merge this one to get #411 off the table for HTCondor before cleaning up the rest a bit more.

riedel:
I closed #434 now, but want to mention again that the majority of work was done there by @matyasselmeci . I only fixed the last bit that was blocking the merge.

matyasselmeci (approved):
This looks fine to me. Thanks for figuring out the issue!

Issue #484, opened 2021-01-25
Fix SGE timeout in CI


lesteve:
See https://github.com/dask/dask-jobqueue/pull/466#issuecomment-707606439 (and comments around) for more details about how to debug and possible work-arounds.

I was able to reproduce locally but this is intermittent ...

Another possible thing to do is to figure out what has changed since SGE was running fine (maybe 3 months ago or so) but this could actually take more time than finding a work-around.

Yet another thing to do would be to look at the repo we are taking the Docker image from and see whether they encountered similar issues ...

PR #515, opened 2021-09-07
Testing CI : Fix the intermittent failing build with SGE

guillaumeeb:
It seems I have broken something in the last merge request, just checking (and hopefully fixing then).

Edit:
Fixes https://github.com/dask/dask-jobqueue/issues/484.

### DIFF
--- a/ci/sge.sh
+++ b/ci/sge.sh
@@ -20,12 +20,12 @@ function jobqueue_install {
 }
 
 function jobqueue_script {
-    #real test
     docker exec sge_master /bin/bash -c "cd /dask-jobqueue; pytest dask_jobqueue --verbose -s -E sge"
 }
 
 function jobqueue_after_script {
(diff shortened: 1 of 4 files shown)


guillaumeeb:
OK, not sure why I got CI builds failure error on the last merge. This seems as before, only the SGE build is failing. I'm looking a bit into that.

guillaumeeb:
So I found some race condition when starting SGE master, and also when restarting the slaves (because it was done two times). I put some more checks to avoid these problems, and just start the slaves once. I also modified the docker-compose setup to use the script from git repo rather than those embedded into the Docker image (which may be out of date).

Hopefully, this will fix the intermittent SGE build failure.

I reproduced it on a VM, but as people may notice not in this PR! In my test VM, with the fix I made, I didn't reproduce any problems in about 20 tries (before that, I observed a problem about one time out of three).

PR #519, opened 2021-09-11
make tls default if security is set...

### LINKED ISSUE #496: protocol should default to tls if security is set
if there is nothing that speaks against that I would file a PR for this.

Currently you will get an runtime error:
`RuntimeError: encryption required by Dask configuration, refusing communication from/to 'tcp://'`

(Actually I then accidentially set tls via the scheduler options which made things worse: the workers silently failed to connect)

riedel:
…  or add temporary security if none is given and tls is set

fixes #496


riedel:
3d6a576  fixes #520

riedel:
467c328 fixes #521 , this needs to be very carefully reviewed! It actually skips far more test than before. Those should be all duplicates, though. 

Why is this with in this PR? It is a rather stupid reason. The htcondor ci image had problems installing cryptography as a test dependency.

ci/environment.yml
```diff
@@ -10,3 +10,4 @@ dependencies:
   - black
   - pytest
   - pytest-asyncio
+  - cryptography
```
riedel, line 13: this in needed to generate temporary certificates

ci/htcondor.sh
```diff
@@ -16,7 +16,7 @@ function jobqueue_before_install {
 
 function jobqueue_install {
     cd ./ci/htcondor
-    docker-compose exec -T submit /bin/bash -c "cd /dask-jobqueue; pip3 install -e .;chown -R submituser ."
+    docker-compose exec -T submit /bin/bash -c "cd /dask-jobqueue; pip3 install -e .[test]; chown -R submituser ."
```
riedel, line 19: In the end does not really work because the wheel for cryptography is missing as well as a rust compiler to build it from source. I left it in because it still seemed a sound thing for me to do.

conftest.py
```diff
@@ -20,15 +20,17 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     # register an additional marker
     config.addinivalue_line(
-        "markers", "env(name): mark test to run only on named environment"
+        "markers", "env(NAME): only run test if environment NAME matches"
```
riedel, line 23: I changed this because it contradicted the -E option

conftest.py
```diff
@@ -20,15 +20,17 @@ def pytest_addoption(parser):
...
     )
 
 
 def pytest_runtest_setup(item):
     envnames = [mark.args[0] for mark in item.iter_markers(name="env")]
-    if envnames:
-        if item.config.getoption("-E") not in envnames:
-            pytest.skip("test requires env in %r" % envnames)
+    if (item.config.getoption("-E") is None and envnames) or (
```
riedel, line 29: Now only tests are run either on the local target. This actually a major change in testing!!

dask_jobqueue/core.py
```diff
@@ -450,7 +454,7 @@ def __init__(
         scheduler_cls=Scheduler,  # Use local scheduler for now
         # Options for both scheduler and workers
         interface=None,
-        protocol="tcp://",
+        protocol=None,
```
riedel, line 457: to me this default did not make immediate sense. However one could revert this line I guess

dask_jobqueue/core.py
```diff
@@ -500,6 +504,17 @@ def __init__(
                 "jobqueue.%s.scheduler-options" % config_name, {}
             )
 
+        if protocol is None and security is not None:
+            protocol = "tls://"
+        if security is None and protocol is not None and protocol.startswith("tls"):
+            try:
+                security = Security.temporary()
+            except ImportError:
```
riedel, line 512: I did not want to add an extra dependency. This is also the reason why tls should not be a default.

dask_jobqueue/core.py
```diff
@@ -521,7 +536,26 @@ def __init__(
         job_kwargs["config_name"] = config_name
         job_kwargs["interface"] = interface
         job_kwargs["protocol"] = protocol
-        job_kwargs["security"] = security
+        job_kwargs["security"] = copy.copy(security)
```
riedel, line 539: Adding it at this point changes the object only for the job. I did it centrally though to not generate one temp file per certificate and job

dask_jobqueue/core.py
```diff
@@ -521,7 +536,26 @@ def __init__(
...
+        if security is not None:
+            worker_security_dict = job_kwargs["security"].get_tls_config_for_role(
+                "worker"
+            )
+            for key, value in worker_security_dict.items():
+                # dump worker in-memory keys for use in job_script
+                if value is not None and "\n" in value:
+                    f = tempfile.NamedTemporaryFile(mode="wt")
+                    # make sure that tmpfile survives by keeping a reference
+                    setattr(self, "_job_" + key, f)
```
riedel, line 550: this is ugly, but I having to clean up manually seemed more ugly. Ideas are welcome

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -427,12 +427,11 @@ def test_security():
         require_encryption=True,
     )
 
-    with LocalCluster(
-        cores=1, memory="1GB", security=security, protocol="tls"
-    ) as cluster:
```
riedel, line 432: I reused the old test to test if a security context is enough to use tls

setup.py
```diff
@@ -25,7 +25,7 @@
     packages=["dask_jobqueue"],
     include_package_data=True,
     install_requires=install_requires,
-    tests_require=["pytest >= 2.7.1"],
```
riedel, line 28: as tests_require  seems to be deprecated I changed this line to install the dependencies for the ci automatically. Should make it easier for normal devs as well

riedel (commented):
For reference I commented on the changes I made

riedel:
I am happy to strip down some parts of the PR if anyone is too scared of merging it :) .

ci/htcondor.sh
guillaumeeb: Can you precise what it does?

conftest.py
guillaumeeb: So this is the big unrelated change? Tests not marked are run on local, and those "local" tests are not run anymore on specific job scheduling system tests? But tests are run at least once in some environment right? 

I guess this should be OK. This might in some edge case make some bugs in CI more difficult to spot, but I don't think the situation where "local" tests pass in "no scheduling system env" and fail in "scheduling system marked env" ever occurred.

And this is because of the cryptography dependency with htcondor? It would still be better to fix this, but I imagine you already tried...

dask_jobqueue/core.py
guillaumeeb: Fine by me.

dask_jobqueue/core.py
guillaumeeb: Sounds good.

dask_jobqueue/core.py
guillaumeeb: Why do we need copy?

dask_jobqueue/core.py
guillaumeeb: Can you explain what must be done here? Maybe @jacobtomlinson understands better.

dask_jobqueue/tests/test_jobqueue_core.py
guillaumeeb: Maybe the tests should be duplicated? Not a big deal though.

setup.py
guillaumeeb: Sounds good!

guillaumeeb (commented):
Overall this looks good, I'm just asking for some details in order to decide if we should merge it as is or not.

ci/htcondor.sh
riedel: I am adding the optional test dependencies for HTCondor here, which actually fails, but does not matter due to the change in the test semantics. This might actually work at any given time: I have no clue why this wheel was not prebuild.

conftest.py
riedel: To me the reason is not HTCondor but #521. I just pulled it into this PR because it works around the problem that there is no prebuild wheel for cryptography.  If you think that #521 is debatable, then I would remove this part and try to fix the prebuild image (e.g. by using conda/mamba to preseed the dependencies)

dask_jobqueue/core.py
riedel: because I modify this pass by reference item: I am inlining the certs: see #520. IMHO it confuses people if their object is changed.

riedel:
Like I said: I am happy to try moving the issues related to #521 out of this PR. It is true that the failing HTCondor CI dependency is not really a reason. The reason was that fixing the prebuild CI-Images cannot easily be done in the same PR and there seems not really a reason to fix it in a separate PR, because only this PR breaks it. So it was like a bit of a chicken and egg decision. Would fix this after this PR. However, if you think that #521 should not be fixed, I am happy to take another path.

PR #524, opened 2021-10-20
dump temporary security contexts to file in specified shared temporary directory for use in worker submit scripts

### LINKED ISSUE #520: In memory security contexts are not supported
See https://github.com/dask/distributed/blob/3754a6694f72207b69aaf661445aecf18fb5f0cb/distributed/security.py#L211

i.e. one cannot use Security.temporary() to create a cluster 

distrinuted allows using in memory keys where the content is in memory. It does so by checking if the value is multiline.

the expansion to the extra arguments does not support this:
https://github.com/dask/dask-jobqueue/blob/79a9a3bee55c56195f97b3e70bda15c12c3b19be/dask_jobqueue/core.py#L224

in this case a temporary file needs to be created

### LINKED ISSUE #534: Add pytest xfails for known issue in CI
E.g. #520 could be easily tested even before a fix is merged.

riedel:
Fixes #520 
Closes #534

### DIFF
--- a/ci/pbs.sh
+++ b/ci/pbs.sh
@@ -16,7 +16,7 @@ function jobqueue_before_install {
 }
 
 function jobqueue_install {
-    docker exec pbs_master /bin/bash -c "cd /dask-jobqueue; pip install -e ."
+    docker exec pbs_master /bin/bash -c "cd /dask-jobqueue; pip install -e .; chown -R pbsuser ."
 }
 
 function jobqueue_script {
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -595,6 +595,19 @@ class JobQueueCluster(SpecCluster):
         if n_workers:
             self.scale(n_workers)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        # delete any temporary security
+        if "security" in self._job_kwargs and self._job_kwargs["security"] is not None:
+            worker_security_dict = self._job_kwargs["security"].get_tls_config_for_role(
+                "worker"
+            )
+            for key, value in worker_security_dict.items():
+                if hasattr(self, "_job_" + key):
+                    delattr(self, "_job_" + key)
+
     @property
     def _dummy_job(self):
         """
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -629,7 +629,7 @@ class JobQueueCluster(SpecCluster):
 
         # dumping of certificates only needed if multiline in-memory keys are contained
         if not any(
-            [value is not None and "\n" for value in worker_security_dict.items()]
+            [(value is not None and "\n" in value) for value in worker_security_dict.values()]
         ):
             return security
         # a shared temp directory should be configured correctly
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -595,20 +595,6 @@ class JobQueueCluster(SpecCluster):
         if n_workers:
             self.scale(n_workers)
 
-    def __enter__(self):
-        return self
-
-    def __exit__(self, type, value, traceback):
-        # delete any temporary security
-        if "security" in self._job_kwargs and self._job_kwargs["security"] is not None:
-            worker_security_dict = self._job_kwargs["security"].get_tls_config_for_role(
-                "worker"
-            )
-            for key, value in worker_security_dict.items():
-                if hasattr(self, "_job_" + key):
-                    delattr(self, "_job_" + key)
-        self.close()
-
     @property
     def _dummy_job(self):
         """
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -223,7 +223,6 @@ class Job(ProcessInterface, abc.ABC):
             extra = extra + ["--protocol", protocol]
         if security:
             worker_security_dict = security.get_tls_config_for_role("worker")
-
             security_command_line_list = [
                 ["--tls-" + key.replace("_", "-"), value]
                 for key, value in worker_security_dict.items()
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -537,6 +537,11 @@ class JobQueueCluster(SpecCluster):
             )
 
         if security is not None:
+            if shared_directory is None:
+                raise ValueError(
+                    "Passing a security object to workers is only supported if a shared directory is configured"
+                )
+
             worker_security_dict = job_kwargs["security"].get_tls_config_for_role(
                 "worker"
             )
@@ -556,10 +561,15 @@ class JobQueueCluster(SpecCluster):
                     setattr(self, "_job_" + key, f)
                     f.write(value)
                     f.flush()
+                    # allow expanding of vars and user paths in remote script
+                    if shared_directory is not None:
+                        fname = os.path.join(shared_directory, os.path.basename(f.name))
+                    else:
+                        fname = f.name
                     setattr(
                         job_kwargs["security"],
                         "tls_" + ("worker_" if key != "ca_file" else "") + key,
-                        f.name,
+                        fname,
                     )
 
         self._job_kwargs = job_kwargs
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -627,11 +627,12 @@ class JobQueueCluster(SpecCluster):
 
         worker_security_dict = security.get_tls_config_for_role("worker")
 
-        # only needed if temporary security is specified which only works
+        # dumping of certificates only needed if multiline in-memory keys are contained
         if not any(
             [value is not None and "\n" for value in worker_security_dict.items()]
         ):
             return security
+        # a shared temp directory should be configured correctly
         elif self.shared_temp_directory is None:
             shared_temp_directory = os.getcwd()
             warnings.warn(
@@ -667,7 +668,7 @@ or by setting this value in the config file found in `~/.config/dask/jobqueue.ya
                         )
                     ) from e
 
-                # make sure that tmpfile survives by keeping a reference
+                # make sure that the file is bound to life time of self by keeping a reference to the file handle
                 setattr(self, "_job_" + key, f)
                 f.write(value)
                 f.flush()
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -8,6 +8,8 @@ import subprocess
 import sys
 import weakref
 import abc
+import tempfile
+import copy
 
 import dask
 
@@ -219,6 +221,7 @@ class Job(ProcessInterface, abc.ABC):
             extra = extra + ["--protocol", protocol]
         if security:
             worker_security_dict = security.get_tls_config_for_role("worker")
+
             security_command_line_list = [
                 ["--tls-" + key.replace("_", "-"), value]
                 for key, value in worker_security_dict.items()
@@ -523,7 +526,26 @@ class JobQueueCluster(SpecCluster):
         job_kwargs["config_name"] = config_name
         job_kwargs["interface"] = interface
         job_kwargs["protocol"] = protocol
-        job_kwargs["security"] = security
+        job_kwargs["security"] = copy.copy(security)
+
+        if security is not None:
+            worker_security_dict = job_kwargs["security"].get_tls_config_for_role(
+                "worker"
+            )
+            for key, value in worker_security_dict.items():
+                # dump worker in-memory keys for use in job_script
+                if value is not None and "\n" in value:
+                    f = tempfile.NamedTemporaryFile(mode="wt")
+                    # make sure that tmpfile survives by keeping a reference
+                    setattr(self, "_job_" + key, f)
+                    f.write(value)
+                    f.flush()
+                    setattr(
+                        job_kwargs["security"],
+                        "tls_" + ("worker_" if key != "ca_file" else "") + key,
+                        f.name,
+                    )
+
         self._job_kwargs = job_kwargs
 
         worker = {"cls": self.job_cls, "options": self._job_kwargs}
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -525,15 +525,11 @@ class JobQueueCluster(SpecCluster):
             "options": scheduler_options,
         }
 
-        job_kwargs["config_name"] = config_name
-        job_kwargs["interface"] = interface
-        job_kwargs["protocol"] = protocol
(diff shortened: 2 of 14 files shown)


riedel:
Forget about the deprecation: accidentally looked at the wrong file. Thought tmpfile was finally deprecated as I remembered that there was a discussion in dask#7800 that it was not needed anymore due to the new python tmpfile, which I misunderstood

So I guess it makes sense then also to use the built-in tmpfile in this PR. I cannot remember why I didn't do that.

riedel:
I think #532  should be completed/merged first. Without any testing this is far to shaky...

dask_jobqueue/core.py
```diff
@@ -521,7 +524,26 @@ def __init__(
...
+                    f = tempfile.NamedTemporaryFile(mode="wt")
+                    # make sure that tmpfile survives by keeping a reference
+                    setattr(self, "_job_" + key, f)
+                    f.write(value)
+                    f.flush()
+                    setattr(
+                        job_kwargs["security"],
+                        "tls_" + ("worker_" if key != "ca_file" else "") + key,
+                        f.name,
+                    )
```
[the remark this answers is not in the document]
guillaumeeb: So just to answer even if it looks like we agree: jobscript file creation has a separated function you mentioned, And jobscript need to be visible only by the client instantiating JobQueueCluster, not the workers. So a file in /tmp is perfect.

guillaumeeb:
So instead of #532, I should look into #533 if I understand?

riedel:
Moved test across all clusters here: #537 (will fail for current implementation)

riedel:
Now, it basically works, the only problem is identifying a writable shared directory for testing. Local still works, htcondor and slurm xfail due to the missing shared filesystem, sge works, however, pbs seems has the permissions on the test directory set as non-writable: what would a better place to write the temporary files? 

Second question: Does it make sense to default to the home dir as a shared directory? I guess most clusters work like this, however e.g. the CI does not...

riedel:
BTW: the last remaining error shows that the permissions on the temporary certificate files are set in a secure manner to 0600 .

riedel:
I thought I was done: now I get an error in the unmodified 
` FAILED dask_jobqueue/tests/test_jobqueue_core.py::test_security[LocalCluster]`
and
`"distributed.scheduler - ERROR - Workers don't have promised key: ['tls://10.1.0.206:44087'], lambda-a54fa6a78ad624099ae9b275efae5945"`
this is not good! I was careful not to modify this code...

riedel:
I find 86977af failing very scary. If someone can explain, why closing the cluster can have such effects. However, now everything works as expected (only the temp files live until the garbage collection runs)

riedel:
@guillaumeeb : can you have another look. I fixed the part with the shared filesystem by explicitly introducing an option that defaults to the home directory. I could not think of a way to do this differently.

ci/pbs.sh
```diff
@@ -16,7 +16,7 @@ function jobqueue_before_install {
 }
 
 function jobqueue_install {
-    docker exec pbs_master /bin/bash -c "cd /dask-jobqueue; pip install -e ."
+    docker exec pbs_master /bin/bash -c "cd /dask-jobqueue; pip install -e .; chown -R pbsuser ."
```
riedel, line 19: To dump the certificates the test directory needs to be writable

dask_jobqueue/core.py
```diff
@@ -83,6 +85,8 @@
     scheduler_cls : type
         Changes the class of the used Dask Scheduler. Defaults to  Dask's
         :class:`distributed.Scheduler`.
+    shared_directory : str
```
riedel, line 88: I thought of no better way to specify where to dump the files

dask_jobqueue/core.py
```diff
@@ -520,10 +525,23 @@ def __init__(
...
+            if shared_directory is None:
+                raise ValueError(
+                    "Passing a security object to workers is only supported if a shared directory is configured"
+                )
+
         job_kwargs["config_name"] = config_name
         job_kwargs["interface"] = interface
         job_kwargs["protocol"] = protocol
-        job_kwargs["security"] = security
+        job_kwargs["security"] = self._get_worker_security(security)
```
riedel, line 543: as @guillaumeeb suggested I moved stuff to a function here

dask_jobqueue/core.py
```diff
@@ -607,6 +625,46 @@ def _new_worker_name(self, worker_number):
             cluster_name=self._name, worker_number=worker_number
         )
 
+    def _get_worker_security(self, security):
```
riedel, line 628: This function could be extended to also handle cases where file certificates are outside the shared directory path. For now it only handles creating temporary files. Unfortunately the files will survive until garbage collected (which is probably longer than needed). Security is passed as an argument because the initialization of self done is too late by the super class.

dask_jobqueue/jobqueue.yaml
```diff
@@ -10,6 +10,7 @@ jobqueue:
     interface: null             # Network interface to use like eth0 or ib0
     death-timeout: 60           # Number of seconds to wait if a worker can not find a scheduler
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
+    shared-directory: "~"       # Shared directory between scheduler and worker
```
riedel, line 13: I hope this is OK. It could also default to null, however, this default should work for most users. If a cluster has no shared directory it should be set to null though manually.

riedel (commented):
added a few explainations.

dask_jobqueue/core.py
```diff
@@ -520,10 +525,23 @@ def __init__(
...
+            shared_directory = dask.config.get(
+                "jobqueue.%s.shared-directory" % config_name
+            )
+        self.shared_directory = shared_directory
+
+        if security is not None:
+            if shared_directory is None:
+                raise ValueError(
+                    "Passing a security object to workers is only supported if a shared directory is configured"
+                )
```
andersy005, line 538:
```suggestion
        if security is not None and shared_directory is None:
            raise ValueError(
                "Passing a security object to workers is only supported if a shared directory is configured"
            )
```

andersy005: Feel free to ignore this, but since this is a new feature it might be helpful to make this error message a little verbose to include pointers on how to set this shared directory. e.g. 

```python
raise ValueErrror("Passing a security object to workers is only supported if a shared directory is configured. \
                  You can set this value by using dask for e.g. `dask.config.set({'jobqueue.pbs.shared_directory': '/home/foo'})` or by setting this value in the config file found in `~/.config/dask/jobqueue.yaml` ")
```

riedel: tI like this. Actually it made me think that it should be an warning at this point because there is the rare case that certificates can be predeployed on the compute nodes. So I will a) make this verbose and b) make it a warning and only later an error if temp certificates are used. The basic problem is that the actual error a user will encounter is very much hidden and only will lead to timeouts because no workers are started.

dask_jobqueue/jobqueue.yaml
riedel: There is pros and cons. 
for null:
- any user who wants to use security without thinking needs to set this
+ it is a safe default and will make it easier to find errors if it does not work

for "~":
+ the home directory is probably a sensible default that 90% of the cluster users
+ if a user wants to just use security together with #527 it will work out of the box 
- if things go wrong here it gets difficult

So together with a verbose warning/error message I feel confortable to set this to `null` particularly since adding "shared_directory="~", security=True" is not that much of an overhead. It makes sense to adds docs as part of #527 . I guess better safe than sorry.

dask_jobqueue/core.py
guillaumeeb: Please consider @andersy005 suggestion here!

Also, we might decide to specify a default even if this is None (something like: use the current working directory), so maybe a warning would be better.

dask_jobqueue/core.py
```diff
@@ -607,6 +625,46 @@ def _new_worker_name(self, worker_number):
...
+        """Dump temporary parts of the security object into a shared_directory"""
+        if security is not None:
+            copied_on_write = False
+            worker_security_dict = security.get_tls_config_for_role("worker")
+            for key, value in worker_security_dict.items():
+                # dump worker in-memory keys for use in job_script
+                if value is not None and "\n" in value:
+                    if not copied_on_write:
+                        security = copy.copy(security)
+                        copied_on_write = True
```
guillaumeeb, line 638: Can't we just copy security object as the beginning and dropt this boolean variable? I do not find it super useful... But maybe I've missed something.

dask_jobqueue/core.py
```diff
@@ -607,6 +625,46 @@ def _new_worker_name(self, worker_number):
...
+                    f = tempfile.NamedTemporaryFile(
+                        mode="wt",
+                        prefix=".dask-jobqueue.worker." + key + ".",
+                        dir=(
+                            os.path.expanduser(
+                                os.path.expandvars(self.shared_directory)
+                            )
+                            if self.shared_directory is not None
+                            else None
```
guillaumeeb, line 647: What is the behavior if dir is None here?

dask_jobqueue/jobqueue.yaml
guillaumeeb: For me, this config option is more like `log_directory`. And I would put it to null by default, and in the code just try the current directory (with a warning?) as a base path for writing security temp files.

This way, I'd think than in most setups this should work without setting something (you usually run your scripts from a shared directory, HOME dir or an HPC shared file system). But I might be wrong here.

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -413,3 +413,62 @@ def test_security(EnvSpecificCluster, loop):
...
     ) as cluster:
         assert "tls://" in job_script
+
+
+@pytest.mark.xfail_env({"htcondor": "#535 no shared filesystem in htcondor ci"})
+@pytest.mark.xfail_env({"slurm": "#535 no shared filesystem in slurm ci"})
+def test_security_temporary(EnvSpecificCluster, loop):
+    dirname = os.path.dirname(__file__)
+    with open(os.path.join(dirname, "key.pem"), "r") as file:
+        key = file.read()
```
guillaumeeb, line 423: I've just some doubts here: what file are you reading? When was it created?

guillaumeeb (commented):
We need to find an agreement on the configuration name and the default (even that is looks every one agrees for defaulting the config option to None/null).

dask_jobqueue/tests/test_jobqueue_core.py
guillaumeeb: OK, just seen this came from previous modifications. This certs were already there.

dask_jobqueue/core.py
riedel: I am indifferent here.  ```temp_security_directory``` also sounds OK to me or maybe ```shared_temp_directory```, so there is a chance for reuse?

dask_jobqueue/jobqueue.yaml
riedel: OK with me if we do the renaming above. For a shared directory I would have though "null" means that there is none. I'll fix once the name is decided.

dask_jobqueue/core.py
riedel: Yes I will do a warning and make this more narrow. Actually this covers cases also where there is a fixed cert given.

dask_jobqueue/core.py
riedel: This just makes sure that everything is the same if no temp cert is given. Not super useful indeed, just wanted to be safe. I'll remove that

dask_jobqueue/core.py
riedel: Actually I will rather assert that this is not None by using a default. Then this is more clear.

dask_jobqueue/tests/test_jobqueue_core.py
riedel: yes the are part of the test suite. I could generate them however this adds a test dependency to `cryptography`

dask_jobqueue/core.py
guillaumeeb: OK for `shared_temp_directory`, but make it clear in the comment that this is currently used for writing temporary security files.

dask_jobqueue/jobqueue.yaml
guillaumeeb: We have to make it clear somewhere in the comment (or warning) that a null shared directory default to current directory if needed.

dask_jobqueue/tests/test_jobqueue_core.py
guillaumeeb: It's OK that way I think. Except we don't test the generation though. I'm okay to add cryptography as a test dependency!

riedel: I changed this so it to use `Security.temporary() `. This also better shows the most common use case I guess.

dask_jobqueue/core.py
```diff
@@ -83,6 +86,8 @@
     scheduler_cls : type
         Changes the class of the used Dask Scheduler. Defaults to  Dask's
         :class:`distributed.Scheduler`.
+    shared_temp_directory : str
+        Shared directory between scheduler and worker defaults to "~"
```
guillaumeeb, line 90:
```suggestion
        Shared directory between scheduler and worker (used for example by temporary security certificates) defaults to current working directory if not set.
```

guillaumeeb (approved):
Just one last suggestion, and then I think we can merge!!

dask_jobqueue/core.py
```diff
@@ -607,6 +620,73 @@ def _new_worker_name(self, worker_number):
...
+    def _get_worker_security(self, security):
+        """Dump temporary parts of the security object into a shared_temp_directory"""
+        if security is None:
+            return None
+
+        worker_security_dict = security.get_tls_config_for_role("worker")
+
+        # dumping of certificates only needed if multiline in-memory keys are contained
+        if not any(
+            [value is not None and "\n" for value in worker_security_dict.items()]
```
riedel, line 632: Found a last problem: is always true as in value is missing in a copy and paste error!

riedel:
Since #527 (which is the final thing I want to get merged) passes I am happy now.

guillaumeeb:
Thanks a lot for all the work here @riedel, I know this was long and with a lot of waiting periods.

riedel:
Thank you alot, @guillaumeeb! I am just happy that this part is in now and encrypting traffic now easily works. Also @andersy005 for reviewing.

PR #532, opened 2021-10-31
add functions for testing in cluster specific enviroments

riedel:
instead of testing different scheduler classes with the same test code on a local system, actually also allow iterating through different CIs

add test for security inside ci as an example

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -25,7 +25,13 @@ def pytest_configure(config):
 
 
 def pytest_runtest_setup(item):
-    envnames = [mark.args[0] for mark in item.iter_markers(name="env")]
+    envnames = sum(
+        [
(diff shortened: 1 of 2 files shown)


riedel:
As expected htcondor fails here because actually no shared filesystem is configured...

riedel:
I actually just noticed that this is a different take at #353 without breaking much. Would it make sense to move the helper functions a central place and rather start rebuild things more iteratively. One test by one. The merged pytest -E makes this rather straight forward without overtesting.

riedel:
If this is seriously considered for merging I would mark the failing ones as ```xfail```. Any feedback welcome! I think this is really needed to efficiently test security across all schedulers.

riedel:
close in favour of general overhaul in #533

Issue #521, opened 2021-09-12
pytest -E option does not do what it says


riedel:
```  -E NAME               only run tests matching the environment NAME.```

instead also all unmarked tests are run.

riedel:
to explain a bit more: the side effect is that common tests are run locally on the CI of the different cluster managers. There seems to be no real benefit from that other than it's slow

PR #533, opened 2021-11-01
rewrite testing across clusters using fixtures

riedel:
* removes considerably more lines than added
* rigorously tests accross all scheduler implementations: tests increase from 168 items to 231 items (not all are actually tested because there is no seperate CI for some)
* move logic for paramaterized testing across multiple schedulers to two fixtures (one that runs locally and one that requires the CI environment): this allows adding new schedulers easily
* add additional required disk argument for HTCondor to fixture using a temporary config default
* test security on all clusters and add xfail for security on HTCondor and Slurm CI
* modify the LocalCluster job_script to use a shebang line in order to behave equally to "real" schedulers (instead of "parsing" the jobscript and "emulating" a shell interpreter)

Please not that this conflicts and is not in sync with @lesteve's #353 . But it might be easier now since #521 was "fixed" .

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -31,11 +31,13 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
-    # register an additional marker
+    # register additional markers
     config.addinivalue_line(
         "markers", "env(NAME): only run test if environment NAME matches"
     )
-    config.addinivalue_line("markers", "xfail_ci(NAME): know CI failure")
+    config.addinivalue_line(
+        "markers", "xfail_env(NAME): known failure for environment NAME"
+    )
 
 
 def pytest_runtest_setup(item):
@@ -55,7 +57,7 @@ def pytest_runtest_setup(item):
         pytest.skip("test requires env in %r" % envnames)
     else:
         xfail = {}
-        [xfail.update(mark.args[0]) for mark in item.iter_markers(name="xfail_ci")]
+        [xfail.update(mark.args[0]) for mark in item.iter_markers(name="xfail_env")]
         if env in xfail:
             item.add_marker(pytest.mark.xfail(reason=xfail[env]))
 
@@ -87,22 +89,19 @@ all_envs = {
 }
 
 
-@pytest.fixture
-def EnvSpecificCluster3(pytestconfig):
-    return all_envs[pytestconfig.getoption("-E")]
-
-
 @pytest.fixture(
     params=[pytest.param(v, marks=[pytest.mark.env(k)]) for (k, v) in all_envs.items()]
 )
 def EnvSpecificCluster(request):
+    """Run test in only with the specific cluster class set by the environment"""
     if request.param == HTCondorCluster:
-        dask.config.set({"jobqueue.htcondor.disk": "20MB"})
+        dask.config.set({"jobqueue.htcondor.disk": "1GB"})
     return request.param
 
 
 @pytest.fixture(params=list(all_envs.values()))
 def Cluster(request):
+    """Run for each cluster class independent when no environment is set"""
     if request.param == HTCondorCluster:
         dask.config.set({"jobqueue.htcondor.disk": "1GB"})
     return request.param
--- a/conftest.py
+++ b/conftest.py
@@ -7,6 +7,18 @@ import pytest
 
 import dask_jobqueue.lsf
 
+from dask_jobqueue import (
+    PBSCluster,
+    MoabCluster,
+    SLURMCluster,
+    SGECluster,
+    LSFCluster,
+    OARCluster,
+    HTCondorCluster,
+)
+
+from dask_jobqueue.local import LocalCluster
+
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -22,15 +34,29 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "env(NAME): only run test if environment NAME matches"
     )
+    config.addinivalue_line("markers", "xfail_ci(NAME): know CI failure")
 
 
 def pytest_runtest_setup(item):
-    envnames = [mark.args[0] for mark in item.iter_markers(name="env")]
-    if (item.config.getoption("-E") is None and envnames) or (
-        item.config.getoption("-E") is not None
-        and item.config.getoption("-E") not in envnames
+    env = item.config.getoption("-E")
+    envnames = sum(
+        [
+            mark.args[0] if isinstance(mark.args[0], list) else [mark.args[0]]
+            for mark in item.iter_markers(name="env")
+        ],
+        [],
+    )
+    if (
+        None not in envnames
+        and (env is None and envnames)
+        or (env is not None and env not in envnames)
     ):
         pytest.skip("test requires env in %r" % envnames)
+    else:
+        xfail = {}
+        [xfail.update(mark.args[0]) for mark in item.iter_markers(name="xfail_ci")]
+        if env in xfail:
+            item.add_marker(pytest.mark.xfail(reason=xfail[env]))
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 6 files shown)


riedel:
IMHO this makes testing alot easier to read

dask_jobqueue/tests/test_job.py
```diff
@@ -1,61 +1,24 @@
...
-    OARCluster,
-    HTCondorCluster,
-]
-
-
-@pytest.mark.parametrize("job_cls", job_protected)
 @pytest.mark.asyncio
-async def test_job(job_cls):
+async def test_job(EnvSpecificCluster):
+    job_cls = EnvSpecificCluster.job_cls
```
riedel, line 21: The other changes are similar: The logic `pytest.mark.env` logic is moved into the EnvSpecificClusterFixture

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -15,44 +15,22 @@
...
 def test_errors():
     match = re.compile("Job type.*job_cls", flags=re.DOTALL)
     with pytest.raises(ValueError, match=match):
         JobQueueCluster(cores=4)
 
 
-def test_command_template():
-    with PBSCluster(cores=2, memory="4GB") as cluster:
+def test_command_template(Cluster):
+    with Cluster(cores=2, memory="4GB") as cluster:
```
riedel, line 33: I tried to make that test, that run across all clusters are actually tested across all clusters (or in rare cases only `LocalCluster`)

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -275,25 +229,22 @@ def __init__(self, *args, **kwargs):
         MyCluster(cores=1, memory="1GB")
 
 
-@pytest.mark.parametrize("Cluster", all_clusters)
 def test_default_number_of_worker_processes(Cluster):
-    with create_cluster_func(Cluster, cores=4, memory="4GB") as cluster:
+    with Cluster(cores=4, memory="4GB") as cluster:
```
riedel, line 233: Because I moved also the special handling of the `HTCondorCluster ` to the fixture even for local testing things have been becoming much more concise

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -117,25 +91,6 @@ def test_forward_ip():
         assert cluster.scheduler.ip == default_ip
 
 
-@pytest.mark.parametrize("Cluster", [])
```
riedel, line 120: Actually I did not know what to make out of those. There are a few later that probably should also be addressed. Eg. also by moving to the apropriate cluster specific testing...

riedel (commented):
I tried to add a few explanations

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -407,12 +362,12 @@ def test_wrong_parameter_error(Cluster):
...
-        create_cluster_func(
-            Cluster, cores=1, memory="1GB", wrong_parameter="wrong_parameter_value"
-        )
+        Cluster(cores=1, memory="1GB", wrong_parameter="wrong_parameter_value")
 
 
-def test_security():
+@pytest.mark.xfail_ci({"htcondor": "#535 no shared filesystem in htcondor ci"})
+@pytest.mark.xfail_ci({"slurm": "#535 no shared filesystem in slurm ci"})
+def test_security(EnvSpecificCluster, loop):
```
riedel, line 370: This was only tested with LocalCluster before, so that it would not fail although a shared file system is really needed for this to work.

conftest.py
```diff
@@ -22,15 +35,29 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "env(NAME): only run test if environment NAME matches"
     )
+    config.addinivalue_line("markers", "xfail_ci(NAME): know CI failure")
```
guillaumeeb, line 38:
```suggestion
    config.addinivalue_line("markers", "xfail_ci(NAME): known CI failure")
```

conftest.py
```diff
@@ -46,3 +73,36 @@ def mock_lsf_version(monkeypatch, request):
...
+    "slurm": SLURMCluster,
+    "sge": SGECluster,
+    "lsf": LSFCluster,
+    "oar": OARCluster,
+    "htcondor": HTCondorCluster,
+}
+
+
+@pytest.fixture
+def EnvSpecificCluster3(pytestconfig):
```
guillaumeeb, line 91: Not sure about the final `3` here? Not sure about this definition at all... Could you explain?

dask_jobqueue/core.py
```diff
@@ -123,7 +123,7 @@ class Job(ProcessInterface, abc.ABC):
 
 %(job_header)s
 %(env_header)s
-%(worker_command)s
+exec %(worker_command)s
```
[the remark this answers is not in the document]
guillaumeeb: I don't get it, and I have to admit I'm reluctant to add this change.

dask_jobqueue/jobqueue.yaml
```diff
@@ -195,6 +195,7 @@ jobqueue:
     death-timeout: 60           # Number of seconds to wait if a worker can not find a scheduler
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
     extra: []
+    shebang: "#!/usr/bin/env bash"
```
guillaumeeb, line 198: Why do we need this?

dask_jobqueue/local.py
```diff
@@ -38,11 +38,7 @@ def __init__(
     ):
         # Instantiate args and parameters from parent abstract class
         super().__init__(
-            scheduler=scheduler,
-            name=name,
-            config_name=config_name,
-            shebang="",
-            **kwargs
+            scheduler=scheduler, name=name, config_name=config_name, **kwargs
```
guillaumeeb, line 41: Does this work with black linting?

conftest.py
```diff
@@ -46,3 +73,36 @@ def mock_lsf_version(monkeypatch, request):
...
+def EnvSpecificCluster(request):
+    if request.param == HTCondorCluster:
+        dask.config.set({"jobqueue.htcondor.disk": "20MB"})
+    return request.param
+
+
+@pytest.fixture(params=list(all_envs.values()))
+def Cluster(request):
+    if request.param == HTCondorCluster:
+        dask.config.set({"jobqueue.htcondor.disk": "1GB"})
```
guillaumeeb, line 107: Why 1GB here and 20MB above?

conftest.py
```diff
@@ -46,3 +73,36 @@ def mock_lsf_version(monkeypatch, request):
...
+    params=[pytest.param(v, marks=[pytest.mark.env(k)]) for (k, v) in all_envs.items()]
+)
+def EnvSpecificCluster(request):
+    if request.param == HTCondorCluster:
+        dask.config.set({"jobqueue.htcondor.disk": "20MB"})
+    return request.param
+
+
+@pytest.fixture(params=list(all_envs.values()))
+def Cluster(request):
```
guillaumeeb, line 105: So Cluster is all envs, but without specific CI?

guillaumeeb (commented):
This looks really neat!!

I just have a few questions and clarification, and a bit reluctance changing the submission script template.

Otherwise this would be a much welcomed change!

dask_jobqueue/core.py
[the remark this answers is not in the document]
riedel: I will try to remove this part from the PR. It is a direct consequence of changing this to actually use the script instead of "interpreting" it: https://github.com/dask/dask-jobqueue/blob/8264b8cd6fa028f35665a554c5fea4be34a97781/dask_jobqueue/local.py#L53-L59 . My guess is that it could make a second PR.

conftest.py
riedel: Yes. I thus runs as part of the default pytests without -E . I'll add a comment here to document this.

riedel:
@guillaumeeb , thanks for reviewing and catching some issues, I tried to incorporate them. I removed the changes to local.py and core.py as they are debatable and can go in a seperate PR . Hope no concerns are left. Added some docstrings to the fixtures.

riedel (commented):
all your comments should be addressed ?!?

guillaumeeb:
Thanks once more @riedel !

PR #536, opened 2021-11-23
Fix docstring as HTCondor needs no shared filesystem

### LINKED ISSUE #535: HTCondor and slurm CIs have no shared filesystem
This why one cannot e.g. test security contexts. This actually could be even desirable as jobqueue probably does not even need one. However, in this case the security context code needs to be rewritten (found nothing else depending on this). 

This is a bit paradox due to this line:
https://github.com/dask/dask-jobqueue/blob/8264b8cd6fa028f35665a554c5fea4be34a97781/dask_jobqueue/htcondor.py#L236

riedel:
Closes #535

### DIFF
--- a/dask_jobqueue/htcondor.py
+++ b/dask_jobqueue/htcondor.py
@@ -233,7 +233,7 @@ def quote_environment(env):
 
 
 class HTCondorCluster(JobQueueCluster):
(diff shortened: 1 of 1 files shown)


guillaumeeb:
I think it is safe to merge this one, as CI demonstrate that shared file system is not required!

PR #537, opened 2021-11-27
added xfailing test for known temporary security bug

### LINKED ISSUE #534: Add pytest xfails for known issue in CI

riedel:
adds (failing) test for #520
Closes #534


guillaumeeb:
Is this really needed as #524 is on the way? 

If I'm correct, this is a test that will be updated in #524?

riedel:
You are right. I did this before and used the code in #524 , so I close it now.

PR #538, opened 2021-11-28
Fix parallel testing

riedel:
to support testing using e.g. pytest-xdist or if dask is already running


guillaumeeb (approved):
A bit complicated to generate the port number, but as this is in some test functions, it's okay for me.

PR #545, opened 2022-02-18
Drop support for Python 3.6

### LINKED ISSUE #541: Push python ci dependency to python=3.7
as dask dropped support for 3.6 in dask/dask#7006

https://github.com/dask/dask-jobqueue/blob/4980e746e9be15e5fe6736b6c496b8faea737fd7/ci/environment.yml#L6

actually fixing #539 could help making all CI compatible as well 

The actual problem is that one might acutally pull in a newer click dependency which triggers this bug dask/dask#7658 which is only fixed in newer dask versions

lesteve:
The docker builds have been failing for a few days, see e.g. https://github.com/dask/dask-jobqueue/runs/5240847715?check_suite_focus=true. For the record Dask has moved to 3.8 in the dev version: https://github.com/dask/dask/pull/8572 so we are lagging a bit behind.

Fix #541.

I am pretty sure this error should go away with using 3.7.

For now, the condor build is still using 3.6 (this is the system python in the image). Probably we could install a conda env like in the other scheduler images.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python=3.6
+  - python=3.7
   - dask
   - distributed
   - flake8
--- a/ci/htcondor.sh
+++ b/ci/htcondor.sh
@@ -7,6 +7,7 @@ function jobqueue_before_install {
     # start htcondor cluster
     cd ./ci/htcondor
     docker-compose pull
+    docker-compose build
     ./start-htcondor.sh
     cd -
 
--- a/ci/pbs.sh
+++ b/ci/pbs.sh
@@ -7,6 +7,7 @@ function jobqueue_before_install {
     # start pbs cluster
     cd ./ci/pbs
     docker-compose pull
+    docker-compose build
     ./start-pbs.sh
     cd -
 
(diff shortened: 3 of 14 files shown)


jacobtomlinson (commented):
Just FYI we've just dropped Python 3.7 in dask/community#213. So it might be best to jump to 3.8/3.9/3.10 here.

lesteve:
Hmmm CI fails because it is using the Docker images and the Docker images still have Python 3.6. This setup is a bit fragile when we need to update something in the Docker images. I see two possible quick fixes:
- merge this PR and wait for the daily scheduled build to update the Docker images
- update the docker image by hand

There is probably a better way to do it, it seems like using `daskdev/dask` Docker image for the conda environment may be something we want to look at.

@jacobtomlinson yep I saw afterwards that Dask had dropped 3.7 in the development version.

lesteve:
Hmmm there appears to be real test failures, maybe because we were using an old dask and distributed versions in our Docker image for quite a while.

lesteve:
I am going to merge this one to at least get the docker images build and open an issue about the problems we have in the CI

Also I think it is better to do `docker-compose pull` + `docker-compose build` to take into account changes in the docker images in a PR.

PR #554, opened 2022-06-11
Fix a bug in resource specification

SultanOrazbayev:
Tracing the edited lines back to about 4 years ago, I believe this was a typo/error in how resources are specified. There is a related StackOverflow question here: https://stackoverflow.com/a/72581445/10693596

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -177,6 +177,4 @@ specified for certain steps in the processing. For example:
     stage_1 = [delayed(step_1_w_single_GPU)(i) for i in range(10)]
     stage_2 = [delayed(step_2_w_local_IO)(s2) for s2 in stage_1]
 
-    result_stage_2 = client.compute(stage_2,
-                                    resources={tuple(stage_1): {'GPU': 1},
-                                               tuple(stage_2): {'ssdGB': 100}})
+    result_stage_2 = client.compute(stage_2, resources={'GPU': 1})


docs/source/examples.rst
```diff
@@ -177,6 +177,4 @@ specified for certain steps in the processing. For example:
     stage_1 = [delayed(step_1_w_single_GPU)(i) for i in range(10)]
     stage_2 = [delayed(step_2_w_local_IO)(s2) for s2 in stage_1]
 
-    result_stage_2 = client.compute(stage_2,
-                                    resources={tuple(stage_1): {'GPU': 1},
-                                               tuple(stage_2): {'ssdGB': 100}})
+    result_stage_2 = client.compute(stage_2, resources={'GPU': 1})
```
delgadom, line 180: To me, it seems like there's an issue with this whole example. Your patch does make the code executable but now the text doesn't make sense. Specifically:

> Additionally, required resources can be specified for certain steps in the processing. For example:

This indicates to me that the example will illustrate how to specify specific resource constraints at the task level.

You point to some very relevant links in the SO post, including pointers to how to do this using [annotations](https://distributed.dask.org/en/stable/resources.html) or a [high level graph](https://docs.dask.org/en/stable/high-level-graphs.html).

I'll admit I have very little experience with heterogenous SpecClusters, and have no idea how to test them. But it seems like this example should either be deleted or should read something along these lines:

```python
with dask.annotate(resources={'GPU': 1}):
    stage_1 = [delayed(step_1_w_single_GPU)(i) for i in range(10)]

with dask.annotate(resources={'ssdGB': 100}):
    stage_2 = [delayed(step_2_w_local_IO)(s2) for s2 in stage_1]

result_stage_2 = client.compute(stage_2)
```

SultanOrazbayev: Agree, best case is to rewrite the whole example.

guillaumeeb: Thanks for the PR @SultanOrazbayev, but I agree with @delgadom, this dos not demonstrate what was intended into the example where different resources are needed for different steps of the processing.

guillaumeeb (changes requested):
Either we delete the whole example, either we find how to specify multiple resources with current Dasks version.

First solution is simpler, second solution is better (if this is still possible).

jacobtomlinson:

Issue #393, opened 2020-03-16
Incorrect description for env_extra for HTCondorCluster


matyasselmeci:
Hi,

The description for env_extra in [HTCondorCluster](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.HTCondorCluster.html#dask_jobqueue.HTCondorCluster) is not correct: the job that HTCondorCluster creates calls dask-worker directly instead of through a bash wrapper script, so you cannot put arbitrary shell commands into env_extra.

The interface supports environment variables as `key=value` pairs, which will be inserted into dask-worker's environment (via the "Environment" attribute in the submit file). (For consistency, you can write `export foo=bar` but the word "export" will be ignored.)

This is also important to keep in mind with regards to #323; renaming env_extra to job_script_extra or similar would be even more inaccurate (for the HTCondor case anyway).

lesteve:
Thanks for the info!

Not sure what to do about this I have to say. I would be tempted to say that there should be a HTCondor-specific argument to pass environment variables to HTCondor and that `env_extra` should be lines inserted as is in the job script.

I know nothing about HTCondor so that may not make that much sense ... I think it would help a lot if you had examples of
- example use cases for `job_extra` and `env_extra` in `HTCondor` now
- example use cases for `job_script_extra` if it existed i.e. some lines that are inserted in the job script as is.

matyasselmeci:
This is a complicated answer because HTCondor has different semantics than the other batch schedulers.  In PBS for example, you've got a shell script with some magic comments on top for controlling the batch scheduling, so essentially you have the entire job in a single file.

In HTCondor, the commands for controlling the scheduling are in a separate file, called a _submit file_, which is mostly a key=value file (aside from the `Queue` at the end).  Here's an example:

```
Output = job.out
Error = job.err
Log = job.log

RequestCpus = 1
RequestMemory = 10MB
Executable = /usr/bin/sleep
Arguments = 30

Queue
```

`HTCondor.job_script()` essentially turns a dictionary into this file.


So if you want to run multiple remote commands in a single job, put them in a shell script and use that script as the `Executable`.  (HTCondor will transfer the Executable to the remote machine.)


How this applies to your questions:

* `job_extra` is for adding arbitrary KV pairs to the submit file; this can be useful for requesting custom resources (as in `Request_GPUs=1`) or requiring certain systems (`Requirements=(OpSysAndVer=="CentOS7")`).

[...]

Issue #323, opened 2019-08-23
job_extra, extra and env_extra naming could be improved


lesteve:
* `job_extra` -> `job_header_extra`
* `env_extra` -> `job_script_extra`? Not great, better suggestions welcome! Not sure if there is a good way to indicate that this is something which is at the beginning of the script but after the shebang and job header (and before we run dask-worker).
* `extra` -> `dask_worker_additional_args`, better suggestion welcome!

We may want to want to warn for a few releases and point to the new name parameter names, same thing for values from the config `job-extra` and `env-extra` if they are used.

guillaumeeb:
I agree that the naming is not ideal, but do we really want to modify it? I'm not sure...

That said, that's not a strong opposition.

In batchspawner, `env_extra` is called `prologue`, https://github.com/jupyterhub/batchspawner/blob/master/batchspawner/batchspawner.py#L507.

mrocklin:
We can always make a choice to break many APIs, and announce a major version bump.  

Most dask-jobqueue users today are early adopters, and would probably be tolerant of such an action.  It might be worth thinking about what other API changes we might want to make.

lesteve:
This seems another candidate for breaking API changes, we have both `mem` and `memory` in `LSFCluster`:
https://github.com/dask/dask-jobqueue/issues/205

riedel:
One could easily have both and add a [deprecation warning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning), so I would argue that there could be coexisting until a major bump.

jolange:
I'd like to give this a shot soon, because I'm eager to have possible changes included in the next release.

> One could easily have both and add a [deprecation warning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning), so I would argue that there could be coexisting until a major bump.

@riedel I agree, but to my understanding, a [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) would be better suited. In the used terminology, I think we want to warn "end users", not only "other developers".
If I add a `DeprecationWarning` in `Job` and run a script instantiating a `*Cluster`, the warning will be suppressed by default.
See also [below this table](https://docs.python.org/3/library/warnings.html#warning-categories):
[...]

PR #556, opened 2022-07-08
add `pre_exec` job parameter to allow for setup commands

jolange:
This allows to specify additional commands to be run before the actual `"%(python)s -m distributed.cli.dask_worker"`-call that starts the worker. This can be used to set up/source a virtual environment (or conda, mamba, ...), load a module, or whatever setup is needed.

Example usage for a venv:
```python
pre_exec = ['cd ~/projects/dask-jobqueue/example', 'source venv/bin/activate']
cluster = HTCondorCluster(cores=1, memory="2GB", disk="4GB", log_directory = 'logs',
                          pre_exec=pre_exec, python='python3')
```

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -60,6 +60,10 @@ job_parameters = """
     python : str
         Python executable used to launch Dask workers.
         Defaults to the Python that is submitting these jobs
+    pre_exec : list
+        Additional commands to execute before launching the actual
+        Dask worker. For example sourcing a virtual environment,
+        loading a module, etc.
     config_name : str
         Section to use from jobqueue.yaml configuration file.
     name : str
@@ -159,6 +163,7 @@ class Job(ProcessInterface, abc.ABC):
         log_directory=None,
         shebang=None,
         python=sys.executable,
+        pre_exec=None,
         job_name=None,
         config_name=None,
     ):
@@ -268,7 +273,10 @@ class Job(ProcessInterface, abc.ABC):
         if extra is not None:
             command_args += extra
 
-        self._command_template = " ".join(map(str, command_args))
+        self._command_template = ""
+        if pre_exec is not None:
+            self._command_template += "; ".join(pre_exec) + "; "
+        self._command_template += " ".join(map(str, command_args))
 
         self.log_directory = log_directory
         if self.log_directory is not None:


guillaumeeb (commented):
Hi @jolange,

Thanks for raising this PR. It's really appreciated.

In this case, I'm not sure it is needed: doesn't the `env-extra` kwarg already suites your need?

See https://github.com/dask/dask-jobqueue/blob/main/dask_jobqueue/core.py#L248.

Or in the docs: https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html for example.

Maybe the name is not rightly chosen.

jolange:
Hi @guillaumeeb ,
thanks for looking into this! I also first thought that `env_extra` might work, but unfortunately this is not sufficient: At least for HTCondor this only lets you specify environment variables (the function parsing that is defined [here](https://github.com/dask/dask-jobqueue/blob/ac0ca25d5df1b5975f18e878f32c83285b8b4798/dask_jobqueue/htcondor.py#L121-L133)).
So something like
```python
cluster = HTCondorCluster(cores=1, memory="2GB", disk="4GB", log_directory = 'logs',
                          env_extra=['TESTVAR=1', 'export TESTVAR2=2'],
                          )
```
Generates this in an HTCondor submission file:
```
Environment = "TESTVAR=1 TESTVAR2=2"
```
But things like this
```python
env_extra=['cd ~/projects/dask-jobqueue/example', 'source venv/bin/activate'],
```
simply get swallowed and do not appear anywhere.

jolange:
*Addendum:*
You are right that this works for e.g. SLURM. There all lines are simply added to the script (which is much simpler than the HTCondor submit files).
Do you consider this a "bug" in `HTCondorCluster`? I could also take all lines that are not selected by `env_lines_to_dict()` (see link above) and prepend those to the command instead of introducing a new parameter. This would be a bit intransparent to the user, though, because some commands are prepended and some are put into the HTCondor `Environment` parameter.

mivade:
The way HTCondor submit files work I think it would be rather confusing to allow `env_extra` to include arbitrary commands (frankly I don't think that makes sense for any cluster). I think having some kind of helper script argument like the one proposed here would make sense if it's necessary, but I'm also confused about the purpose. As implemented `sys.executable` sets the Python interpreter to use which should be the same as the one in your virtual environment. Is there a particular reason that was not sufficient?

ocaisa:
A common use case that I've had is to use `env_extra` to load environment modules to get the computing environment that I need for a particular set of tasks:
```
GROMACS_gpu_cluster = CustomSLURMCluster(
    name="GROMACS_gpu_cluster",
    nodes=2,
    mpi_mode=True,
    fork_mpi=True,
    queue_type="gpus",
    maximum_jobs=5,
    env_extra=[
        "module load Intel",
        "module load IntelMPI",
        "module load GROMACS",
        "module load dask",
        "module load jobqueue_features",
    ],
)
```
The code here is customised but the use case is hopefully clear.

jolange:
> The way HTCondor submit files work I think it would be rather confusing to allow `env_extra` to include arbitrary commands (frankly I don't think that makes sense for any cluster).

What I got from the discussion so far is that the parameter rather is a misnomer.

> ... but I'm also confused about the purpose. As implemented `sys.executable` sets the Python interpreter to use which should be the same as the one in your virtual environment.

Unfortunately that's not sufficient, because that's a symlink (`venv/bin/python3 -> /usr/bin/python3`) and `sys.executable` is set to the target. Furthermore we regularly need to load environment modules like @ocaisa described and also load software/packages from other sources (like cvmfs).

jolange:
Either way would be fine for me. I guess it's up to the maintainers to decide here.

jolange:
Thanks  @riedel for the feedback!
> differ from slurm in that they are no scripts themselves but have their own syntax which makes things a bit different

Right, this is why I prepend commands in the `Arguments =` parameter to achieve what the scripts do for e.g. SLURM. Here the commands are separated by semicolon instead of newlines, but the logic is the same, isn't it?
Anyway, I would be happy to have a solution with a "proper" script (=with statements separated with newlines) to run in the job, because currently it is a quoting nightmare.

That being said, I don't fully get your proposal, yet:
- Do you mean `condor_submit` in the script instead of `condor_q`?
- Is `EOF` supposed to go before the `fi`?
- What is `submit command` supposed to be? The actual python call starting the worker? (`%(python)s -m distributed.cli.dask_worker ...`)
- I don't get where the "script that submits itself"-magic would happen.

guillaumeeb:
Hi everyone,

We all agree this is important for any queuing system to be able to run some commands before starting the Dask worker.

Currently, this is supposed to be done using `env_extra` as stated in the doc: 

> Other commands to add to script before launching worker.

For HTCondor, the command only add environment variables, probably because of a misunderstanding due to the kwarg name, and also because HTCondor is a bit specific.

So I will probably reject this as is, but two things could be done:
- propose and implement a better name for env_extra, with deprecation and everything.
- Correct the behavior of HTCondorCluster. I'd say this would be by moving the current behavior of `env_extra` to another kwarg specific to HTCondor, and in the same time implementing `env_extra` as it should, with any of your proposal.

Any thoughts?

riedel:
> * propose and implement a better name for env_extra, with deprecation and everything.

#393 proposes job_script_extra, see also #323

jolange:
> propose and implement a better name for env_extra, with deprecation and everything.
>> #323 proposes job_script_extra

Sounds good.

> I'd say this would be by moving the current behavior of env_extra to **another kwarg** specific to HTCondor

I think that's not even necessary, because you can already use `job_extra` to inject any key-value pair to the submit file.

guillaumeeb:
Wow, good catch for both issues @riedel, I did not remember those. So I'm closing this PR in favor of those two issues we need to fix.

Issue #559, opened 2022-07-18
Replace --nprocs by --nworkers in the generated job script


ikabadzhov:
When I want to create a SLURM cluster and specify `process = X ` in the constructor, by looking at the job script, I see that the command that is run has a flag `--nprocs X`.

That currently does not error. However, referring to https://docs.dask.org/en/stable/deploying-ssh.html#cmdoption-dask-ssh-nworkers `--nprocs` is deprecated.

I tried to pass as an `extra = ['--nworkers X']`, but then from another terminal I saw that my job got kicked from the squeue. And the terminal where I started the program, remains in a "wait state". I created a standalone SLURM script passing both `--nprocs` and `--nworkers`, but then I get: `ERROR - Both --nprocs and --nworkers were specified. Use --nworkers only.`

Request: would it be possible to translate the input of `process = ` in the SLURMCluster constructor to `--nworkers`, but not `--nprocs`.

guillaumeeb:
Hi @ikabadzhov,

Thanks for opening this issue. This is a bit more general than SlurmCluster and affects every implementation. I actually thought I had opened an issue for this already, which is not the case.

This comes from https://github.com/dask/distributed/pull/5641.

So we should update dask-jobqueue to take this into account.

guillaumeeb:
I modified the title of the issue to reflect what should be done. And maybe we should handle the previous kwarg for old Dask version, but I don't think this is really needed.

ikabadzhov:
I took a closer look. I will give it a try now.

PR #560, opened 2022-07-19
Use `--nworkers` in stead of deprecated `--nprocs` in the generated job scripts

ikabadzhov:
Fix https://github.com/dask/dask-jobqueue/issues/559.


guillaumeeb (commented):
Thanks @ikabadzhov.

Unfortunately, test will fail as CI is borken, but this is not a problem to merge. You didn't let any nprocs in the code?

Anyway, I'm thinking about one thing: we should probably update the minimum required version of Dask/Distributed. Did you check when `nworkers` was introduced?

ikabadzhov:
I did not leave any `--nprocs` that is right.

I think the deprecation was introduced here https://github.com/dask/distributed/blame/main/distributed/cli/dask_ssh.py#L52 (6 months ago). I am not sure how to map it to the right version.

dask_jobqueue/tests/test_job.py
```diff
@@ -113,7 +113,7 @@ def test_header_lines_skip():
 
 
 @pytest.mark.asyncio
-async def test_nprocs_scale():
+async def test_nworkers_scale():
```
ikabadzhov, line 116: I also changed the name of this test. That might make the CI unhappy.

guillaumeeb:
Ok, so the new option name was introduced in 2022.02.0, see https://distributed.dask.org/en/stable/changelog.html#v2022-02-0.

So you should also update required distributed version with this one.

As we just released a dask-jobqueue version, I belive it is safe to just do that.

dask_jobqueue/tests/test_job.py
guillaumeeb: CI problems are unfortunately a bit old. I'm hopping to have a look at it in the following weeks.

guillaumeeb:
Also, There is on failure related to your PR in the CI: the check of formatting using black. Could you run black on your branch to format things correctly?

ikabadzhov:
I just applied `black`. I am not sure for this repository to have code that is not formatted, so I kept all changes in a separate commit. Let me know if I should format only my changes.

ikabadzhov:
> Ok, so the new option name was introduced in 2022.02.0, see [distributed.dask.org/en/stable/changelog.html#v2022-02-0](https://distributed.dask.org/en/stable/changelog.html#v2022-02-0).
> 
> So you should also update required distributed version with this one.
> 
> As we just released a dask-jobqueue version, I belive it is safe to just do that.

I just updated the distributed requirement, now it is:
```py
dask>=2.23
distributed>=2022.02.0
```
Is this what you mean? Shouldn't they be the same?

guillaumeeb:
> Is this what you mean? Shouldn't they be the same?

This is what I mean, but yes I think they should be the same.

We may also need to wait for #562 to be in to merge this one, but not sure about it.

jacobtomlinson (commented):
Since switching to CalVer we release `dask` and `distributed` in lock step and only guarantee that those releases work together. Generally pinning will mean that installing a more recent version of `distributed` will upgrade `dask` to the same version, but to be explicit we might want to bump the minimum `dask` version here to match `distributed`.

jacobtomlinson:
_We squash and merge here so please don't squash and force push prematurely, it makes review harder as it's unclear what has changed since the last review._

guillaumeeb (approved):
Anyway, I checked for nprocs occurences, and all are well taken into account. The requirement version of Dask and distributed are OK too, this is good for me.

I still want to advance a bit on #562 before merging.

@jacobtomlinson any other comment.

jacobtomlinson:
LGTM

guillaumeeb:
All green, merging!

PR #563, opened 2022-07-26
Fix behaviour of `env_extra` for HTCondor

### LINKED ISSUE #393: Incorrect description for env_extra for HTCondorCluster

jolange:
Before only environment variables were considered using HTCondor's `Environment =` feature, whereas the parameter description generally says "Other commands to add".

If the `Environment =` feature is needed for some reason, one can still use the generic `job_extra` parameter to set it.

Example:
```python
from dask_jobqueue.htcondor import HTCondorCluster
env_extra = ['cd /some/path', 'source venv/bin/activate']
cluster = HTCondorCluster(cores=1, memory="2GB", disk="4GB", log_directory = 'logs', python='python3',
                          env_extra=env_extra)
print(cluster.job_script())
```
results in this
```
#!/usr/bin/env condor_submit

MY.DaskWorkerName = "htcondor--$F(MY.JobId)--"
RequestCpus = MY.DaskWorkerCores
RequestMemory = floor(MY.DaskWorkerMemory / 1048576)
RequestDisk = floor(MY.DaskWorkerDisk / 1024)
MY.JobId = "$(ClusterId).$(ProcId)"
MY.DaskWorkerCores = 1
MY.DaskWorkerMemory = 2000000000
MY.DaskWorkerDisk = 4000000000
LogDirectory = logs
Output = $(LogDirectory)/worker-$F(MY.JobId).out
Error = $(LogDirectory)/worker-$F(MY.JobId).err
Log = $(LogDirectory)/worker-$(ClusterId).log
Stream_Output = True
Stream_Error = True

Arguments = "-c 'cd /some/path; source venv/bin/activate; python3 -m distributed.cli.dask_worker tcp://<IP>:<PORT> --nthreads 1 --memory-limit 2.00GB --name dummy-name --nanny --death-timeout 60'"
Executable = /bin/sh

Queue
```

fixes #393
related to #323 #556

### DIFF
--- a/dask_jobqueue/tests/test_htcondor.py
+++ b/dask_jobqueue/tests/test_htcondor.py
@@ -27,7 +27,12 @@ def test_job_script():
         processes=2,
         memory="100MB",
         disk="100MB",
-        env_extra=['export LANG="en_US.utf8"', 'export LC_ALL="en_US.utf8"', 'cd /some/path/', 'source venv/bin/activate'],
+        env_extra=[
+            'export LANG="en_US.utf8"',
+            'export LC_ALL="en_US.utf8"',
+            "cd /some/path/",
+            "source venv/bin/activate",
+        ],
         job_extra={"+Extra": "True"},
         submit_command_extra=["-verbose"],
         cancel_command_extra=["-forcex"],
--- a/dask_jobqueue/tests/test_htcondor.py
+++ b/dask_jobqueue/tests/test_htcondor.py
@@ -40,11 +40,10 @@ def test_job_script():
         assert "MY.DaskWorkerDisk = 100000000" in job_script
         assert "MY.DaskWorkerMemory = 100000000" in job_script
         assert 'MY.JobId = "$(ClusterId).$(ProcId)"' in job_script
-        assert "LANG=en_US.utf8" in job_script
-        assert "LC_ALL=en_US.utf8" in job_script
+        assert 'export LANG=""en_US.utf8""' in job_script
+        assert 'export LC_ALL=""en_US.utf8""' in job_script
         assert "cd /some/path/" in job_script
         assert "source venv/bin/activate" in job_script
-        assert "export" not in job_script
         assert "+Extra = True" in job_script
         assert re.search(
             r"condor_submit\s.*-verbose", cluster._dummy_job.submit_command
--- a/dask_jobqueue/tests/test_htcondor.py
+++ b/dask_jobqueue/tests/test_htcondor.py
@@ -27,7 +27,7 @@ def test_job_script():
         processes=2,
         memory="100MB",
         disk="100MB",
-        env_extra=['export LANG="en_US.utf8"', 'export LC_ALL="en_US.utf8"'],
+        env_extra=['export LANG="en_US.utf8"', 'export LC_ALL="en_US.utf8"', 'cd /some/path/', 'source venv/bin/activate'],
         job_extra={"+Extra": "True"},
         submit_command_extra=["-verbose"],
         cancel_command_extra=["-forcex"],
@@ -42,6 +42,8 @@ def test_job_script():
         assert 'MY.JobId = "$(ClusterId).$(ProcId)"' in job_script
         assert "LANG=en_US.utf8" in job_script
         assert "LC_ALL=en_US.utf8" in job_script
+        assert "cd /some/path/" in job_script
+        assert "source venv/bin/activate" in job_script
         assert "export" not in job_script
         assert "+Extra = True" in job_script
         assert re.search(
--- a/dask_jobqueue/htcondor.py
+++ b/dask_jobqueue/htcondor.py
@@ -16,7 +16,6 @@ class HTCondorJob(Job):
 
 %(job_header)s
 
-Environment = "%(quoted_environment)s"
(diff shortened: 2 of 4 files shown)


mivade:
This looks fine to me but I wonder if it would make sense to add some documentation that would include the example in your comment?

jolange:
@mivade sure, always a good idea. Do you have any specific place in mind?
I could add a section to [Advanced tips and tricks](https://jobqueue.dask.org/en/latest/advanced-tips-and-tricks.html). Maybe "Setting up the worker's environment".

dask_jobqueue/tests/test_htcondor.py
```diff
@@ -42,6 +42,8 @@ def test_job_script():
         assert 'MY.JobId = "$(ClusterId).$(ProcId)"' in job_script
         assert "LANG=en_US.utf8" in job_script
         assert "LC_ALL=en_US.utf8" in job_script
+        assert "cd /some/path/" in job_script
+        assert "source venv/bin/activate" in job_script
         assert "export" not in job_script
```
riedel, line 47: a) Why is export stripped from jobscript ?? Why doesn't this fail?
b) Why can't this be checked independent og htcondor in the generic part

riedel (commented):
I think if we make htcondor align with others we should also should move the checks to the generic tests

mivade:
I would put it under the examples section [here](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.HTCondorCluster.html#dask_jobqueue.HTCondorCluster)

dask_jobqueue/tests/test_htcondor.py
guillaumeeb: For b), yes this could now be checked independently of HTCondor, we could add another test for `env_extra` in test_jobqueue_core.py, and apply it with a fixture to all the cluster classes.

But since the behavior just changed for HTCondor, I'm keen to leave also this part of the assertion in the HTCondor cluster.

> Sadly there seem to be problems with the CI in general, right?

Yes unfortunately, I'm trying to advance on this point, but I've not a lot of time to devote to it.

guillaumeeb:
For the docs, reading your comments, I would recommend three places (and put some doc in at least the first one or the last one):

> I could add a section to [Advanced tips and tricks](https://jobqueue.dask.org/en/latest/advanced-tips-and-tricks.html). Maybe "Setting up the worker's environment".

You could add a generic explanation of `env_extra` here, with HTCondor example if you feel confident with this queuing system.

> I would put it under the examples section [here](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.HTCondorCluster.html#dask_jobqueue.HTCondorCluster).

In the docstring, I would put only a simple example, witthout the explanation. I'm not sure this is mandatory to put an example with `env_extra` here.

The 3rd place I was thinking about is the [Example Deployments section](https://jobqueue.dask.org/en/latest/examples.html).

dask_jobqueue/tests/test_htcondor.py
jolange: Alright, I realized anyway with the addition of f678d37b709eb24e927b9a250ac5aa9d1fca7de0 that HTCondor will stay special because of the quoting. At least for the environment variable part (see that commit) you can't generalize this to all Cluster classes.

guillaumeeb: Oh yeah, special quoting problem. So this confirm we need to keep some specific test for HTCondor. 

If you feel like adding a more general test for every queueing system (without export statement 😀), please do.

jolange:
@mivade , @guillaumeeb :
Ok, I added it in all three places: 036b24e7a6ff80b8f0381c49c895e5dd6a9dde3d
Let me know what you think. Unfortunately I will be offline for the rest of the week, but I will continue this next week with high priority.

guillaumeeb (commented):
Congratulations, this is really neat!

I've just got one minor final request. Your doc is perfect, but I don't like duplications. Could you just remove the contentvin examples.rst? The way you described things, it's perfect in advanced tips!

mivade:
@jolange this looks great, thanks! I agree with @guillaumeeb's suggestion that it's fine to only have the lengthier discussion in one place but language wise it's good.

jolange:
@guillaumeeb , @mivade ok, I removed the part from examples.rst.
Once you think it is ready to be merged, I can rebase this and squash the commits into meaningful ones. But looking at the recent history, it looks like you squash all PRs to single commits anyway, right?

guillaumeeb (approved):
LGTM! And yes, we squash on merge, I'll merge this right now.

Issue #568, opened 2022-08-07
HTCondor CI is failing


guillaumeeb:
I've put some time in trying to debug HTCondor CI (see https://github.com/dask/dask-jobqueue/pull/562#issuecomment-1204389580).

If it did work on my laptop, I've not been able to make it works in github Actions, don't know why yet.

Some details:

- I've updated Docker images to run the correct Python version (same as other CI environments),
- I've updated docker-compose setup with a configuration on HTCondor negociator Interval, the time between to scheduling cycle, as the default of 60s made it possible tests fail because of that, or made tests really long.
- I've added some debugging docker commands to try to understand what's happening.

From the last run [here](https://github.com/dask/dask-jobqueue/runs/7697936005?check_suite_focus=true), we can see from the Cleanup sections and `condor_history` command output that the HTCondor queuing system is working. The problem seems to come from the worker jobs which complete really fast,  Dask workers never connect to Scheduler.

I guess we would need to see the jobs stdout/stderr to debug further the problem.

cc @riedel @mivade @jolange.

jolange:
Thanks  @guillaumeeb ! Yes, this looks like the worker jobs are failing directly.
The command is starting with a semicolon now:
```
 Arguments = "-c '; /opt/anaconda/bin/python -m distributed.cli.dask_worker tcp://172.18.0.4:33755 --nthreads 1 --memory-limit 100.00MiB --name HTCondorCluster-1 --nanny --death-timeout 60'"
```
I certainly broke this with my changes, I guess because `env_extra` does not stay `None` due to [this line](https://github.com/dask/dask-jobqueue/blob/50a6f393b17b273d15f1bfc958019a8a0fcc6fa1/dask_jobqueue/htcondor.py#L67).
I'm not sure if this would cause this to fail, but I'll fix it first before looking further into this.

Otherwise, seeing stderr would help a lot, probably -- I remember that I had to do some manual downgrades of click for some python versions. But I tried a number of setups, so I don't remember really...

guillaumeeb:
Closed by #570.

PR #570, opened 2022-08-08
Fix command template for empty `env_extra` in HTCondor

jolange:
Due to the config read with `default=[]`, `env_extra` will not stay `None` but become an empty list. This resulted in a command template starting with a semicolon.

By first merging `env_extra` and `_command_template` to a single list, this is avoided.

Possibly related to #568, cc @guillaumeeb


jolange:
Now I am confused:
- PBS failed in the container setup step.
- Slurm failed. `<Client: No scheduler connected>` sounds suspicious.
- HTCondor did not run?

guillaumeeb:
@jolange I think you're on something!! It looks like 100MiB is not enough for running a Dask Worker!

However, I believe the default Condor setup is only 1GB available on each condor worker node, so you should use a number lower than that. Maybe try with 500MiB to be safe?

jolange:
Ah, thanks, I was just trying to find out what the available memory could be. With 2GiB the job did not start to run, so that seemed too much ;-) I'm trying with 500GiB now.

jolange:
With 500GiB it worked without the warning in stderr and I also had a successful [CI run for HTCondor](https://github.com/dask/dask-jobqueue/runs/7729441766?check_suite_focus=true).
Still, the last run resultet in a timeout again, but that also happens for "CI / build (none)" for the LocalCluster from time to time.

guillaumeeb:
Just tried a complement fix on your branch, hope it's okay. The second test was probably fragile too because it also used only 100MiB for worker jobs. If that test fails and workers are note cleaned up, other test will fail.

guillaumeeb:
Okay, HTCondor CI is green, nice :clap:. Thanks a lot @jolange! 

I will just make another commit here to re-add some of the debug tricks you used, it could be nice later on to have worker logs again!

guillaumeeb (approved):
All is green, outputs is detailed, again, thanks a lot for the work here @jolange!

jolange:
Nice, thanks!

PR #572, opened 2022-08-10
use the stored env_extra parameter from the base class for htcondor

jolange:
While working on #323, I realized that it would be much easier to manage the deprecation warning if the `env_extra` parameter would only be handled in one place (including the config read). Apart from that, I think it is also cleaner, and it was stored already anyway (as a joined string, though).

Anyway, I wanted to first check in this separate PR if you see any reason not to do this, instead of hiding this change in the renaming+deprecation.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -245,7 +245,7 @@ class Job(ProcessInterface, abc.ABC):
 
         self.shebang = shebang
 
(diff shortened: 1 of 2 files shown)


mivade (approved):
This makes sense to me.

guillaumeeb (approved):
I see no reason not to do this!

PR #574, opened 2022-08-11
CI failure fixe: Increasing memory limit of workers in test_security

guillaumeeb:
Just looking if increasing this limit in the failing test changes something


guillaumeeb:
3 success on 3 tries, merging!

PR #576, opened 2022-08-12
Rename `extra`

jolange:
Renamed `extra` to `worker_extra_args` and added a FutureWarning for `extra`.

If `worker_extra_args` is empty, `extra` will still be respected (see new tests).

Resolves https://github.com/dask/dask-jobqueue/issues/323


dask_jobqueue/jobqueue.yaml
```diff
@@ -11,7 +11,8 @@ jobqueue:
     death-timeout: 60           # Number of seconds to wait if a worker can not find a scheduler
     local-directory: null       # Location of fast local storage like /scratch or $TMPDIR
     shared-temp-directory: null       # Shared directory currently used to dump temporary security objects for workers
-    extra: []
+    extra: null
```
jacobtomlinson, line 14: Nit: Could you add a descriptive comment to each of these like the above config options?

Same for all below.

jacobtomlinson (approved):
This looks great many thanks! One small comment but otherwise looks good to me.

dask_jobqueue/jobqueue.yaml
jolange: Sure, done!

guillaumeeb (approved):
Perfect!

PR #577, opened 2022-08-19
rename `job_extra`

jolange:
to `job_extra_directives`, see #323.

If `job_extra_directives` is empty,  `job_extra` will still be respected.

The warning code is duplicated for different *Cluster classes now. I thought about putting it in the base class, but that does not know this parameter (almost all implentations define it, but not all) and it is not treated exactly the same everywhere.
So I think the repetition is ok for this temporary code.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -235,10 +235,10 @@ class Job(ProcessInterface, abc.ABC):
                 worker_extra_args = extra
 
         if job_extra is None:
-            job_extra = dask.config.get("jobqueue.%s.job-extra" % self.config_name, {})
+            job_extra = dask.config.get("jobqueue.%s.job-extra" % self.config_name, [])
         if job_extra_directives is None:
             job_extra_directives = dask.config.get(
-                "jobqueue.%s.job-extra-directives" % self.config_name, {}
+                "jobqueue.%s.job-extra-directives" % self.config_name, []
             )
         if job_extra is not None:
             warn = (
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -107,6 +107,11 @@ class Job(ProcessInterface, abc.ABC):
     Parameters
     ----------
     {job_parameters}
+    job_extra : list or dict
+        Deprecated: use ``job_extra_directives`` instead. This parameter will be removed in a future version.
+    job_extra_directives : list or dict
+        Unused in this base class:
+        List or dict of other options for the queueing system. See derived classes for specific descriptions.
 
     Attributes
     ----------
@@ -159,6 +164,8 @@ class Job(ProcessInterface, abc.ABC):
         local_directory=None,
         extra=None,
         worker_extra_args=None,
+        job_extra=None,
+        job_extra_directives=None,
         env_extra=None,
         job_script_prologue=None,
         header_skip=None,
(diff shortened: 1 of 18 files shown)


guillaumeeb:
Thanks @jolange for continuing work here, I really appreciate it.

> The warning code is duplicated for different *Cluster classes now. I thought about putting it in the base class, but that does not know this parameter (almost all implentations define it, but not all) and it is not treated exactly the same everywhere

With a quick glance, I'm not sure why it is duplicated and not in the base class. Do you think something prevent to declare a private attribute in the main class, putting the warning in there, and just keep special handling in defferent implementations?

jolange:
I did not check in detail before, which *Cluster classes do not implement it, but it now I see that it's actually only Moab and that this basically is PBS. So all of them use the parameter, really.

The only problem is that the parameter is supposed to be a dict for HTCondor, but a list for every other one.

guillaumeeb:
> The only problem is that the parameter is supposed to be a dict for HTCondor, but a list for every other one.

I guess this is a problem only for the doc string or warning message? Maybe we could indicate list or dict as type? 

I'm not sure if we should duplicate code just for this? Or maybe we should just do something for the HTCondor specific case?

jolange:
Only for the doc string, actually. Indicating "dict or list" would be ok, I think, but the doc strings for the *Cluster classes are rather different (and rightfully so, I think):
```console
$ ack "job_extra_directives : " dask_jobqueue/*.py -A2
dask_jobqueue/htcondor.py
245:    job_extra_directives : dict
246-        Extra submit file attributes for the job as key-value pairs.
247-        They will be inserted as ``key = value``.

dask_jobqueue/lsf.py
211:    job_extra_directives : list
212-        List of other LSF options, for example -u. Each option will be
213-        prepended with the #LSF prefix.

dask_jobqueue/oar.py
134:    job_extra_directives : list
135-        List of other OAR options, for example `-t besteffort`. Each option will be prepended with the #OAR prefix.
136-

dask_jobqueue/pbs.py
143:    job_extra_directives : list
144-        List of other PBS options. Each option will be prepended with the #PBS prefix.
145-

dask_jobqueue/sge.py
117:    job_extra_directives : list
118-        List of other SGE options, for example -w e. Each option will be
119-        prepended with the #$ prefix.

dask_jobqueue/slurm.py
148:    job_extra_directives : list
149-        List of other Slurm options, for example -j oe. Each option will be prepended with the #SBATCH prefix.
150-

```

What do you think about making it a parameter of the base class and store it as member variable, but keeping the doc string in the different base classes? Maybe adding a comment to the parameter in the base class where the doc string(s) can be found?

guillaumeeb:
Sounds good. We should put a generic doc string in the base class, and some more complete description in each cluster implementation.

dask_jobqueue/core.py
```diff
@@ -226,6 +233,26 @@ def __init__(
             warnings.warn(warn, FutureWarning)
             if not worker_extra_args:
                 worker_extra_args = extra
+
+        if job_extra is None:
+            job_extra = dask.config.get("jobqueue.%s.job-extra" % self.config_name, {})
+        if job_extra_directives is None:
+            job_extra_directives = dask.config.get(
+                "jobqueue.%s.job-extra-directives" % self.config_name, {}
```
guillaumeeb, line 241: We should probably put None as the default, or no default, for both job_extra and job_extra_directives?

dask_jobqueue/jobqueue.yaml
```diff
@@ -22,7 +22,8 @@ jobqueue:
     env-extra: null
     job-script-prologue: []
     resource-spec: null
-    job-extra: []
+    job-extra: null
+    job-extra-directives: []
```
guillaumeeb, line 26: We should put null on every default configurations too, don't you think?

guillaumeeb (commented):
This looks really good, I've just got a comment on default values, both in the code and in the config file.

dask_jobqueue/core.py
jolange: For other options (including the old `job_extra`), the default (in the code and *at least* in the yaml-config) seems to be an empty list or dict such that you always have something iterable to do things like
```python
header_lines.extend(["#SBATCH %s" % arg for arg in self.job_extra_directives])
```
without checking if it is empty/None first. (This was at least my interpretation of this general pattern here.)

You are probably right, at this place one could drop the default, but I would certainly keep it in the config, then. Having an empty dict here should be fine, though, for all cases -- up to you to decide.

guillaumeeb: I would at least replace the dict by a list, as most implementations use a list here. Would that work for HTCondor?

dask_jobqueue/jobqueue.yaml
guillaumeeb: Based on remark above, ok to keep empty default in configuration files.

dask_jobqueue/core.py
jolange: I chose dict, because I considered that more general: you can iterate over a dict, i.e., over the values, but you can't use `dict.update()` with a list as argument.
But you have a point here and in the case of an empty list, it also works for HTCondor. I just pushed the change.

guillaumeeb:
Thanks a lot @jolange for sticking with that!!

PR #582, opened 2022-08-30
Make CIs obey environment.yaml and ignoring useless steps

### LINKED ISSUE #324: Add a CONTRIBUTING.md with how to run the tests
This would be very useful, I am just jotting down things quickly so some things may not be completely accurate. If someone wants to do it, that would be more than welcome!

I think for now adding a CONTRIBUTING.md and possibly add a link from the README.md is good enough. An alternative is to put it somewhere in the doc but this is a bit more effort.

### tests who don't need a cluster
```
pytest dask_jobqueue
```

### tests who need a cluster
They have `@pytest.mark.env("scheduler-name")` before the test function. With a cluster (`cluster-name` needs to match the name in the `pytest.mark.env`):
```
pytest dask_jobqueue -E <cluster-name>
```

e.g. on my `SGE` cluster I do:
```
pytest dask_jobqueue -E sge
```

### running tests locally with our docker-compose setup
For some clusters (SGE, PBS, SLURM) at the time of writing, we have a `docker-compose` setup to run the tests locally.

```
cd ci/sge
./start-sge.sh
```

Sometimes there is a problem (never figured out when exactly) and you want to regenerate the docker images from scratch (there may be a better way to do this to be honest, because my feeling is that you want only to rerun the part of the conda install in the docker image not the full thing):
```
cd ci/sge
docker-compose build --no-cache
./start-sge.sh
```

### LINKED ISSUE #539: Make CIs obey environment.yaml
Actually ` environment.yaml` is installed outside the actual docker images and mostly even does not make sense.

- [ ] use the yaml file inside the dockerfile for building the CI images
- [ ] no need to install dependencies outside the docker for cluster CIs
- [x] use pip versions of dask and distributed since the conda versions are outdated
- [x] make htcondor use conda

At the moment one needs to edit tons of files if you actually want to make sure a test dependency is working

guillaumeeb:
Fixes #539. 

Fixes #324.

### DIFF
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -26,6 +26,7 @@ jobs:
         shell: bash -l {0}
         run: |
           cd ./ci/${{ matrix.jobqueue }}
+          cp ../environment.yml environment.yml
           docker-compose build
       - name: List images
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,8 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v2
 
-      - name: Setup Conda Environment
+      - name: Setup Empty Conda Environment with Mamba
+        if: matrix.jobqueue == 'none'
         uses: conda-incubator/setup-miniconda@v2
         with:
           channels: conda-forge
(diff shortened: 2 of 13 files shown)


guillaumeeb:
I'm going to merge in order to check also the docker images standalone builds.

Feel free to comment or open issues if you think I've broken something.

PR #583, opened 2022-08-30
Do not build Docker images when testing CI, update docs and workflows

### LINKED ISSUE #566: Check if CI really don't build Docker image if not needed
Since https://github.com/dask/dask-jobqueue/pull/545, we have added in most of the CI runs a `docker compose build` command that was not there (by choice) to use pre-backed Docker images.

Without a `docker compose build`, as we had updated Python version, CI would fail until the Docker images were built.

But I'm wondering with that addition if we've not made a step back, and now we'll build Docker images every time, even if the Dockerfile has not changed and we could use a pre built one.

For instance, you can see the command here: https://github.com/dask/dask-jobqueue/blob/main/ci/slurm.sh#L10.

guillaumeeb:
Fixes #566.

### DIFF
--- a/.github/workflows/build-docker-images.yaml
+++ b/.github/workflows/build-docker-images.yaml
@@ -3,6 +3,7 @@ name: docker-images-build
 on:
   push:
     branches: "main"
+  pull_request:
   schedule:
     - cron: "0 0 * * *" # Daily “At 00:00”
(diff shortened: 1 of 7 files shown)


guillaumeeb:
This looks good, I'm going to merge this one too. Again, if someone notice something wrong, please say it!

PR #586, opened 2022-09-05
Change project to account where appropriate. Update docs

### LINKED ISSUE #272: PBSCluster hard codes account string for project
Currently `PBSCluster` accepts a project code as an argument and uses this with the account string, `-A`, option

https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/pbs.py#L96

PBS Pro has account string (`-A`) and project (`-P`) options:

https://github.com/PBSPro/pbspro/blob/master/src/cmds/qsub.c#L892-L898

For my local HPC setup the `-P` option is used for project based accounting. 

It would be preferable to have the ability to pass either `account` or `project` as options for `PBSCluster` and use the appropriate flag. This would break existing configs that assume the `-A` flag is used.

One possible solutions would be to allow the flags associated with options to be specified in a `dask` config file.

Would the maintainers be happy to accept a PR with this implemented?

guillaumeeb:
Fixes #272.

Currently dask-jobqueue uses incorrect terms and kwarg name for Accounting string. PBS and Slurm implementations are declaring an accounting string, but the kwarg was named `project`. This PR renames the kwarg to `account`, as project should be use with `-P` PBS or Slurm option.

It also fixes wrong comment about `project` kwarg for other job schedulers.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -285,7 +285,7 @@ class Job(ProcessInterface, abc.ABC):
         if header_skip is not None:
             warn = (
                 "header_skip has been renamed to job_directives_skip. "
-                "You are still using it (even if only set to (); please also check config files). "
+                "You are still using it (even if only set to []; please also check config files). "
                 "If you did not set job_directives_skip yet, header_skip will be respected for now, "
                 "but it will be removed in a future release. "
                 "If you already set job_directives_skip, header_skip is ignored and you can remove it."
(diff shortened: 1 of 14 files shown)


dask_jobqueue/core.py
```diff
@@ -285,7 +285,7 @@ def __init__(
         if header_skip is not None:
             warn = (
                 "header_skip has been renamed to job_directives_skip. "
-                "You are still using it (even if only set to (); please also check config files). "
+                "You are still using it (even if only set to []; please also check config files). "
```
guillaumeeb, line 288: This is an unrelated fix.

PR #588, opened 2022-09-12
Clean up inconsistencies in the directory where the tests are run from

### LINKED ISSUE #374: CI: Clean up inconsistencies in the directory where the tests are run from
Left-over from #307. From https://github.com/dask/dask-jobqueue/pull/307#issuecomment-531300128:

> it would be great if we could find a way not to have to cd in a different folder (jobqueue scheduler-specific, mostly PBS and SLURM). In an ideal world the /dask-jobqueue volumes would be mounted on the worker images I think but also pbsuser would have the right to write into /dask-jobqueue. Not sure how easy it is to do the second part in docker

```
❯ ag 'docker.+pytest' ci/
ci/slurm.sh
21:    docker exec -it slurmctld /bin/bash -c "pytest /dask-jobqueue/dask_jobqueue --verbose -E slurm -s"

ci/pbs.sh
22:    docker exec -it -u pbsuser pbs_master /bin/bash -c "cd; pytest /dask-jobqueue/dask_jobqueue --verbose -s -E pbs"

ci/sge.sh
20:    docker exec -it sge_master /bin/bash -c "cd /dask-jobqueue; pytest dask_jobqueue --verbose -s -E sge"
```

* SLURM: we stay in `/` (i.e. no `cd`). Reason from https://github.com/dask/dask-jobqueue/pull/307#issuecomment-526645923
> if you run sbatch /tmp/test.sh from a folder that does not exist in the worker node (e.g. /dask-jobqueue), the job terminates instantly. Somehow this seems related to the SLURM logging mechanism because if I set -e and -o to a folder that exists on the worker node the job runs fine. Maybe fixable in the docker images of the SLURM slaves.
* PBS: we run as `pbsuser` so we have to cd somewhere where we can write (logs + local dask files). Here we `cd` into the `pbsuser` HOME directory. Reminder: you need to be root to write in `/dask-jobqueue`. Maybe fixable inside docker.
* SGE: this was the default way of running the tests before #307.

guillaumeeb:
Fixes #374.

The idea to fix the issue is to alway cd in the submitting user $HOME directory, even if the submitting user is root. I think this is the correct approach as this is closer to a real case setup. At one point, we might want to add a non root submit user to other CI. HTCondor already has one like PBS.

### DIFF
--- a/ci/sge.sh
+++ b/ci/sge.sh
@@ -20,7 +20,7 @@ function jobqueue_install {
 }
 
 function jobqueue_script {
-    docker exec sge_master /bin/bash -c "cd /dask-jobqueue; pytest dask_jobqueue --verbose -s -E sge"
+    docker exec sge_master /bin/bash -c "cd; pytest /dask-jobqueue/dask_jobqueue --verbose -s -E sge"
 }
 
 function jobqueue_after_script {
--- a/ci/slurm.sh
+++ b/ci/slurm.sh
@@ -29,7 +29,7 @@ function jobqueue_install {
 }
 
 function jobqueue_script {
-    docker exec slurmctld /bin/bash -c "pytest /dask-jobqueue/dask_jobqueue --verbose -E slurm -s"
+    docker exec slurmctld /bin/bash -c "cd; pytest /dask-jobqueue/dask_jobqueue --verbose -E slurm -s"
 }
 
 function jobqueue_after_script {


guillaumeeb:
This ended up (again) wit far more changes than I thought at first.

The reason being, on top of launching test from the corresponding docker user home, I also added a shared space in all CIs (but LocalCluster) to be able to launch some tests (for the time being, Security related tests). 

So now, except the fact that some CIs are running with non privileged users inside Docker (PBS and HTCondor) and others are not, the setup from each CI is similar.

PR #590, opened 2022-10-02
Prevent worker_extra_args to adds up extra arguments

### LINKED ISSUE #589: Cluster keeps appending "interface" flag to job script
When I run the following code snippet:

```python

cluster = SLURMCluster(memory="100GB", cores=40, interface="ib0")

for _ in range(20):
    _ = cluster.job_script()
    
print(cluster.job_script())
```

I get the following output:
```
#!/usr/bin/env bash

#SBATCH -J dask-worker
#SBATCH -n 1
#SBATCH --cpus-per-task=40
#SBATCH --mem=94G
#SBATCH -t 00:30:00

/path/to/python -m distributed.cli.dask_worker tcp://xx.xx.xx.xx:pppp --nthreads 5 --nworkers 8 --memory-limit 11.64GiB --name dummy-name --nanny --death-timeout 60 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0 --interface ib0
```
It seems everytime the job script method is called, the interface flag is appended to it. I am running `dask-jobqueue==0.8.0`  off the `conda-forge` channel. (Python 3.10 if relevant)

You can replicate this with `PBSCluster` as well. The issue seems to be in how `worker_extra_args` are handled in the `__init__()` method of `dask_jobqueue.core.Job`

guillaumeeb:
Fixes #589, but I'd like to understand better how this bug was introduced.

Maybe we should also add a test, not mandatory here.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -237,6 +237,10 @@ class Job(ProcessInterface, abc.ABC):
             warnings.warn(warn, FutureWarning)
             if not worker_extra_args:
                 worker_extra_args = extra
+        if worker_extra_args is not None:
+            # Copy worker_extra_args in order to not modify the object comming from dask.config
+            # when we'll assign new args like interface to it. See #589.
+            worker_extra_args = worker_extra_args.copy()
 
         if job_extra is None:
             job_extra = dask.config.get("jobqueue.%s.job-extra" % self.config_name, [])


theXYZT:
I think it's because in `jobqueue.yaml` the default `extra` value is `null`, but the default value for `worker-extra-args` is `[]` which is mutable.

guillaumeeb:
Superseded by #591

Issue #594, opened 2022-10-28
OARCluster implementation does not let OAR take into account the memory parameter


ychiat35:
The current implementation of OARCluster does not let OAR take into account the memory parameter.

For example, if 256 GB are asked for Dask workers as memory, OAR is not aware about this memory request and thus the resources selected by OAR to compute the job might not have enough memory.

Consequently, specifying memory in OARCluster is counter-intuitive. As a Dask-Jobqueue user, we expect this parameter to be taken into account but in practice it is not.

guillaumeeb:
Never used OAR, but it looks you're right! The only way to specify job memory is to use `resource_spec` (https://github.com/dask/dask-jobqueue/blob/main/dask_jobqueue/oar.py#L25). 

> Consequently, specifying memory in OARCluster is counter-intuitive. As a Dask-Jobqueue user, we expect this parameter to be taken into account but in practice it is not.

:+1:

ychiat35:
Maybe @lesteve could help reviewing this pull request #595 ? Thanks in advance!

guillaumeeb:
I don't think @lesteve has time to look at dask-jobqueue for the time being. But I'll have a look to your PR as soon as I can (following days I hope, but may be weeks, sorry).

guillaumeeb:
Fixed in #595.

PR #595, opened 2022-10-28
Extend OARCluster implementation to let OAR take into account the memory parameter

ychiat35:
Following the issue #594 , we propose an extension of the existing OARCluster implementation to let OAR take into account the memory parameter.

The OAR scheduler does not deal with memory internally. Indeed, by default it is not possible to indicate to OAR to reserve a specific amount if memory on the wanted computing resources (e.g., one core with 256 GB memory). However, it is possible to leverage from OAR resource properties to ensure that the wanted resources have at least the wanted amount of memory.

Since the OAR property names are not standardized by OAR, their names might differ from one cluster to another. Consequently, we introduce a new parameter in OARCluster class: `oar_mem_core_property_name`. It lets users specify the name of the memory property of their own OAR cluster. This property will be used by adding a new `#OAR -p` line to the OAR submission. If the parameter is not used or set to None, our modification does not modify the current behavior of the OARCluster class, but users will be warned that the memory parameter will not be taken into account by OAR.

### DIFF
--- a/dask_jobqueue/oar.py
+++ b/dask_jobqueue/oar.py
@@ -27,7 +27,7 @@ class OARJob(Job):
         project=None,
         resource_spec=None,
         walltime=None,
-        oar_mem_core_property_name=None,
+        mem_core_property_name=None,
         config_name=None,
         **base_class_kwargs
     ):
@@ -81,36 +81,33 @@ class OARJob(Job):
         header_lines.extend(["#OAR %s" % arg for arg in self.job_extra_directives])
 
         # Memory
-        memory = self.worker_memory
-        if memory is not None:
-            if oar_mem_core_property_name is None:
-                oar_mem_core_property_name = dask.config.get(
-                    "jobqueue.%s.oar-mem-core-property-name" % self.config_name
+        if self.worker_memory is not None:
+            if mem_core_property_name is None:
+                mem_core_property_name = dask.config.get(
+                    "jobqueue.%s.mem-core-property-name" % self.config_name
                 )
-            if oar_mem_core_property_name is None:
-                warn = "oar_mem_core_property_name is not set. Thus OAR will not take the memory parameter into account"
+            if mem_core_property_name is None:
+                warn = (
+                    "OAR Job memory reserved resources will not be set according to Dask Worker memory limit, "
+                    "which can cause crashes."
+                )
                 warnings.warn(warn, category=UserWarning)
             else:
                 # OAR expects MiB as memory unit
                 oar_memory = int(
                     parse_bytes(self.worker_memory / self.worker_cores) / 2**20
                 )
-                # OAR needs to have the properties on a single line, with SQL syntaxe
+                # OAR needs to have the properties on a single line, with SQL syntax
                 # If there are several "#OAR -p" lines, only the last one will be taken into account by OAR
                 last_job_property = return_last_job_property(self.job_extra_directives)
                 if last_job_property is not None:
                     header_lines.append(
-                        "#OAR -p "
-                        + '"'
-                        + last_job_property
-                        + " AND "
-                        + oar_mem_core_property_name
-                        + ">=%s" % oar_memory
-                        + '"'
+                        "#OAR -p '%s AND %s>=%s'"
+                        % (last_job_property, mem_core_property_name, oar_memory)
                     )
                 else:
                     header_lines.append(
-                        "#OAR -p " + oar_mem_core_property_name + ">=%s" % oar_memory
+                        "#OAR -p %s>=%s" % (mem_core_property_name, oar_memory)
                     )
 
         self.job_header = "\n".join(header_lines)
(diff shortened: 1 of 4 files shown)


dask_jobqueue/tests/test_oar.py
```diff
@@ -28,28 +28,55 @@ def test_header():
         assert "#OAR -q regular" in cluster.job_header
         assert "#OAR -t besteffort" in cluster.job_header
 
-    with OARCluster(cores=4, memory="8GB") as cluster:
```
guillaumeeb, line 31: I'm not sure if you left some tests checking the generated job_script without using `oar_mem_core_property_name` kwarg?

dask_jobqueue/tests/test_oar.py
```diff
@@ -28,28 +28,55 @@ def test_header():
...
+    with OARCluster(
+        walltime="00:02:00",
+        processes=4,
+        cores=8,
+        memory="28GB",
+        oar_mem_core_property_name="mem_core",
+    ) as cluster:
+        assert "#OAR -n dask-worker" in cluster.job_header
+        assert "#OAR -l /nodes=1/core=8,walltime=00:02:00" in cluster.job_header
+        assert "#OAR -p mem_core>=3337" in cluster.job_header
```
guillaumeeb, line 49: Be careful with computation like this, we've already seen errors with memory string parsing. At least, I would suggest to sue `GiB` syntax in the `OARCluster` constructor, so this can be more easily divided into MiB.

guillaumeeb (commented):
I left a few comments on details, but in the general principle this looks good to me! I don't see any blocker to have this in, especially if it can ease the use of dask-jobqueue on OAR based clusters!

dask_jobqueue/oar.py
```diff
@@ -76,6 +80,39 @@ def __init__(
...
+                oar_mem_core_property_name = dask.config.get(
+                    "jobqueue.%s.oar-mem-core-property-name" % self.config_name
+                )
+            if oar_mem_core_property_name is None:
+                warn = "oar_mem_core_property_name is not set. Thus OAR will not take the memory parameter into account"
+                warnings.warn(warn, category=UserWarning)
+            else:
+                # OAR expects MiB as memory unit
+                oar_memory = int(
+                    parse_bytes(self.worker_memory / self.worker_cores) / 2**20
```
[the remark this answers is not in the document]
ychiat35: In fact, OAR does not handle memory internally. So there is no computation made by OAR to calculate the total amount of memory.

To reserve resources with the wanted memory, we use OAR properties. It ensures that the resources given by OAR have at least the wanted memory. Note that the PR does not change any configuration made by OAR on the node. It is just filtering on the available resources to match the required memory.

As OAR configures by default the usable memory of a job:
- either the total amount of memory of the node (if all the cores of the node are reserved by the job)
- or a proportion of the total amount of memory (if only a subset of the cores are reserved by the job)

These two configurations of OAR memory correspond to memory_per_node and memory_per_core properties. But the first one can be relevant only if we reserve all the cores of one node.

We cannot filter resources using the total amount of memory on OAR. Instead, to ensure that usable memory given by OAR matches the required memory, we have to use the memory_per_core property and do the maths on Dask-jobqueue side..

ychiat35:
Thanks a lot for the your quick review! 
Updates following your comments are just committed. About your "worker_memory/worker_cores calculation" question, you can find above my explanations. Hope that it's clear for you and do not hesitate if you have other comments :)

guillaumeeb (approved):
Thanks @ychiat35, this looks good to me!

lesteve:
A bit late to the party:

lesteve:
A bit late to the party, sorry! Here are the few comments I had, this would be very welcome if they could be tackled in a further PR:

- use `memory_per_core_property` rather than `mem_core_property` to be more consistent with the `memory` parameter and also to be more explicit
- test the case where there is already a property defined and you need to combine both properties (I don't think it is tested but maybe I am wrong)
- there is no way to silence the warning right now, right? So maybe have something like `memory_per_core_property='not_applicable'` for the clusters where there is no such property. If you tell me this rarely happens in practice then maybe it is fine to forget about this item.

More questions (not related to this PR in particular but to OAR support more generally):
- where you running the tests on a OAR cluster?
- it would be great to add a CI for OAR, if there was a single docker image with everything that is needed (similarly to the SLURM docker image we are using for the CI) or maybe with oar-docker if that does not exist

ychiat35:
Thanks @lesteve for your comments! Please find my commit in #598 

- I renamed the memory_per_core_property_name parameter to be more explicit.

- I added also possibility to silence the warning for clusters where the property does not exist. One unit test is also added at the end of test_oar.py to include this case.

- About the case where there is already a property defined, the unit test exists normally. At the end of the test_header() part, I tested the 'memory_per_core_property' combined with 'the gpu_count=1' request. Does it correspond to your expectation?

By the way, the OAR clusters that I use are Grid5000 and igrida :)

lesteve:
> About the case where there is already a property defined, the unit test exists normally. 

Yep I missed it somehow, thanks!

PR #598, opened 2022-11-17
[OAR] Rename memory per core property name and other tweaks

ychiat35:
update following @lesteve 's comments on the PR #595 already merged

### DIFF
--- a/dask_jobqueue/oar.py
+++ b/dask_jobqueue/oar.py
@@ -27,7 +27,7 @@ class OARJob(Job):
         project=None,
         resource_spec=None,
         walltime=None,
-        mem_core_property_name=None,
+        memory_per_core_property_name=None,
         config_name=None,
         **base_class_kwargs
     ):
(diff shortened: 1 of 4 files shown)


lesteve:
Yep I'll try to have a look, thanks a lot @ychiat35 for the follow-up PR! 

Probably not this week, but next week should be doable!

dask_jobqueue/oar.py
```diff
@@ -82,33 +81,41 @@ def __init__(
...
-                # If there are several "#OAR -p" lines, only the last one will be taken into account by OAR
-                last_job_property = return_last_job_property(self.job_extra_directives)
-                if last_job_property is not None:
-                    header_lines.append(
-                        "#OAR -p '%s AND %s>=%s'"
-                        % (last_job_property, mem_core_property_name, oar_memory)
+                if memory_per_core_property_name != "not_applicable":
+                    # OAR expects MiB as memory unit
+                    oar_memory = math.ceil(
+                        self.worker_memory / (self.worker_cores * 2**20)
```
lesteve, line 98: I *think* self.worker_memory is already an int, so we don't need `parse_bytes` here, which is why I removed it.

dask_jobqueue/oar.py
```diff
@@ -82,33 +81,41 @@ def __init__(
...
-                # OAR needs to have the properties on a single line, with SQL syntax
-                # If there are several "#OAR -p" lines, only the last one will be taken into account by OAR
-                last_job_property = return_last_job_property(self.job_extra_directives)
-                if last_job_property is not None:
-                    header_lines.append(
-                        "#OAR -p '%s AND %s>=%s'"
-                        % (last_job_property, mem_core_property_name, oar_memory)
+                if memory_per_core_property_name != "not_applicable":
+                    # OAR expects MiB as memory unit
+                    oar_memory = math.ceil(
```
lesteve, line 97: I use `math.ceil` here to be sure that the rounding is done towards bigger values, so that the allocated nodes has a bit more memory that the Dask worker memory limit

lesteve (commented):
This looks good to me.

I pushed a commit trying to improve the documentation a bit, let me know if you think this can be improved.

I also left some inline comments to explain some of the other changes I made, any comments, let me know!

dask_jobqueue/oar.py
ychiat35: You are right! Maybe I confused it with "memory" parameter.

ychiat35:
Hello @guillaumeeb and @lesteve,

We are wondering if you are planning to make a release (including the OAR modifs) soon?

Thanks a lot and best wishes for 2023 :)

lesteve:
Personally I am not planning on doing a release in the near future. Looking at it there are only the OAR changes since the last release (0.8.1).

Have you considered installing from source for now, for example with:

```
pip install git+https://github.com/dask/dask-jobqueue
```

While I am at it, I noticed there were no changelog added for the OAR PRs in https://github.com/dask/dask-jobqueue/blob/main/docs/source/changelog.rst. If you have the time to do a PR, it would be super welcome!

guillaumeeb:
As always, I was planning on working on dask-jobqueue more than I did in the past months... As @lesteve said, there is only this OAR change, so I'm not considering making a release for now, but we can discuss this. It would be better if you could install from source though.

And if you want a release soon, you could help on other subjects :grin: https://github.com/dask/dask-jobqueue/milestone/6.

Happy new year!

lesteve:
@guillaumeeb do you have any objections if I make a release at the end of next week (say around June 8th) ?

I was asked about this by Inria colleagues that want to document how to use dask-jobqueue on their OAR clusters. Having a release with the OAR improvements would make it easier for them!

guillaumeeb:
Not at all, please go @lesteve, and sorry again for the lack of commitment here, always complicated!

lesteve:
@guillaumeeb FYI I just did a 0.8.2 release. Any unexpected issues, let me know!

guillaumeeb:
Thanks!

Issue #543, opened 2022-01-31
Proposal: `NamedCluster`


alisterburt:
Hi! First, huge thanks for the work in this repo - it has made working on clusters so easy.

Currently, I must know in advance which type of cluster a user will be running on to decide which `Cluster` object (slurm/sge etc) to instantiate, the appropriate values are then filled from the config file.

Would you consider a PR adding a `NamedCluster` class? The idea is quite simple

- add a `jobqueue-named-clusters` section to the existing cluster config specification
- make `scheduler` a key at the same level as the rest of the entries in existing cluster config

this would look something like

```yml
jobqueue-named:
  my-awesome-cluster:
    scheduler: 'slurm'
    name: dask-worker
    cores: 36                   # Total number of cores per job
    memory: '109 GB'            # Total amount of memory per job
    processes: 9                # Number of Python processes per job
    interface: ib0              # Network interface to use like eth0 or ib0

    queue: regular
    walltime: '00:30:00'
    resource-spec: select=1:ncpus=36:mem=109GB
```

`NamedCluster` would then dispatch to the appropriate `Cluster` class based on the value found in 'scheduler'

The use case I envision is providing daskified programs for non-python users to run in HPC environments - I would be able to use `NamedCluster` liberally in my programs across projects and leave all cluster config to the end user in a config file

guillaumeeb:
Hi @alisterburt,

> Hi! First, huge thanks for the work in this repo - it has made working on clusters so easy.

Thanks, this is always good to have such feedback 😄 

Considering your proposal, this sounds good to me! We should probably discuss the details a bit.

I think we should try if going this way to keep the config file consistent. Perhaps we could just add new entries at the same level as pbs, slurm, etc. ? Those would be kind of default configs or templates. Also, I think we should avoid the `scheduler` term alone that could confuse Dask users, maybe at least `job-scheduler` or `job-scheduling-system`. 

We could have something in the config file like:

```yaml
jobqueue:
  pbs:
    job-scheduling-system: 'pbs' # or have some default in the code?
    name: dask-worker
    
    ...
    
    scheduler-options: {}

  my-awesome-pbs-cluster:
    job-scheduling-system: 'pbs'
    name: dask-worker
    cores: 36                   # Total number of cores per job
    memory: '109 GB'            # Total amount of memory per job
    processes: 9                # Number of Python processes per job
    interface: ib0              # Network interface to use like eth0 or ib0

    queue: regular
    walltime: '00:30:00'
[...]

Issue #95, opened 2018-07-14
Named configurations as an option to cluster constructors


josephhardinee:
Would there be interest in having named configuration sets for the different constructors in addition to the pbs and slurm (and other) default configuration sets? 

Use Case
----------
There are several use cases. The first is clusters that share filesystems between different scheduling systems (Cori and Edison on NERSC). The second would be supporting multiple queues (high_mem, debug, burst, scavenger, etc). This would allow you to have different configuration sets for different job types. 

Implementation
----------------
Currently when issuing a `PBSCluster()` call, the configuration is drawn from the 
jobqueue.pbscluster dictionary inside yaml. The proposed implementation would leave that alone, but would add a optional keyword `scheduler_name`. 
The user could then call 
```python
cluster = PBSCluster(scheduler_name='high_mem')
```

to access the portion of the yaml file given by the high_mem option 

```yaml
jobqueue:
    high_mem:
        cores: 64
        memory: 128GB
        processes: 4
        queue: debug
        walltime: '00:10:00'
        job-extra: ['-C haswell', '-L project, SCRATCH, cscratch1']
    pbs:
        cores: 64
        memory: 128GB
        processes: 4
        queue: debug
        walltime: '00:10:00'
        job-extra: ['-C haswell', '-L project, SCRATCH, cscratch1']
```
The keyword would default to the original pbs(or slurm, etc). 

The implementation is pretty straightforward, just have each of the cluster initiation objects set self.scheduler_name to optional parameter scheduler name. Then change each of the `dask.config.get` calls to use scheduler_name instead of their hardcoded values. 

If everyone is okay with this I can submit a PR next week with the implementation for comment.

mrocklin:
In general I can see value behind this idea.  Presumably we would define classes like the following:

```python
class PBSCluster:
    def __init__(..., name='pbs'):
        cores = cores or dask.config.get('jobqueue.%s.cores' % name
```

I'm curious to hear @jhamman 's thoughts.  I suspect that he has run into similar issues before

jhamman:
I like this idea. I would use it in a similar fashion as how @josephhardinee suggested. Currently, I've been overriding the default values in my class instantiation.

guillaumeeb:
@josephhardinee do you think you will have time/interest on getting this done? Would you prefer someone else to implement this?

guillaumeeb:
Closing thanks to #204. Thanks again.

PR #604, opened 2023-03-06
Add ability to instantiate named clusters from jobqueue config

alisterburt:
Discussed in #543 

This PR allows custom clusters of any type to be specified in the jobqueue yaml

Below is an example config file with a second PBS cluster which uses a different partition called 'gpu'

```yaml
jobqueue:
  pbs:
    cores: 36
    memory: 100GB
    queue: regular
  custom-pbs-cluster:
    job-scheduling-system: pbs
    queue: gpu

```

This can be instantiated with `JobQueueCluster.from_name()`

```python
from dask_jobqueue import JobQueueCluster

cluster = JobQueueCluster.from_name("custom-pbs-cluster")
```

This also works directly with the standard job scheduling systems
```python
from dask_jobqueue import JobQueueCluster

cluster = JobQueueCluster.from_name("pbs")
```

The main reasons for this PR are:
1. to enable instantiating clusters of arbitrary type (SLURM, PBS) from scripts without writing against a specific cluster class
2. ability to have multiple cluster configurations for the same scheduling system which differ slightly
3. pave the way for cluster discovery from dask-ctl (which would enable named cluster specification from there)

#543 was closed because the features provided here can be made available through a dask-ctl specific config yaml. I think this PR is still necessary because implementing [autodiscovery of dask-jobqueue clusters](https://dask-ctl.readthedocs.io/en/latest/integrating.html#discovery) in dask-ctl requires some way of 'discovering' possible clusters. With this PR, we can simply iterate over the keys in the jobqueue config and instantiate the specified clusters using `JobQueueCluster.from_name()`. 

Additionally, the config in dask-jobqueue is less Python specific than the dask-ctl yaml spec - this may be less of a burden for configuration by HPC admins who are not experts in Python.

#95 is similar functionality but implemented at the cluster class level so does not solve point 1 above.

cc @guillaumeeb because you were involved in earlier discussions

Thank you in advance for any time spent reviewing!


guillaumeeb:
Thanks @alisterburt for this proposal.

I went through the discussion in https://github.com/dask-contrib/dask-ctl/issues/61, and I'm not sure of the outcome. @jacobtomlinson do you think the change here are necessary or should we wait for a PR in `dask-ctl`?

alisterburt:
Thank you for taking the time to go through the PR/discussion @guillaumeeb!

I think the correct answer here depends on how you think dask-jobqueue users should ideally configure their clusters moving forwards. Is that through a jobqueue config or a dask-ctl config? There is a valid argument against having both, but...

- I doubt dask-jobqueue will ever disallow config through a jobqueue cluster in favour of dask-ctl completely
- dask-jobqueue only config minimises the surface area of dask API users have to understand
- this PR will enable discovery of clusters specified in a jobqueue config within dask-ctl

I see the two PRs (this and the future one in dask-ctl) as orthogonal and complementary unless there is an effort to move to using dask-ctl as the only way to create/manage clusters in the dask ecosystem

alisterburt:
fixed the silly test failure but workflow requires re-approval from a maintainer 🙂

dask_jobqueue/core.py
```diff
@@ -830,3 +829,24 @@ def adapt(
...
             kwargs["maximum"] = maximum_jobs * self._dummy_job.worker_processes
         return super().adapt(*args, **kwargs)
+
+    @classmethod
+    def from_name(cls, name: str):
+        """Initialise a named Dask cluster specified in the jobqueue config.
+
+        Named clusters should specify the appropriate job scheduling system under the
+        'job-scheduling-system' key in the jobqueue config alongside specific keyword
+        arguments passed to the JobQueue cluster class on initialisation.
```
guillaumeeb, line 839:
```suggestion
        arguments passed to the JobQueueCluster class on initialisation.
```

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -469,3 +470,45 @@ def test_security_temporary_defaults(EnvSpecificCluster, loop):
             future = client.submit(lambda x: x + 1, 10)
             result = future.result(timeout=30)
             assert result == 11
+
+
+def test_jobqueuecluster_from_name_for_existing_scheduling_systems():
+    for scheduling_system in CLUSTER_CLASSES.keys():
+        temporary_config = {
```
guillaumeeb, line 477: Is setting this temporary config really useful to this test?

dask_jobqueue/tests/test_jobqueue_core.py
```diff
@@ -469,3 +470,45 @@ def test_security_temporary_defaults(EnvSpecificCluster, loop):
...
+            f"jobqueue.{scheduling_system}.memory": "24GB",
+            f"jobqueue.{scheduling_system}.disk": "1GB",
+        }
+        with dask.config.set(temporary_config):
+            cluster = JobQueueCluster.from_name(scheduling_system)
+            assert isinstance(cluster, CLUSTER_CLASSES[scheduling_system])
+
+
+def test_jobqueuecluster_from_name_for_custom_cluster():
+    temporary_config = {
```
guillaumeeb, line 488: Here also you don't seem to check the custom config. Wasn't that mean to check that htcondor default config wasn't taken into account?

guillaumeeb (commented):
If this is all it takes in code modifications, clearly I'm in favor of this!
I'm almost surprised the `CLUSTER_CLASSES[job_sheduling_system](**config)` actually does the trick! We have consistent naming everywhere in kwargs and yaml entries?

I've left a few comments, and also it would be good to have some documentation about this functionality, could you add some?

Also, I'd like to have @jacobtomlinson approval before merging, at least on the idea!

jacobtomlinson (commented):
**TL;DR** I see this PR as stepping too far in the direction of implementing cluster lifecycle management and would prefer to see the broader community collaborating on `dask-ctl` rather than building duplicate solutions independently.

I disagree about this being orthogonal to the `dask-ctl` discussion. From a project scope and overall Dask maintenance perspective this feels like it shouldn't live in `dask-jobqueue`.

The problem this PR solves is "How can admins preconfigure user environments with some cluster templates?". The discussion in `dask-ctl` solves the same problem but for everyone. 

> **Note**
> [Dask Gateway also already has a solution for this for HPC](https://gateway.dask.org/cluster-options.html). I'm keen to upstream things from gateway into ctl too because gateway has limited scope and use cases.

> **Note**
> Dask Labextension also has similar functionality (which inspired the `dask-ctl` code) for [defining a template for the "New cluster" button](https://github.com/dask/dask-labextension/blob/99c4d2a1a1cbff61322356be6566432e87c4d72f/dask_labextension/labextension.yaml#L2-L6). The goal is for this to leverage `dask-ctl` in the future instead.

My goal is to try and push as much cluster lifecycle logic as possible into `dask-ctl` so that it is all in one place and is easily reusable between subprojects. I dislike the idea of individual deployment projects implementing their own solutions to this problem, and other problems like it, rather than collaborating on a shared tool like `dask-ctl`.

> I see the two PRs (this and the future one in dask-ctl) as orthogonal and complementary unless there is an effort to move to using dask-ctl as the only way to create/manage clusters in the dask ecosystem

The goal of `dask-ctl` is to be a control plane for all Dask clusters. It wont be the *only* way to do things but it will be the most consistent and convenient way to do cluster lifecycle things.

> I doubt dask-jobqueue will ever disallow config through a jobqueue cluster in favour of dask-ctl completely

I agree and I expect in the future that users can still use the individual libraries directly to "pull the cables and move the flaps" if they want to. 

I also think cluster creation will always be simplest by using the library directly. The value in `dask-ctl` is value add things like discovery, listing, aggregating, templating and single-pane-of-glass views.

> dask-jobqueue only config minimises the surface area of dask API users have to understand
this PR will enable discovery of clusters specified in a jobqueue config within dask-ctl

I would push back on this. This PR expands the Dask API here instead of guiding folks to use an existing API. From a user perspective there is still API to learn.

We are on a journey with `dask-ctl` where hopefully one day it will be upstreamed to `distributed` or made directly available via the `dask` namespace like `dask.distributed` is today. But to get there we need folks to use it rather than working independently.

My question to @alisterburt is what can we do to make `dask-ctl` better and reduce friction there? For example I totally sympathise with your point about the use of the keywords `args` and `kwargs` in the cluster spec. Maybe we want to change that to something easier to understand in a `v2` spec? (this is why I versioned the spec, so we can iterate nicely)

alisterburt:
@guillaumeeb @jacobtomlinson thank you both for the reviews!

@guillaumeeb as there currently is no consensus around whether this PR is wanted I will hold off on updating for now.

@jacobtomlinson I am completely on board with the centralisation of cluster lifecycle management in dask-ctl, one of the main reasons for this PR was to simplify the implementation of dask-ctl discovery in dask-jobqueue in a followup PR 🙂

> The problem this PR solves is "How can admins preconfigure user environments with some cluster templates?". The discussion in dask-ctl solves the same problem but for everyone.

> The goal of dask-ctl is to be a control plane for all Dask clusters. It wont be the *only* way to do things but it will be the most consistent and convenient way to do cluster lifecycle things.

> I expect in the future that users can still use the individual libraries directly to "pull the cables and move the flaps" if they want to.

I think the main problem this PR solves is actually "how can users preconfigure multiple clusters of the same type without leaving jobqueue?". This is currently not possible using the jobqueue config and adding dask-ctl into the mix is a good (i.e. not ad-hoc) solution. I think not being able to do this is a deficiency of the current jobqueue config that should be solved in jobqueue.

> I also think cluster creation will always be simplest by using the library directly. The value in dask-ctl is value add things like discovery, listing, aggregating, templating and single-pane-of-glass views.

> My goal is to try and push as much cluster lifecycle logic as possible into dask-ctl so that it is all in one place and is easily reusable between subprojects. I dislike the idea of individual deployment projects implementing their own solutions to this problem, and other problems like it, rather than collaborating on a shared tool like dask-ctl.

> We are on a journey with dask-ctl where hopefully one day it will be upstreamed to distributed or made directly available via the dask namespace like dask.distributed is today. But to get there we need folks to use it rather than working independently.

This is exciting! It feels like clearly delineating what dask-ctl and dask-foo should be responsible for (in the ideal case) moving forwards would be useful here. I'm still a bit unsure 

> My question to @alisterburt is what can we do to make dask-ctl better and reduce friction there? For example I totally sympathise with your point about the use of the keywords args and kwargs in the cluster spec. Maybe we want to change that to something easier to understand in a v2 spec? (this is why I versioned the spec, so we can iterate nicely)

For my own personal use dask-ctl is working wonderfully, I'm happy creating clusters from yaml paths for now and will be happier creating from names once we get dask/dask-ctl#61 going. As a developer making things for non-python folks most likely working in HPC, I really like the idea of developing against dask-ctl and creating/discovering dask-foo clusters from there. 

This is how I imagine the top of programs I will provide to people

```python
from dask_ctl import get_cluster, create_cluster

# cluster_name comes in from the outside world
try:
    cluster = get_cluster(cluster_name)
except NameError:
    cluster = create_cluster(cluster_name)
```

The friction for my use case comes in when asking users to configure a cluster and they need jobqueue and ctl rather than just jobqueue - I'm optimising to keep the set of things users interact with as minimal as possible

Overall I am happy to go either way in this repo and will PR to dask-ctl soon regardless but I do think there is merit to this including this PR in jobqueue 🙂

jacobtomlinson:
> I think not being able to do this is a deficiency of the current jobqueue config that should be solved in jobqueue.

I think the core of my comments are that I think creating cluster templates is out of scope for `dask-{jobqueue,kubernetes,cloudprovider,yarn,etc}` and we should be doing this in `dask-ctl`.

To quote the [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python) `There should be one– and preferably only one –obvious way to do it.`.

> one of the main reasons for this PR was to simplify the implementation of dask-ctl discovery in dask-jobqueue

I haven't reviewed this PR line by line yet as I wanted to have the high level discussion first. However at a quick glance I don't think this satisfies the `dask-ctl` expectations for `from_name`. Two (or more) processes should both be able to call `dask_jobqueue.SLURMCluster.from_name("awesomecluster")` and all get a pointer to exactly the same cluster. My reading of this PR would be that a new cluster would be spawned each time.

In this PR `name` seems to be a template name. In a `dask-ctl` contect `name` is the unique name of a cluster instance.

> This is how I imagine the top of programs I will provide to people

That looks like a nice example. In `dask-kubernetes` we do a check for an existing cluster with that name before trying to create a new one so the code example is further simplified. It would be nice to do this in more places.

```python
from dask_kubernetes.operator import KubeCluster

cluster = KubeCluster(name="foo")  # If 'foo' exists it connects to it, if not it creates it (most users want this)

cluster = KubeCluster.from_name("foo")  # If 'foo' exists it connects to it, if not it raises an exception

cluster = KubeCluster(name="foo", create_mode="CREATE_ONLY")  # If 'foo' exists it raises an exception, if not it creates it
```

> The friction for my use case comes in when asking users to configure a cluster and they need jobqueue and ctl rather than just jobqueue

I totally understand, ultimately these are just all bits of Dask but we avoid installing them all by default to reduce bloat in the core package. Do you see this problem being more of an install-time issue or import-time issue? Would making `dask-ctl` a dependency of `dask-jobqueue` resolve the install-time friction?

alisterburt:
> I think the core of my comments are that I think creating cluster templates is out of scope for dask-{jobqueue,kubernetes,cloudprovider,yarn,etc} and we should be doing this in dask-ctl.

> To quote the [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python) There should be one– and preferably only one –obvious way to do it..

Cool! Thank you for being patient, this is clear guidance I can get behind 🙂 let's close this PR and push for realising the ideal ecosystem then.

> Two (or more) processes should both be able to call dask_jobqueue.SLURMCluster.from_name("awesomecluster") and all get a pointer to exactly the same cluster. My reading of this PR would be that a new cluster would be spawned each time.

Your reading is correct and I think my understanding of dask here is lacking, if I had multiple users in a HPC environment what are the advantages/disadvantages of having them connect to the same Client/Cluster vs creating their own? Naively I thought we would want users to run their own for resiliency (other users can't crash my thing) and to avoid overwhelming the task graph if there are multiple heavy users.

Thanks for the dask-kubernetes example - that's clean and flexible! 

> I understand that separate namespaces can add friction. Do you see this problem being more of an install-time issue or import-time issue? Would making dask-ctl a dependency of dask-jobqueue resolve the install-time friction?

Sure - I think ~all friction would be removed if dask-ctl were bundled with jobqueue/dask itself at install time and the jobqueue documentation used dask-ctl for cluster templates rather than its own config. @guillaumeeb would you/the other maintainers here accept PRs in this direction?

alisterburt:
You're a star @jacobtomlinson - I learned a bunch and the motivation behind centralising this management is now crystal clear, thanks for writing this up!

jacobtomlinson:
> Should we drop (part of?) the underlying code at some point?

I'm not suggesting removing any of the existing configuration in `dask-jobqueue`. All the configuration does today is set the default kwargs for the `HPCCluster` objects which totally makes sense.

My comment was more that the config here shouldn't be expanded to include lifecycle things like cluster templates. Instead we should work together to do this in `dask-ctl` so that all `dask-foo` projects can benefit from shared functionality.

PR #605, opened 2023-04-22
start of work to add flux

vsoch:
This is the start of work to add [Flux Framework](https://flux-framework.readthedocs.io/en/latest/) to dask. For some context, we are looking to run Dask in the [Flux Operator](https://flux-framework.org/flux-operator/) so I'm looking into how to make that possible!

Update: everything working locally! I'll tweak here as necessary for linting / tests to pass.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -128,6 +128,7 @@ class Job(ProcessInterface, abc.ABC):
     See Also
     --------
     PBSCluster
+    FluxCluster
     SLURMCluster
     SGECluster
     OARCluster
@@ -745,7 +746,6 @@ or by setting this value in the config file found in `~/.config/dask/jobqueue.ya
         for key, value in worker_security_dict.items():
             # dump worker in-memory keys for use in job_script
             if value is not None and "\n" in value:
-
                 try:
                     f = tempfile.NamedTemporaryFile(
                         mode="wt",
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -353,7 +353,6 @@ class Job(ProcessInterface, abc.ABC):
             command_args += worker_extra_args
 
         # Save command args to be extended, etc.
-        self._command_args = command_args
         self._command_template = " ".join(map(str, command_args))
 
         self.log_directory = log_directory
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -352,6 +352,8 @@ class Job(ProcessInterface, abc.ABC):
         if worker_extra_args is not None:
             command_args += worker_extra_args
 
+        # Save command args to be extended, etc.
+        self._command_args = command_args
         self._command_template = " ".join(map(str, command_args))
 
         self.log_directory = log_directory
--- a/dask_jobqueue/flux.py
+++ b/dask_jobqueue/flux.py
@@ -149,18 +149,12 @@ class FluxCluster(JobQueueCluster):
     ----------
     queue : str
         Destination queue for each worker job. Passed to `#SBATCH -p` option.
-    project : str
-        Deprecated: use ``account`` instead. This parameter will be removed in a future version.
-    account : str
-        Flux does not have support for an account
     {job}
     {cluster}
     walltime : str
         Walltime for each worker job.
     job_cpu : int
         Number of cpu to book in Flux, if None, defaults to worker `threads * processes`
-    job_mem : str
-        Flux does not have support to control memory.
     job_extra : list
         Deprecated: use ``job_extra_directives`` instead. This parameter will be removed in a future version.
     job_extra_directives : list
--- a/dask_jobqueue/flux.py
+++ b/dask_jobqueue/flux.py
@@ -1,8 +1,6 @@
-from contextlib import contextmanager, suppress
+from contextlib import contextmanager
 import logging
-import json
 import os
-import shlex
 import stat
 import tempfile
 import warnings
@@ -13,10 +11,42 @@ from .core import Job, JobQueueCluster, job_parameters, cluster_parameters
 logger = logging.getLogger(__name__)
 
 
+def timestr2seconds(timestr):
+    """
+    Given a timestring in two formats, return seconds (float).
+
+    DAYS-HOURS:MINUTES:SECONDS
+    0-00:00:00
+    """
+    if timestr.count(":") == 1:
+        minutes, seconds = timestr.split(":")
+        return (int(minutes) * 60) + float(seconds)
+
+    # days hours, minutes and seconds, MM:SS.mm
+    elif "-" in timestr and timestr.count(":") == 2:
+        days, rest = timestr.split("-", 1)
+        hours, minutes, seconds = rest.split(":")
+        return (
+            (int(days) * 86400)
+            + (int(hours) * 3600)
+            + (int(minutes) * 60)
+            + float(seconds)
+        )
+
+    # hours, minutes, seconds HH:MM:SS.mm
+    elif timestr.count(":") == 2:
+        hours, minutes, seconds = timestr.split(":")
+        return (int(hours) * 3600) + (int(minutes) * 60) + float(seconds)
+
+    raise ValueError(f"Unrecognized time format {timestr}")
+
+
 class FluxJob(Job):
     # Override class variables
     submit_command = "flux batch"
-    cancel_command = "flux cancel"
+
+    # This can be updated to flux cancel when release 0.48 been out a while
+    cancel_command = "flux job cancel"
     config_name = "flux"
 
     def __init__(
@@ -30,12 +60,11 @@ class FluxJob(Job):
         job_cpu=None,
         job_mem=None,
         config_name=None,
-        **base_class_kwargs
+        **base_class_kwargs,
     ):
         super().__init__(
             scheduler=scheduler, name=name, config_name=config_name, **base_class_kwargs
         )
-
         # This would better be shared logic across backends...
         if queue is None:
             queue = dask.config.get("jobqueue.%s.queue" % self.config_name)
@@ -53,30 +82,26 @@ class FluxJob(Job):
 
         header_lines = []
 
+        # Error and output files
+        error = "%s/%s-%%J.err" % (self.log_directory, self.job_name or "worker")
+        output = "%s/%s-%%J.out" % (self.log_directory, self.job_name or "worker")
+
         # Flux header build
         if self.job_name is not None:
             header_lines.append("#flux: --job-name=%s" % self.job_name)
         if self.log_directory is not None:
-            header_lines.append(
-                "#flux: --error %s/%s-%%J.err"
-                % (self.log_directory, self.job_name or "worker")
-            )
-            header_lines.append(
-                "#flux: --output %s/%s-%%J.out"
-                % (self.log_directory, self.job_name or "worker")
-            )
+            header_lines.append("#flux: --error %s" % error)
+            header_lines.append("#flux: --output %s" % output)
+
         if queue is not None:
             header_lines.append("#flux: --queue=%s" % queue)
 
-        # Init resources, always 1 task,
-        # and then number of cpu is processes * threads if not set
+        # Init resources, always 1 task
         header_lines.append("#flux: -n 1")
-        header_lines.append(
-            "#flux: --cores-per-slot=%d" % (job_cpu or self.worker_cores)
-        )
+        header_lines.append("#flux: --cores-per-slot=%s" % self.worker_cores)
 
         if walltime is not None:
-            header_lines.append("#flux: -t %s" % walltime)
+            header_lines.append("#flux: -t %s" % timestr2seconds(walltime))
 
         # Skip requested header directives
         header_lines = list(
@@ -85,7 +110,6 @@ class FluxJob(Job):
                 header_lines,
             )
         )
-
         # Add extra header directives
         header_lines.extend(["#flux: %s" % arg for arg in self.job_extra_directives])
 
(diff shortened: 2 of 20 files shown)


ocaisa:
I am really happy to see this! I think adding Flux support brings a ton of possibilities since it would allow us to run our own scheduler inside a larger allocation (which is great for my use case in [jobqueue_features](https://github.com/E-CAM/jobqueue_features))

vsoch:
Okay! Did some more work / debugging today and I learned the following:

1. Flux does better to be given an executable, full path (not a relative path in /tmp that isn't executable)
2. Writing the file as a context makes a race condition where it's not always the case that the file has been read and submit by the time the context is finished. To get around this I removed the context, an we will need to do cleanup later / elsewhere. 

I think I'm making progress and will push some updates soon! I also think it shouldn't be the case we have some kind of race condition with the submit and getting the job id back - I'll minimally report this to the flux team to see if it's something we can fix (so the behavior is similar to the other managers).

vsoch:
Either GitHub is having issues (I saw something similar with Flux earlier this week) or some change I introduced / the docker compose setup with flux is leading to the others to stall, although strangely htcondor seems to have worked okay? I'll keep watching!

vsoch:
@ocaisa I am totally new to Dask - could you tell me about jobqueue features? The MPI mention has peaked my interest - Flux is very good at that! :laughing: 

If the Flux job queue here could work, if you want to point me to a vanilla example there I could definitely try it out or help think about it.

vsoch:
Doing more testing (with an actual kubernetes cluster and flux!) I think the cleanup might not be in the right space still - I am running a distributed training and everything looked ok, but I started seeing the job errors (failures) because the job file is reported to not be found. 
image

I'm testing a slightly modified branch to not do the cleanup and will report back.

vsoch:
okay update - what I'm seeing is that the model actually finishes _before_ all the jobs are done, maybe because it launches more than it needs / decides it's finished / at some threshold and then doesn't need the rest? Because they are technically just workers waiting to receive work? Here is the output printed, which happens after the joblib context:

image

and here is another terminal that is just watching the flux queue - green "CD" means completed and most are running

image

When I look at one of the running workers it seems to just be registered / waiting

image

First I tried changing the cluster and client to be used as context managers, and left the cleanup. This still led to a few errored jobs, so I removed it for now (PR updated here). I also noticed that the cluster.wait_for_workers seems to launch one job, and then wait for workers as processes on a single node? I'm not sure if there is a "wait for the entire cluster" function, but that isn't super relevant I think. I also noticed when I didn't clean up the temporary space, for the next round I would get a timeout starting the cluster, I think (maybe) because node/worker information is cached there, and then it never finds them.

TLDR: my guess with the above is that the training finishes, and jobs were left that were scheduled, but then when they actually run, the script has long since been cleaned up. For the time being (until we have a better idea) it makes sense to disable the cleanup. I don't see any errors in this case:

image

And since one of those job scripts is technically just starting a worker, e.g.,

```bash
fluxuser@flux-sample-0:/tmp/workflow$ cat tmp/flux-job-okvkaeiy.sh 
#!/usr/bin/env bash

#flux: --job-name=dask-worker
#flux: -n 1
#flux: --cores-per-task=2
#flux: -t 00:30:00

/opt/conda/bin/python3 -m distributed.cli.dask_worker tcp://10.244.0.64:38251 --nthreads 0 --nworkers 4 --memory-limit 476.84MiB --name FluxCluster-4 --nanny --death-timeout 60
```
It probably doesn't hurt that it just randomly runs - the worker seems to start and be waiting around!  In this context, when the job finishes (and the broker exits with 0) the rest of the jobs/pods exit too so it largely doesn't matter. Here is the full workflow and documentation I'm working on:

https://github.com/flux-framework/flux-operator/blob/add/dask-example-scikit-learn/examples/machine-learning/dask/scikit-learn/README.md

Specifically check out the [notes section](https://github.com/flux-framework/flux-operator/blob/add/dask-example-scikit-learn/examples/machine-learning/dask/scikit-learn/README.md#notes) where I've added more of the comments above. I'd like to understand the best practices for ending / cleaning up so we don't have dangling jobs or workers. Thank you!

vsoch:
Another update: I was wrong in using flux submit, it should be flux batch! I thought the directives worked in either (but they don't). I'm working on fixing it now.

vsoch:
okay updated to use `flux batch` and thanks @grondo and @garlick for the immense help today! The reason it took me so long is because flux only detected one core per docker compose container, so when I tried specifying two, it gave me an error. Instead of adding a deploy section with resources I just manually set the cores per container to 2. That works locally and seems to work here (and will mirror the slurm tests). I'm testing again in the operator now.

vsoch:
I think I found the bug for why the worker jobs weren't cleaning up - it catches any runtime error (assuming the worker might be done) but lo and behold, `flux cancel` isn't a command for this build I used. I guess I'm kind of confused because I was told here https://github.com/flux-framework/cheat-sheet/pull/3 that we are trying to use `flux cancel` (which doesn't seem to be a command). So for now I'm falling back to `flux job cancel`.

garlick:
FYI: `flux cancel` became a thing in flux-core 0.48 (released march 8).

vsoch:
This is interesting - with this change (so cancel should work now) I get an almost immediate failure (it doesn't wait 120s) when I do a `client.restart()`

```
broker.info[0]: quorum-full: quorum->run 32.7692s
       nodes: 4
       cores: 2
     timeout: 60
Preparing cluster with 4 workers
Client: <Client: 'tcp://10.244.0.30:38095' processes=4 threads=8, memory=7.44 GiB>
Value of n is 8
Iteration 0 tasks
2023-04-24 00:05:46,519 - distributed.core - ERROR - 4/4 nanny worker(s) did not shut down within 120s
  File "/opt/conda/lib/python3.10/site-packages/distributed/utils.py", line 752, in wrapper
    return await func(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/distributed/scheduler.py", line 5962, in restart
    raise TimeoutError(
asyncio.exceptions.TimeoutError: 4/4 nanny worker(s) did not shut down within 120s
```

vsoch:
okay confirmed that when cancel works, it actually cancels the batch job before it's being done waited for! I just pushed a change that is going to test removing the weakref in the start command, will report back.

vsoch:
I think I'm going to need some feedback or insight from as Dask developer here - when I remove the cleanup weakref line in start my jobs run as I'd expect, but now the tests are failing. I think we probably want a way to do the cleanup but not have it killing the workers before the cluster has been waited for!

vsoch:
@ocaisa do you know how the review process works here? Is the project even active?

guillaumeeb:
Hi @vsoch, sorry I did not find the time to review your work here, and as I don't know anything about Flux, I need to take some time to deep into it. Please be patient 🙂.

vsoch:
oh hey!! Yeah I'm totally good with that! I just wanted to make sure I didn't muck this up by putting it in the wrong place, etc. Thank you for the response @guillaumeeb I didn't mean to be impatient (more anticipating that I did something wrong, really, and wanting to fix it) :laughing:

dask_jobqueue/core.py
```diff
@@ -387,6 +390,7 @@ def job_file(self):
             with open(fn, "w") as f:
                 logger.debug("writing job script: \n%s", self.job_script())
                 f.write(self.job_script())
+                print(self.job_script())
```
ocaisa, line 393: This is for debugging, so should be removed

dask_jobqueue/jobqueue.yaml
```diff
@@ -125,6 +125,39 @@ jobqueue:
...
+    processes: null
+
+    interface: null
+    death-timeout: 60
+    local-directory: null
+    shared-temp-directory: null
+    extra: null
+    worker-extra-args: []
+
+    # SLURM resource manager options
```
ocaisa, line 143:
```suggestion
    # flux resource manager options
```

ocaisa (changes requested):
@guillaumeeb I saw in https://github.com/dask/dask-jobqueue/pull/598#issuecomment-1573473861 there was a mention of a release. If that did happen it would be nice to include the flux support as well.

I did a basic review of the code here, but haven't taken it for a test drive. I'd like to think the test suite is ensuring that capturing the functionality support and if there are issues for users I'm sure @vsoch would be willing to address them. 

What is missing is some documentation under https://jobqueue.dask.org/en/latest/examples.html and perhaps https://jobqueue.dask.org/en/latest/configurations.html

vsoch:
Thank you @ocaisa ! It's been a while since I tested this, but I think I had some questions about the expected functionality in my original description / post (for example, the worker cleanup). If you want me to go back and walk through a simple case, maybe that would be a good start?

dask_jobqueue/flux.py
```diff
@@ -0,0 +1,187 @@
...
+
+        if queue is not None:
+            header_lines.append("#flux: --queue=%s" % queue)
+
+        # Init resources, always 1 task
+        header_lines.append("#flux: -n 1")
+        header_lines.append("#flux: --cores-per-slot=%s" % self.worker_cores)
+
+        if walltime is not None:
+            header_lines.append("#flux: -t %s" % timestr2seconds(walltime))
```
guillaumeeb, line 104: If I understand correctly, flux takes a walltime directive as an integer (seconds), so I'm not sure why you are bothering with a different format in input and define a timestr2seconds function?

dask_jobqueue/flux.py
```diff
@@ -0,0 +1,187 @@
...
+class FluxCluster(JobQueueCluster):
+    __doc__ = """Launch Dask on a Flux cluster
+
+    Parameters
+    ----------
+    queue : str
+        Destination queue for each worker job. Passed to `#SBATCH -p` option.
+    project : str
+        Deprecated: use ``account`` instead. This parameter will be removed in a future version.
+    account : str
```
guillaumeeb, line 154: Maybe you should just remove the project, account, and job_mem from the docs if it isn't supported by Flux? There was some discussions to move them up, but all Scheduler doesn't take same kwargs, so you should be able to just remove all the related code.

dask_jobqueue/jobqueue.yaml
```diff
@@ -125,6 +125,39 @@ jobqueue:
...
+    shebang: "#!/usr/bin/env bash"
+    queue: null
+    account: null
```
guillaumeeb, line 146: Same here, you can remove what is not needed.

dask_jobqueue/tests/test_flux.py
```diff
@@ -0,0 +1,242 @@
...
+            worker_names = [
+                w["id"] for w in client.scheduler_info()["workers"].values()
+            ]
+            assert sorted(worker_names) == [
+                "test-my-env-variable-value-0",
+                "test-my-env-variable-value-1",
+            ]
+
+
+@pytest.mark.env("flux")
```
guillaumeeb, line 215: Are there some tests which can be run without spawning a complete Flux cluster? If so, they should be without this fixture I think?

dask_jobqueue/core.py
```diff
@@ -351,6 +352,8 @@ def __init__(
         if worker_extra_args is not None:
             command_args += worker_extra_args
 
+        # Save command args to be extended, etc.
+        self._command_args = command_args
```
guillaumeeb, line 356: I'm not sure why you add this line, do you use it anywhere else?

guillaumeeb (commented):
Thanks a lot @vsoch for al the work here. I still don't know Flux, but it looks like it's a good fit to dask-jobqueue. 

What is really nice is that you built a complete CI setup, this is great!! I have to trust you (and the CI runs) for this part, because I cannot at all tell if it is good or not. Anyway, in its structure, it looks good!

I made a few comments to be addressed on the review. 

Once this is done, since the PR is almost completely additions and not any modification of existing code, I think it would be safe to merge as soon as the CI is green.

dask_jobqueue/flux.py
grondo: Clarification: I don't know the format of `walltime` here, but the Flux `-t, --time-limit` option is in floating point minutes by default (though it can tak a `s`, `m`, `h`, or `d` suffix for explicit seconds, minutes, hours, or days)

vsoch:
Starting work on this now! We have had some changes the Flux Operator in the last month and a half, so I'm going to start fresh and walk through the tests / feedback above. Stay tuned!

dask_jobqueue/flux.py
vsoch: Okay sounds like I should do:

```python
header_lines.append("#flux: -t {walltime}s")
```

dask_jobqueue/tests/test_flux.py
vsoch: Maybe if we mocked it? But it seems easier to just spin up a container base with flux. Let me know what you have in mind specifically though!

dask_jobqueue/core.py
vsoch: I had an initial design that did (and removed it) - will remove this line too.

dask_jobqueue/flux.py
vsoch: Nope that definitely didn't work! Going to keep it for now. 

```
#flux: -t 00:02:00'
```
```console
RuntimeError: Command exited with non-zero exit code.
Exit code: 1
Command:
flux batch /tmp/flux-job-47lcw93z.sh
stdout:

stderr:
flux-batch: ERROR: invalid Flux standard duration
```

vsoch:
If / when tests pass please don't merge yet! I'm testing in the Flux operator and trying to optimize for the new design. I'll post an update when I'm done.

vsoch:
okay this is a bad design and I can't support moving forward without more thinking. Right now, dask requires a shared temporary space to write the job scripts and have a scratch space for itself. I was doing a hack to make a host volume that would be this location (via `TMPDIR`) but this gets ugly really quickly because other applications grab onto it, and all of a sudden we have the case of needing to clean up the dask stuff, but not the other temporary files (e.g., a flux local socket) and then if you aren't absolultely careful, the entire thing breaks (and it's a really bad setup / standard to start from). If we use this approach:

```
# Ensure the place dask writes files is shared by all nodes!
dask.config.set(temporary_directory='/tmp/workflow/tmp')
```
That works for scratch, but not for the job files, so then we run the job and the other workers (different pods) can't find them.

I don't see a way forward with this design - I've gotten it working maybe twice when I first opened the PR, and it seemed to work under some mysterious circumstances once today, but it's not consistently not working and I'm out of ideas. I think likely we need to step back, rethink this design, and (for the time being) remove the example from the operator because it doesn't work. Let me know your thoughts!

jacobtomlinson:
It's been a long time since this PR had any activity and given that you don't want to continue with the current design so I'm going to close it out.

vsoch:
Would dask ever work without having this shared location for the workers? A shared filesystem is nontrivial in Kubernetes.

jacobtomlinson:
The worker temporary directory doesn't need to be shared between all workers. Each worker uses it's own directory.

vsoch:
Gotcha - this PR is old enough that it would be worth trying again. I'll start fresh and open a new one if I have any success. Thanks for keeping on top of this @jacobtomlinson.

PR #606, opened 2023-05-26
Make cli worker parameter flexible

### LINKED ISSUE #229: Provide parameter for "dask-worker"
Issue https://github.com/rapidsai/dask-cuda/issues/3 proposes creating a `dask-cuda-worker` alternative to `dask-worker` which starts one process per GPU with appropriate environment variables.  If we go ahead with that then it would be convenient to compose that approach with dask-jobqueue.  This would be particularly easy if the following line could be made configurable to allow other executables

https://github.com/dask/dask-jobqueue/blob/5a5585dffea0237e452792c7ed0b13d5d1e3f8bb/dask_jobqueue/core.py#L254

Thoughts?

hmacdope:
Fixes #229 
Depends https://github.com/rapidsai/dask-cuda/pull/1181

Hi all! 

Basic implementation of calling CLI's other than the `distributed` CLI. 

Will require https://github.com/rapidsai/dask-cuda/pull/1181 to be merged first, adding a `main` entrypoint to `dask-cuda-worker`

I have had to add a bit of a shim to filter out some CLI args that are not shared between the `dask-worker` and `dask-worker-cuda` CLIs. 

Very basic test to see that it works:

```python 
import dask
from dask import distributed
from dask_jobqueue.local import LocalCluster

lc_gpu = LocalCluster(worker_command="dask_cuda.cli", cores=2, memory="2GB")
client = distributed.Client(lc_gpu)
lc_gpu.scale(2)
print(lc_gpu.job_script())
> /home/hmacdope/anaconda3/envs/dask_dev/bin/python -m dask_cuda.cli tcp://192.168.1.5:33677 --name dummy-name --nthreads 1 --memory-limit 0.93GiB --death-timeout 60
```

I am a new contributor so please let me know if I am missing anything obvious.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -346,7 +346,7 @@ class Job(ProcessInterface, abc.ABC):
 
         command_args = [dask_worker_command, self.scheduler]
 
-        # common 
+        # common
         command_args += ["--name", str(name)]
         command_args += ["--nthreads", self.worker_process_threads]
         command_args += ["--memory-limit", self.worker_process_memory]
@@ -356,7 +356,6 @@ class Job(ProcessInterface, abc.ABC):
             if processes is not None and processes > 1:
                 command_args += ["--nworkers", processes]
             command_args += ["--nanny" if nanny else "--no-nanny"]
-
 
         if death_timeout is not None:
             command_args += ["--death-timeout", death_timeout]
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -48,6 +48,8 @@ job_parameters = """
         Seconds to wait for a scheduler before closing workers
     extra : list
         Deprecated: use ``worker_extra_args`` instead. This parameter will be removed in a future version.
+    worker_command : list
+        Command to run when launching a worker.  Defaults to "distributed.cli.dask_worker"
     worker_extra_args : list
         Additional arguments to pass to `dask-worker`
     env_extra : list
@@ -166,6 +168,7 @@ class Job(ProcessInterface, abc.ABC):
         death_timeout=None,
         local_directory=None,
         extra=None,
+        worker_command="distributed.cli.dask_worker",
         worker_extra_args=None,
         job_extra=None,
         job_extra_directives=None,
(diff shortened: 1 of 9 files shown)


hmacdope:
Another option we could pursue is to just call them as scripts  without `python -m` let me know what you think

dask_jobqueue/core.py
```diff
@@ -166,6 +168,7 @@ def __init__(
         death_timeout=None,
         local_directory=None,
         extra=None,
+        worker_command="distributed.cli.dask_worker",
```
guillaumeeb, line 171: To be consistent with other kwargs, the default here should be None, and the true default defined in the configuration file. See how other kwargs are handled.

dask_jobqueue/core.py
```diff
@@ -332,17 +335,24 @@ def __init__(
...
 
-        command_args += ["--memory-limit", self.worker_process_memory]
+        # common 
         command_args += ["--name", str(name)]
-        command_args += ["--nanny" if nanny else "--no-nanny"]
+        command_args += ["--nthreads", self.worker_process_threads]
+        command_args += ["--memory-limit", self.worker_process_memory]
+
+        #  distributed.cli.dask_worker specific
+        if worker_command == "distributed.cli.dask_worker":
```
guillaumeeb, line 351: This is not ideal, but to be honest I don't see how to handle this differently.

guillaumeeb (changes requested):
Thanks a lot @hmacdope for contributing here!!

Just made a comment on the code, and it would be really nice to add the test you propose in your first post.

dask_jobqueue/core.py
hmacdope: Done!

hmacdope:
@guillaumeeb hopefully I have addressed your comments.

guillaumeeb (approved):
This looks good to me! We don't need to wait for the Cuda issue to be fixed to merge, right?

Also, if at some point you could add a bit documentation about using this new option, this would be really nice!

guillaumeeb:
We'll also wait till the CI is green!

hmacdope:
> This looks good to me! We don't need to wait for the Cuda issue to be fixed to merge, right?

It is merged!

> Also, if at some point you could add a bit documentation about using this new option, this would be really nice!

Yes I will raise an issue.

hmacdope:
@guillaumeeb any idea why the container builds are failing?

guillaumeeb:
Nope, I need to check these. This might not be a problem, but the CI / build (none) failing is, could try to check this one?

guillaumeeb:
I'm not sure what the problem is with the build, it looks like the image takes too much time to build, weird, more than 6 hours !!

jacobtomlinson:
Dask/distributed recently dropped Python 3.8 and I noticed that the failing CI uses it, maybe that is the problem?

https://github.com/dask/distributed/blob/129b7cb70e2b77f4e13e27aefe9a7dbfc31a53e4/pyproject.toml#L26

hmacdope:
@guillaumeeb need anything more from me here?

hmacdope:
@guillaumeeb @jacobtomlinson @lesteve would be great to get this finalised if possible?

guillaumeeb:
Sorry for the delay here @hmacdope, I wanted to check the CI but didn't have the time. I'll merge anyway. Would you need an official release at some point?

hmacdope:
> Sorry for the delay here @hmacdope, I wanted to check the CI but didn't have the time. I'll merge anyway. Would you need an official release at some point?

Thanks so much @guillaumeeb! No worries everyone is busy. :) all good without an official release.

Issue #517, opened 2021-09-08
Python entry is missing in config file


PhML:
I don’t know if it is on purpose or not but `python` argument can’t be set from config file (if not provided at cluster initialization, it is set to `sys.executable`).

My use case is that I run dask in a singularity container on a PBS cluster so it can’t rely on `sys.executable`.

guillaumeeb:
Hi @PhML, thanks for the issue.

I don't think it is on purpose, it's true that it's one of the few (the only one? Didn't really check) argument that is not available in config file. 

I'm not sure how it impacts your use case, but any PR making it easier to run dask-jobqueue in Singularity container would be welcomed! Are you interested in proposing something? 

At least, could you described more precisely your use case and the problem encountered?

guillaumeeb:
@PhML, do you have more input on this? Can't you use the `python` kwarg when building your `Cluster` object?

Wihtout any news from you in the following weeks, I'm going to close this issue as stale, but any PR proposing to solve you issue will be welcomed.

guillaumeeb:
Closing this issue as stale for now.

PR #623, opened 2024-02-05
Python executable from config file

fnattino:
Hi there! 

I encountered the same issue as #517, also for a similar (the same?) use case. 

I am running Dask Jobqueue on a SLURM system where all Python applications have to run inside (Apptainer) containers. Thus, I would like to be able to configure Dask jobqueue to use a containerized Python executable.

Currently, I can specify a custom Python path using the Dask Jobqueue Python API, but I cannot do it via the config file. It would be nice to allow for such feature, especially for users who use the Jupyter dask-lab extension, which relies uniquely on the values from the config files.

Introducing the changes as in this PR allows me to set the following in the Dask config file, so that all workers spawned by the cluster run inside containers:

```yaml
jobqueue:
  slurm:
    python: 'apptainer exec oras://ghcr.io/fnattino/test-jupyterdask-image-apptainer:latest python'
```


guillaumeeb:
Thanks! I think you need to reformat some files using black.

guillaumeeb:
Thanks, merging!

PR #626, opened 2024-02-09
Fix readthedocs build

fnattino:
`build.os` is now required in the readthedocs configuration file: https://blog.readthedocs.com/use-build-os-config/


guillaumeeb:
Thanks, I wanted to look at this too. Unfortunately it is still failing.

fnattino:
Apologies for the "intrusion", I noticed this failing in #623 and I thought I would help with some housekeeping with what I thought would be a quick fix 😄 .

Now readthedocs starts building, but it fails because sphinx >=5  is required by numpydoc. I tried to add the version requirement to `docs/environment.yml`, but somehow this is bypassed by readthedocs? https://readthedocs.org/projects/dask-jobqueue/builds/23397606/

I could maybe try to drop `conda` and switch to `pip`, also updating Python version (right now it is pinned to 3.8)?

jacobtomlinson:
This workaround from https://github.com/dask/dask-sphinx-theme/issues/68#issuecomment-1893198477 is probably also needed here.

Issue #613, opened 2023-08-04
CI is currently failing


guillaumeeb:
Two things to look at:
- none build which mainly do code formatting checks (see https://github.com/dask/dask-jobqueue/pull/610#issuecomment-1665311514).
- Docker image builds, might be due to Python version problem. (https://github.com/dask/dask-jobqueue/pull/606#issuecomment-1579045899)

jrueb:
One thing I noticed is that this looks really outdated
https://github.com/dask/dask-jobqueue/blob/4bbd0a08342926ee8728ccd9a4559be3aa3ec628/.pre-commit-config.yaml#L7
flake8 moved to github years ago, see https://flake8.pycqa.org/en/latest/release-notes/4.0.0.html "Migrate from GitLab to GitHub"

kaelancotter:
Are you looking for help with this?

guillaumeeb:
Definitely!

kaelancotter:
Created PR [636](https://github.com/dask/dask-jobqueue/pull/636)  to resolve CI failures.  
Unfortunately CI fails in the PR since the updated images aren't pushed from the build step if its a PR - so CI action is still using the old docker images.
LMK if there's another direction I should take this to deal with the docker image catch-22

Specific to SGE tests - adding the --full-trace flag got rid of the seg fault, but I don't have a good root cause analysis.

This PR also doesn't address CI for documentation.

jacobtomlinson:
CI is in a much better place. #653 is still outstanding but I think we can close this issue in favour of that one.

PR #636, opened 2024-03-15
Issue #613 - update conda env setup for CI images

kaelancotter:
No longer attempt to update the base conda env
Use named dask-jobqueue conda environment

### DIFF
--- a/ci/htcondor.sh
+++ b/ci/htcondor.sh
@@ -21,13 +21,13 @@ function jobqueue_before_install {
 
 function jobqueue_install {
     cd ./ci/htcondor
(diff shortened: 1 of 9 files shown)


guillaumeeb:
Ok, so the idea here is to merge this PR, and then attempt to trigger the CI with the new built images?

kaelancotter:
We can double run to confirm the update or I can split the PR if that's preferable.   (Not sure if there are any restrictions on who can manually trigger / re-run Actions)

guillaumeeb:
Okay, let's merge this one, we'll see then.

PR #641, opened 2024-05-24
Graceful Slurm job cancellation

### LINKED ISSUE #640: More graceful job cancellation
Hey all,

This is just a thought for the `SLURMCluster` for now (since that's what I'm familiar with) but similar options may be available in other clusters too. Currently, the `cancel_command` in the `SLURMJob` class is a bare `"scancel"`.

https://github.com/dask/dask-jobqueue/blob/8713202488c664452bf0883bcd4f776536644676/dask_jobqueue/slurm.py#L15

This means that, even when workers are shutdown completely gracefully, the Slurm job is marked as `CANCELLED`. Instead, if the command were `scancel --signal=SIGTERM` the job would be marked as `COMPLETED`. Its possible there could be cases where we would want a job to cancelled, which complicates this somewhat.

In the simple case, however, I think this could be implmented with a simple change of `cancel_command` to:
```python
class SLURMJob(Job):
    # Override class variables
    submit_command = "sbatch"
    cancel_command = "scancel --signal=SIGTERM"
    config_name = "slurm"
```

It'd be great to get some more thoughts on the implications for this.

AlecThomson:
Closes #640

### DIFF
--- a/dask_jobqueue/slurm.py
+++ b/dask_jobqueue/slurm.py
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 class SLURMJob(Job):
     # Override class variables
     submit_command = "sbatch"
-    cancel_command = "scancel"
+    cancel_command = "scancel --signal=SIGTERM"
     config_name = "slurm"
 
     def __init__(


jacobtomlinson (commented):
This looks fine to me! I do wonder if we should expose `cancel_command` as a keyword argument on the `SLURMCluster` and pass that down to `SLURMJob`, in case users want to override it with something else.

AlecThomson:
I found that the HTCondor class already had something similar. I've added this to the base `Job` class in `core`. The downside is that this adds some extra boilerplate.

I've reworked the `HTCondorJob` and `SLURMJob` to make use of the new functionality

dask_jobqueue/lsf.py
```diff
@@ -54,6 +55,28 @@ def __init__(
             use_stdin = dask.config.get("jobqueue.%s.use-stdin" % self.config_name)
         self.use_stdin = use_stdin
 
+        if self.submit_command_extra is None:
```
guillaumeeb, line 58: Why is all this code both in the Core class and the class inheriting it?

guillaumeeb:
Also, the test failures in slurm (and maybe others) looks related to this change.

jacobtomlinson:
CI is now mostly happy on `main` so I've just merged in so we can see up to date CI failures.

AlecThomson:
Sorry for the long lead time on this, everyone. I got myself tripped up between the methods on the base and inheriting classes - as noted by @guillaumeeb's comment. I believe I've got this all sorted now. Hopefully the CI will catch any lingering issues.

jacobtomlinson:
Thanks @AlecThomson! Looks like there are some linting issues (make sure you run `pre-commit install`) and some slurm issues.

AlecThomson:
Hmm - looks like some kind of timing error on the test. I don't quite understand why it's failing... 🤔 
```
>                   assert time() < start + QUEUE_WAIT
E                   assert 1723021595.3378592 < (1723021535.2718754 + 60)
E                    +  where 1723021595.3378592 = time()
```
https://github.com/dask/dask-jobqueue/actions/runs/10278490598/job/28450471395#step:7:425

jacobtomlinson:
It is calling `cluster.scale(n)` and then waiting for the cluster to scale. The time assertion is just a timeout, so it's not scaling to the correct number in the time allowed.

PR #642, opened 2024-05-29
fix bug in OARJob where Job._call is not awaited

sjdv1982:
`OARCluster` is currently broken for me, this simple fix makes it work again.


guillaumeeb (approved):

guillaumeeb:
Thanks @sjdv1982

Issue #644, opened 2024-07-25
No module named 'distutils' in Python 3.12 when using LSFCluster


tcztzy:
**Describe the issue**:

When I try to run dask on LSF cluster with Python 3.12, I found that some exception raised

```
Python 3.12.0 (main, Oct  3 2023, 01:27:23) [Clang 17.0.1 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dask_jobqueue import LSFCluster
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/build/user_home/2022212002/fastoai/.venv/lib/python3.12/site-packages/dask_jobqueue/__init__.py", line 8, in <module>
    from .lsf import LSFCluster
  File "/build/user_home/2022212002/fastoai/.venv/lib/python3.12/site-packages/dask_jobqueue/lsf.py", line 1, in <module>
    from distutils.version import LooseVersion
ModuleNotFoundError: No module named 'distutils'
```

**Minimal Complete Verifiable Example**:

```python
# Put your MCVE code here
from dask_jobqueue import LSFCluster
```

**Anything else we need to know?**:

`distutils` is deprecated since 3.10 and removed in 3.12

https://github.com/dask/dask-jobqueue/blob/a35ac91bbf616ef193327c72fe0d4fced80e768b/dask_jobqueue/lsf.py#L1

https://peps.python.org/pep-0632/#migration-advice

**Environment**:

- Dask version: 2024.7.1
- Dask jobqueue version: 0.8.5
- Python version: 3.12.0
- Operating System: CentOS 7
- Install method (conda, pip, source): pip

PR #646, opened 2024-07-25
Remove unused lsf_version function for Python 3.12 and fix never awaited error for LSFCluster

tcztzy:
Close https://github.com/dask/dask-jobqueue/issues/644


lesteve:
Ah nice, I was doing the same thing in parallel, can you also remove `mock_lsf_version` from `conftest.py` while you are at it.

tcztzy:
> Ah nice, I was doing the same thing in parallel, can you also remove `mock_lsf_version` from `conftest.py` while you are at it.

OK, I am also trying to fix several problems to run LSFCluster

jacobtomlinson (approved):
Things look good here. Merging!

PR #647, opened 2024-08-02
Local slurm pytest

jacobtomlinson:
Today I was experimenting with running the Slurm tests on my development workstation. This PR resolves a couple of nitpicks I found when running tests:

- Slurm generates `*.out` files which get added to version control. Added them to `.gitignore`.
- My workstation doesn't have an `eth0` network interface, so skip the test if not present


jacobtomlinson:
CI seems happy. Merging.

PR #648, opened 2024-08-02
Replace docker-compose with docker compose

jacobtomlinson:
Looks like `docker-compose` v1 has been removed from GitHub Actions. The newer command `docker compose` should be used instead.

All compose files here seem to use v2 or v3 manifests so this should be a simple command replacement.

https://github.com/actions/runner-images/issues/9557


jacobtomlinson:
Build failures appear to be unrelated ot this change. I'm going to merge so I can start trying to fix the individual builds.

PR #650, opened 2024-08-02
Migrate Slurm to use rockylinux

jacobtomlinson:
It looks like the container that the Slurm container was previously based on hasn't been updated on Docker Hub for four years. However the GitHub repo is still active, you just have to clone and build the image yourself. 

It felt easier to vendor in the `Dockerfile` and add our modifications to install Python. In `dask-docker` we used to do a remote build, but since `docker buildx` was made the default this got a lot harder, so I'm apprehensive to go down that road again. I also explored cloning the repo locally and building it before building the main container but using `docker compose build` to build images for pushing doesn't support using cached images that are on Docker Hub.


jacobtomlinson:
Builds are passing again so I'm going to go ahead and merge.

PR #651, opened 2024-08-05
Handle outdated root certificates

jacobtomlinson:
The current SGE image is based on `ubuntu:14.04` as it is a similar age to the last SGE release. The root ca certificates have expired which is causing the miniconda download to fail. This PR tells `curl` to ignore the expired certificates.

In an ideal world we update the container to a more modern OS, but given the age of the SGE packages this feels like a large undertaking.


jacobtomlinson:
SGE builds are now passing. Merging.

PR #652, opened 2024-08-05
Bump minimum Python to 3.9

jacobtomlinson:
Dask dropped 3.8 support in https://github.com/dask/community/issues/315. This PR makes 3.9 the minimum version here.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,6 +40,7 @@ jobs:
       - name: Setup Job queuing system
         if: matrix.jobqueue != 'none'
         run: |
+          cp ci/environment.yml ci/${{ matrix.jobqueue }}/environment.yml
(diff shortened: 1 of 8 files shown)


jacobtomlinson:
Dask Jobqueue uses container images in CI that were built by the latest run on `main`. So int his case those images still use 3.8. In this PR I changed things to build instead of pull to ensure things work with 3.9. 

Everything is passing except for the known SGE test issue (see #653) so I'm confident with the changes here. I'm going to revert the builds back to pulls and merge which should get things happy again once `main` is built.

jacobtomlinson:
Looking at the times here it actually only took a couple of minutes to build the containers. Pull an image built from a previous branch feels a little funny and causes a headache in PRs like this. So I'm going to leave the builds in here instead of the pulls. We can always revert this in the future if it becomes a problem, or explore caching the images in a different way.

PR #654, opened 2024-08-05
Narrow down SGE segfault

jacobtomlinson:
This is a draft PR where I'm disabling all the SGE tests and then adding them back one by one to try and figure out which is causing the segfault.


jacobtomlinson:
Looks like any of the tests that use `SGECluster` cause the segfault. Interestingly the error only appears once regardless f how many tests are enabled. Closing as narrowing down clearly isn't a useful thing to do.

PR #655, opened 2024-08-06
Replace pytest-asyncio with anyio

jacobtomlinson:
In other Dask projects we've found that `pytest-asyncio` causes a bunch of headaches. The test tooling from `anyio` is much more reliable.

This PR replaces `pytest-asyncio` with `anyio` as a test dependency. It configures `anyio` to only use the `asyncio` backend for testing and replaces the pytest marks where needed.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -9,5 +9,5 @@ dependencies:
   - flake8
   - black
   - pytest
-  - pytest-asyncio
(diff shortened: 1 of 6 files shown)


jacobtomlinson:
Looks like we still need `pytest-asyncio` to be included in the environment.

PR #656, opened 2024-08-06
Enable GitHub Actions build concurrency groups

jacobtomlinson:
Replace the `styfle/cancel-workflow-action` action with GitHub Actions concurrency groups. This is the modern way to cancel running workflows if a new commit is pushed to a PR.


jacobtomlinson:
I can see that 421aaf6 successfully cancelled when I pushed 81b724e so this should be good to merge.

Issue #638, opened 2024-05-09
Should we merge Dask HPC Runners in here?


jacobtomlinson:
For a while I've been playing around with [this prototype repo](https://github.com/jacobtomlinson/dask-runners) which implements Dask Runners for HPC systems. I'm motivated to reduce the fragmentation and confusion around tooling in the Dask HPC community, so perhaps this new code should live here.

In https://github.com/dask/community/issues/346 I wrote up the difference between _Dask Clusters_ and _Dask Runners_. The **TL;DR** is that a Cluster creates the scheduler and worker tasks directly, for example `dask_jobqueue.SLURMCluster` submits jobs to SLURM for each worker. A Dask Runner is different because it is invoked from within an existing allocation and populates that job with Dask processes. This the same as how [Dask MPI](http://mpi.dask.org/en/latest/#) works.

### `SlurmRunner` Example

If I write a Python script and call it with `srun -n 6 python myscript.py` the script will be invoked by Slurm 6 times in parallel on 6 different nodes/cores on the HPC. The Dask Runner class then uses the Slurm process ID environment variable to decide what role reach process should play and uses the shared filesystem to bootstrap communications with a scheduler file.

```python
# myscript.py
from dask.distributed import Client
from dask_hpc_runner import SlurmRunner

# When entering the SlurmRunner context manager processes will decide if they should be 
# the client, schdeduler or a worker.
# Only process ID 1 executes the contents of the context manager.
# All other processes start the Dask components and then block here forever.
with SlurmRunner(scheduler_file="/path/to/shared/filesystem/scheduler-{job_id}.json") as runner:

    # The runner object contains the scheduler address info and can be used to construct a client.
    with Client(runner) as client:

        # Wait for all the workers to be ready before continuing.
        client.wait_for_workers(runner.n_workers)

        # Then we can submit some work to the Dask scheduler.
        assert client.submit(lambda x: x + 1, 10).result() == 11
        assert client.submit(lambda x: x + 1, 20, workers=2).result() == 21

# When process ID 1 exits the SlurmRunner context manager it sends a graceful shutdown to the Dask processes.
```

### Should this live in `dask-jobqueue`?

I'm at the point of trying to decide where this code should live within the Dask ecosystem. So far I have implemented `MPIRunner` and `SlurmRunner` as a proof-of-concept. It would be very straight forward to write runners for other batch systems provided it is possible to detect the process ID/rank from the environment.

[...]

Issue #653, opened 2024-08-05
SGE Tests segfault in CI


jacobtomlinson:
Opening an issue to triage the segfault that seems to be happening in th SGE tests.

For some time the SGE tests have been failing. When you look at the logs of [a recent run](https://github.com/dask/dask-jobqueue/actions/runs/10250992141/job/28357989642) on `main` it contains the following error.

```
 *** Error in `/opt/anaconda/envs/dask-jobqueue/bin/python3.8': double free or corruption (!prev): 0x00005626c18fa470 ***
/bin/bash: line 1:   588 Aborted                 (core dumped) pytest /dask-jobqueue/dask_jobqueue --verbose --full-trace -s -E sge
```

I also opened #652 to bump the minimum Python version here to 3.9 and I see a [similar issue happening](https://github.com/dask/dask-jobqueue/actions/runs/10251115360/job/28358406864?pr=652) but with a slightly different error.

```
*** Error in `/opt/anaconda/envs/dask-jobqueue/bin/python3.9': corrupted size vs. prev_size: 0x00005560404681a0 ***
/bin/bash: line 1:   592 Aborted                 (core dumped) pytest /dask-jobqueue/dask_jobqueue --verbose --full-trace -s -E sge
```

Strangely in both cases `pytest` reports everything has passed.

```
9 passed, 270 skipped in 26.69s
```

jacobtomlinson:
In #654 I've been playing around with skipping various tests and enabling them again. It seems like enabling any of the tests results in the segfault. Enabling more than one only causes the error to appear once though.

jacobtomlinson:
I have a local reproducer now. Here are the steps I took to get it set up on my machine.

```bash
# Build SGE container
cd ci/sge
cp ../environment.yaml .
docker compose build

# Start SGE stack (based on ci/sge.sh)
./start-sge.sh
docker exec sge_master /bin/bash -c "chmod -R 777 /shared_space"

# Install dask-jobqueue in editible install
docker exec sge_master conda run -n dask-jobqueue /bin/bash -c "cd /dask-jobqueue; pip install -e ."
```

I also installed `anyio` and used `@pytest.mark.anyio` instead of `@pytest.mark.asyncio` because I find the behaviour a lot more consistent. See #655.

I then created a new test file with a single test that consistently reproduces the segfault.

```python
# dask_jobqueue/tests/test_jsge_segfault.py
from dask_jobqueue.sge import SGECluster
from dask.distributed import Client

[...]

PR #659, opened 2024-08-07
Add SLURMRunner from jacobtomlinson/dask-hpc-runners

jacobtomlinson:
Towards #638 

As we decided that the code from [jacobtomlinson/dask-hpc-runners](https://github.com/jacobtomlinson/dask-runners) would be best to live in this repo this PR adds the `SLURMRunner` along with the base class and other core code.

### Design

The core assumption in the design of the runner model is that the same script will be executed many times by a job scheduler. 

```
(mpirun|srun|qsub|etc) -n4 myscript.py
├── [0] myscript.py
├── [1] myscript.py
├── [2] myscript.py
└── [3] myscript.py
```

Within the script the runner class is created. 

```python
from dask_jobqueue import SomeRunner

with SomeRunner(**kwargs) as runner:
    with Client(runner) as client:
        client.wait_for_workers(2)
        # Do some Dask work
```

This will result in multiple processes runnning on an HPC that are all instantiating the runner class. 

The processes need to coordinate to decide which process should run the Dask Scheduler, which should be Dask Workers and which should continue running the rest of the client code within the script. This coordination happens during the `__init__()` of the runner class.

The Scheduler and Worker processes exit after they complete to avoid running the client code multiple times. This means that only one of the processes will continue past the `__init__()` of the runner class, the rest will exit at that point after the work is done.

### Base class

In the new file `runner.py` contains the `BaseRunner` class that can be used for implementing other runners. It also includes an `AsyncRunner` class that is used for testing.

The minimum required to implement a new runner is the following methods.

```python
from dask_jobqueue.runner import BaseRunner

class MyRunner(BaseRunner):

    async def get_role(self) -> str:
        """Figure out whether I am a scheduler, worker or client.

        A common way to do this is by using a process ID. Many job queues give each process
        a monotonic index that starts from zero. So we can assume proc 0 is the scheduler, proc 1
        is the client and any other procs are workers.
        """
        ...

    async def get_scheduler_address(self) -> str:
        """If I am not the scheduler discover the scheduler address.

        A common way to do this is to read a scheduler file from a shared filesystem.

        Alternatively if the scheduler process can broadcast it's address via something like MPI
        we can define ``BaseRunner.set_scheduler_address()`` which will be called on the scheduler 
        and then recieve the broadcast in this method.
        """
        ...
```

The `BaseRunner` class handles starting up Dask once these methods have been implemented. It also provides many stubbed out hooks to allow you to write code that runs before/after each component is created. E.g `BaseRunner.before_scheduler_start()`, `BaseRunner.before_worker_start()` and `BaseRunner.before_client_start()`.

The runner must know the address of the scheduler so that it can coordinate the clean shutdown of all processes when we reach the end of the code (either via `__exit__()` or a finalizer). This communication happens independently of any clients that may be created.

### Slurm implementation

This PR also adds a Slurm implementation to `slurm.py`.

In the `get_role()` method I use the `SLURM_PROCID` environment variable to infer the role.

I also add a default scheduler option to set the `scheduler_file="scheduler-{job_id}.json"` and I look up the Job ID from the `SLURM_JOB_ID` environment variable to ensource uniqueness. This effectively allows us to broadcast the scheduler address via the shared filesystem.

Then in the `get_scheduler_address()` method I wait for the scheduler file to exist and then open and read the address from the scheduler file in the same way the `dask.distributed.Client` does. 

#### Example

```python
# myscript.py
from dask_jobqueue.slurm import SLURMRunner

with SLURMRunner(scheduler_file="/path/to/shared/filesystem/dask-slurm-{job_id}.json") as runner:
    with Client(runner) as client:
        client.wait_for_workers(2)
        # Do some Dask work
 ```

Then I can submit this script via `srun` or an `sbatch` script.

```bash
srun -n4 myscript.py
```

### TODO

- [x] Add base runner class
- [x] Add base tests
- [x] Add Slurm runner
- [x] Add Slurm tests
- [x] Update documentation


jacobtomlinson:
I'm going to mark this as ready to review as all the tehnical work is done and the tests are passing (ignore the failed SGE test #653). 

It still needs docs but I'd love to get a review, even just a high level design review, before I dive into writing up the documentation. @guillaumeeb @lesteve @kmpaul if any of you have time to take a look I'd really appreciate it!

jacobtomlinson:
I've gone over the documentation and given is an overhaul to make space for the new features.

- I've consolidated the existing documentation under the **Dynamic Clusters** section.
- I've added a new **Batch Runners** section to document the new classes.
- I've updated the index to describe the difference between the two paradigms.
- I've also fixed up a few Sphinx warnings, but the existing documentation is largely untouched.
- Also added redirects for some of the pages I renamed under the new structure.

jacobtomlinson:
Other maintainers of this repo expressed to me via email they they have limited capacity for reviews at the moment. To avoid being blocked on review any longer I'm going to self-merge this. But if anyone wants to revisit this work with a review down the line then I would encourage you to open an issue where we can discuss things.

PR #663, opened 2024-08-21
Use GitHub Actions for PyPI publish on release

jacobtomlinson:
Adds a GitHub Actions workflow that builds new tags and pushes them to PyPI. This removes a little bit of potential human error when it comes to building the distributions.

The new release process will be:
- Update the Changelog
- Create a new tag on main
- Push the tag to the upstream repo
- Wait for bots to do their thing

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,26 @@
+name: Publish to PyPI
+on:
+  push:
+    tags:
+      - "*.*.*"
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    if: github.repository == 'dask/dask-jobqueue'
+    environment:
+      name: pypi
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v1
+        with:
+          python-version: "3.10"
+      - name: Build package
+        run: pip install build && pyproject-build
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@release/v1
```
jacobtomlinson, line 26: For anyone interested in how this works without passing any credentials I had to log into PyPI and set up GitHub Actions as a [trusted publisher](https://docs.pypi.org/trusted-publishers/adding-a-publisher/).

PR #675, opened 2024-12-22
Fix slurm_format_bytes_ceil logic and add unit tests

### LINKED ISSUE #674: The function slurm_format_bytes_ceil raises Typerror if input is less than 1024
The function "[slurm_format_bytes_ceil](https://github.com/dask/dask-jobqueue/blob/b07308ea36da20ba4a109e53799e1dc26b613762/dask_jobqueue/slurm.py#L115)" parses number of bytes to a string compatible with Slurm.

If input number of bytes "n" is less than 1024, then it reaches the end of the conditional statements, which is missing a formatter string "%s". Thus, it throws an exception:
```
TypeError: not all arguments converted during string formatting
```
This also goes against its intended behavior, as shown in the examples in the docstring.

A simple fix is to replace `return "1K" % n` with `return "%sK" % n`

maneesh29s:
Closes #674

### DIFF
--- a/dask_jobqueue/slurm.py
+++ b/dask_jobqueue/slurm.py
@@ -117,12 +117,24 @@ def slurm_format_bytes_ceil(n):
 
     SLURM expects KiB, MiB or Gib, but names it KB, MB, GB. SLURM does not handle Bytes, only starts at KB.
 
+    Parameters
+    ----------
+    n: int
+        Number of bytes
+
+    Returns
+    -------
+    str
+        Bytes formatted as string
+
+    Examples
+    --------
     >>> slurm_format_bytes_ceil(1)
     '1K'
     >>> slurm_format_bytes_ceil(1234)
     '2K'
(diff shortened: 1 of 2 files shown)


dask_jobqueue/slurm.py
```diff
@@ -134,7 +146,7 @@ def slurm_format_bytes_ceil(n):
         return "%dM" % math.ceil(n / (1024**2))
     if n >= 1024:
         return "%dK" % math.ceil(n / 1024)
-    return "1K" % n
+    return "%sK" % n
```
jacobtomlinson, line 149: I think the intent here is that is you enter a number like `34` for 34 bytes then it will automatically round up to `1K` which is the minimum that SLURM supports. So I think the actual fix is to just return `1K`.

```suggestion
    return "1K"
```

jacobtomlinson (changes requested):

jacobtomlinson (approved):
This looks great thanks!

PR #677, opened 2025-01-11
Add argument 'job_script_epilogue'

### LINKED ISSUE #639: Job script epilogue
Would be nice to have `job_script_prologue` equivalent but for cleanup. I.e. run some commands after worker exists. The argument could be called `job_script_epilogue`.

maneesh29s:
Closes #639 

This allows user to specify commands that should be run after the worker command has exited.

The working is similar to 'job_script_prologue' argument.

### DIFF
--- a/dask_jobqueue/core.py
+++ b/dask_jobqueue/core.py
@@ -55,6 +55,8 @@ job_parameters = """
         Deprecated: use ``job_script_prologue`` instead. This parameter will be removed in a future version.
     job_script_prologue : list
         Other commands to add to script before launching worker.
+    job_script_epilogue : list
+        Commands to add to script which will run after the worker command has exited.
     header_skip : list
         Deprecated: use ``job_directives_skip`` instead. This parameter will be removed in a future version.
     job_directives_skip : list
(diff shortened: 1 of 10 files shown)


maneesh29s:
I have not added any new example in the documentation for this (I see there's an entire section dedicated to [job_script_prologue](https://jobqueue.dask.org/en/latest/clusters-advanced-tips-and-tricks.html#run-setup-commands-before-starting-the-worker-with-job-script-prologue)).

Do you think it is necessary to give examples for this paramter seperately?

guillaumeeb:
So sorry that we never provided any feedback here, this looks perfect to me. @jacobtomlinson any thought?

About the documentation, just a mention of this arg in the doc section you are pointing to would have been great, but I guess you are probably not into this, are you @maneesh29s?

guillaumeeb (approved):

maneesh29s:
No problem, I will update the documentation

jacobtomlinson (approved):
Sure this looks great to me!

maneesh29s:
@guillaumeeb I have updated the docs as you suggested.

guillaumeeb:
Thanks @maneesh29s !

PR #680, opened 2025-01-30
Set -N=1/--nodes=1 for SLURM workers

lpsinger:
The SLURM worker hardcodes --ntasks=1/-n=1, but on the SLURM cluster that I have experience with (SDSC Expanse), you also have to set -N=1/--nodes=1.

### DIFF
--- a/dask_jobqueue/slurm.py
+++ b/dask_jobqueue/slurm.py
@@ -81,9 +81,10 @@ class SLURMJob(Job):
         if account is not None:
             header_lines.append("#SBATCH -A %s" % account)
 
-        # Init resources, always 1 task,
+        # Init resources, always 1 task, 1 node,
         # and then number of cpu is processes * threads if not set
         header_lines.append("#SBATCH -n 1")
+        header_lines.append("#SBATCH -N 1")
         header_lines.append(
             "#SBATCH --cpus-per-task=%d" % (job_cpu or self.worker_cores)
         )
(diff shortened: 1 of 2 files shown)


guillaumeeb:
These configurations are always tricky and specific, and I'm not sure this particular option is really needed in most Slurm cluster? The real question actually is: does this option is available and mandatory in most Slurm cluster configuration?

You can always set it with dask-jobqueue kwargs in your context.

lpsinger:
> These configurations are always tricky and specific, and I'm not sure this particular option is really needed in most Slurm cluster? The real question actually is: does this option is available and mandatory in most Slurm cluster configuration?

I think it is quite likely that it is _available_ in every Slurm cluster configuration because this option is documented by the [sbatch man page](https://slurm.schedmd.com/sbatch.html). I don't know whether it is mandatory on every cluster, but I also don't see that there would be any harm even if it isn't required. Slurm won't let you set `--nodes` to a value greater than `--ntasks`. Here's what happens if you try to do that:

```
$ srun --pty -A project -p debug --nodes=2 --ntasks=1 --cpus-per-task=1 -t 00:00:05 sleep 1
srun: warning: can't run 1 processes on 2 nodes, setting nnodes to 1
srun: job 38602701 queued and waiting for resources
srun: job 38602701 has been allocated resources
```

guillaumeeb (approved):
Okay, I agree this should always be available, so I'm OK to add this if this simplifies things for users.

lpsinger:
Thank you!

lpsinger:
Would you please do a release that includes this patch?

PR #689, opened 2025-05-28
Update .gitignore to follow Python.gitignore

bewygs:
Just a small PR that updates `.gitignore` to match the latest [Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore).

I ran into issues where setting up my environment with `.venv` caused lots of files to appear in git that should have been ignored.

If you notice any project-specific ignores missing, just let me know!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-# C extensions
-*.so
-
 # Distribution / packaging
 .Python
 build/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,25 +1,139 @@
...
+.idea/
+.vscode/
+
+# OS and swap files
 .DS_Store
 *.swp
 *.swo
-.cache/
-.pytest_cache
-docs/source/generated
```
jacobtomlinson, line 16: This one seems to have been lost

